PR #18, opened 2015-08-09
Mc pre

srikrishnagopu:


srikrishnagopu:
yes I added the dissociate node support now. 
Other changes looks good.

PR #27, opened 2015-10-07
Mirroring

antoninbas:


pbhide:
Is there a way to specify queue_rate and queue_depth as command line options when starting the bmv2 model ? That would be nice.

antoninbas:
No, because right now there is no way to parse target specific options(as opposed to target-independent options like --interface or --nanolog). That's a good point though, I'll look into adding this feature later. However this is very easy using the bmv2 CLI using commands `set_queue_depth` and `set_queue_rate`. Send me an email of you need to run them and want more info.

PR #37, opened 2015-11-06
set sswitch_CLI args.pre to SimplePreLAG

tudang:
sswitch_CLI is assumed to use SimplePreLAG but the default argument is set to use SimplePre.


antoninbas:
Hi Tu,
Next time please rebase master into your development branch to avoid the merge commit.
Thanks

PR #40, opened 2015-11-20
Avoid "what" as field name

notmem:
I'm not sure if this has to do with the version of Apache Thrift that
I'm using, but I'm seeing errors

In file included from gen-cpp/standard_constants.h:10:0,
                 from gen-cpp/standard_constants.cpp:7:
gen-cpp/standard_types.h:616:34: error: 'const char* bm_runtime::standard::InvalidTableOperation::what() const' conflicts with a previous declaration
   const char* what() const throw();
                                  ^
gen-cpp/standard_types.h:593:33: note: previous declaration 'bm_runtime::standard::TableOperationErrorCode::type bm_runtime::standard::InvalidTableOperation::what'
   TableOperationErrorCode::type what;
                                 ^

If you squint hard enough you'll see that the problem is that there's a
method called "what" (to be compatible with std::exception, I guess) and
a field called "what" as well. The method comes from the base class.

By changing the .thrift file to use a field name other than "what", the
compilation issue is resolved.

Signed-off-by: Marcelo E. Magallon <marcelo.magallon@hpe.com>

### DIFF
--- a/modules/bm_runtime/src/SimplePreLAG_server.ipp
+++ b/modules/bm_runtime/src/SimplePreLAG_server.ipp
@@ -39,7 +39,7 @@ public:
       pre->mc_mgrp_create(mgrp, &mgrp_hdl);
     if(error_code != McSimplePre::SUCCESS) {
       InvalidMcOperation imo;
-      imo.what = (McOperationErrorCode::type) error_code;
+      imo.code = (McOperationErrorCode::type) error_code;
       throw imo;
     }
     return mgrp_hdl;
@@ -51,7 +51,7 @@ public:
       pre->mc_mgrp_destroy(mgrp_handle);
     if(error_code != McSimplePre::SUCCESS) {
       InvalidMcOperation imo;
-      imo.what = (McOperationErrorCode::type) error_code;
+      imo.code = (McOperationErrorCode::type) error_code;
       throw imo;
     }
   }
@@ -65,7 +65,7 @@ public:
       pre->mc_node_create(rid, port_map, lag_map, &l1_hdl);
     if(error_code != McSimplePre::SUCCESS) {
       InvalidMcOperation imo;
-      imo.what = (McOperationErrorCode::type) error_code;
+      imo.code = (McOperationErrorCode::type) error_code;
       throw imo;
     }
     return l1_hdl;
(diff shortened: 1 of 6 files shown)


antoninbas:
Hi Marcello,

This is the first time I am hearing of this issue, but I would be happy to take your fix. However this change has an impact on other parts of the code (e.g. [runtime_CLI.py] (https://github.com/p4lang/behavioral-model/blob/master/tools/runtime_CLI.py)) which I also need to update. Since this is not showing in my setup or in the Travis regression, would you mind telling me which Thrift version you are using as well as which C++ compiler?

antoninbas:
Somebody just confirmed that this was happening with the latest thrift 0.9.3. While we officially only support thrift 0.9.2, changing this name seems like a small price to pay.
I will pick up Marcelo's commit and write the required changes to ([runtime_CLI.py] (https://github.com/p4lang/behavioral-model/blob/master/tools/runtime_CLI.py) and [p4c-bm] (https://github.com/p4lang/p4c-bm).

antoninbas:
commit picked up in #41

notmem:
Hi Antonin,

Thanks for looking at this. I didn't catch the issue with Python.

I have:

$ thrift --version
Thrift version 0.9.3

$ gcc --version
gcc (Debian 4.9.2-10) 4.9.2

g++ is the same version.

What's actually relevant for what I'm doing (OpenSwitch), is:

Thrift version 0.9.2
x86_64-linux-gnu-gcc (Debian 4.9.2-10) 4.9.2

I see the same issue in both cases.

Thanks,

Marcelo

On 11/20/2015 05:17 PM, Antonin Bas wrote:
>
> Hi Marcello,
>
> This is the first time I am hearing of this issue, but I would be 
> happy to take your fix. However this change has an impact on other 
> parts of the code (e.g. runtime_CLI.py 
> <https://github.com/p4lang/behavioral-model/blob/master/tools/runtime_CLI.py>) 
> which I also need to update. Since this is not showing in my setup or 
> in the Travis regression, would you mind telling me which Thrift 
> version you are using as well as which C++ compiler?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/p4lang/behavioral-model/pull/40#issuecomment-158554073>.
>

-- 
Marcelo E. Magallón

┏━━━━┓
┗━━━━┛
Hewlett Packard
Enterprise

notmem:
Thanks!

On 11/20/2015 10:32 PM, Antonin Bas wrote:
>
> Closed #40 <https://github.com/p4lang/behavioral-model/pull/40>.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/p4lang/behavioral-model/pull/40#event-471086184>.
>

-- 
Marcelo E. Magallón

┏━━━━┓
┗━━━━┛
Hewlett Packard
Enterprise

PR #72, opened 2016-02-18
simplified passing compile time options to configure

antoninbas:
Logging macros are enabled by default, they can be disabled by passing `--disable-logging-macros` to configure.


vgurevich:
Nice! Thanks a lot, Antonin!

PR #89, opened 2016-03-04
P4thrift

pierce-m:
Allows people to compile bmv2 with a local copy of p4thrift or an installed copy of apache thrift like usual.


Makefile.am
```diff
@@ -1,5 +1,9 @@
 ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS} -I m4
 
+if P4THRIFT
+  AM_CPPFLAGS += -DP4THRIFT
```
antoninbas: I don't think this propagate to the subdirs, have you checked? You may need to add it to `MY_CPPFLAGS` in `configure.ac`

configure.ac
```diff
@@ -80,6 +80,20 @@ AC_SUBST([AM_CPPFLAGS], ["$MY_CPPFLAGS \
                           -isystem\$(top_srcdir)/third_party/spdlog/include \
                           -I\$(top_srcdir)/modules/BMI/include"])
 
+# Thrift
```
antoninbas: I am a little confused. I think you are asking the user to give you the path to the `thrift` compiler executable that he wants to use (either the apache one or the p4 one). But what about the thrift library? What if the wrong thrift library is picked, i.e. not the one corresponding to the `thrift` executable you want? Please correct me if I am missing something.
I believe that you should not make this a configure option. You should pick whichever comes first in the path (which is what is happening for the library right now). Instead of your `configure.ac` script should detect which library is being used and define the  `P4THRIFT` file if needed. I think th easier way to do this is to have a header file in your p4 thrift package that does not exist in the apache one. You can then do the detection with code similar to this:
```
AC_CHECK_HEADER([thrift/dummy_p4_thrift.h], [use_p4_thrift=yes], [use_p4_thrift=no])])
// one of these 2 lines I guess, depending on what you need, but I am guessing the second one
AM_CONDITIONAL([P4THRIFT], [test "$use_p4_thrift" = "yes"])
AS_IF([test "$use_p4_thrift" = "yes"], [MY_CPPFLAGS="$MY_CPPFLAGS -DP4THRIFT])
```

modules/bm_apps/src/learn.cpp
```diff
@@ -87,12 +87,22 @@ LearnListener::ack_buffer(cxt_id_t cxt_id, list_id_t list_id,
 
 void
 LearnListener::start() {
+#ifdef P4THRIFT
+  using p4::thrift::protocol::TProtocol;
```
antoninbas: Please use a namespace alias (applies to sections below as well):
```
#ifdef P4THRIFT
namespace thrift_provider = p4::thrift
#else
namespace thrift_provider = apache::thrift
#endif
using thrift_provider::protocol::TProtocol;
using thrift_provider::protocol::TBinaryProtocol;
using thrift_provider::protocol::TMultiplexedProtocol;
using thrift_provider::transport::TSocket;
using thrift_provider::transport::TTransport;
using thrift_provider::transport::TBufferedTransport;
```

pierce-m:
So the idea of this pull request is to make it so people who might have another version of thrift installed on their system can simply install our p4thrift and build behavioral model as usual without having to uninstall anything.

modules/bm_apps/Makefile.am
```diff
@@ -5,12 +5,16 @@ AM_CPPFLAGS += \
 -I$(srcdir)/include \
 -I$(top_builddir)/thrift_src/gen-cpp
 
+if P4THRIFT
+  AM_CPPFLAGS += -DP4THRIFT $(THRIFT_INC)
```
antoninbas: `THRIFT_INC` is not defined anywhere, I don't think it is needed

pierce-m: Right I noticed that. Working on the style check now but it looks like it's failing on files I haven't changed.

modules/bm_apps/examples/Makefile.am
```diff
@@ -1,4 +1,14 @@
 AM_CPPFLAGS = -I$(top_srcdir)/modules/bm_apps/include
+
+if P4THRIFT
+  AM_CPPFLAGS+=$(THRIFT_INC) $(THRIFT_LIB)
```
antoninbas: why is `THRIFT_LIB` in `CPPFLAGS`?

pierce-m: Just trying to link the correct library and this worked. Is there a better variable I should use?

modules/bm_apps/examples/Makefile.am
```diff
@@ -1,4 +1,14 @@
...
+  ifndef AM_LDFLAGS
```
antoninbas: please just add `$(THRIFT_LIB)` to `LDADD` below

modules/bm_apps/examples/Makefile.am
```diff
@@ -1,4 +1,14 @@
 AM_CPPFLAGS = -I$(top_srcdir)/modules/bm_apps/include
+
+if P4THRIFT
+  AM_CPPFLAGS+= $(THRIFT_LIB)
```
antoninbas: `CPPFLAGS` are the preprocessor flags...
You can add `$(THRIFT_LIB)` to `LDADD` below, and you don't need the `ifndef` block that follows

thrift_includes.h
```diff
@@ -0,0 +1,25 @@
+#ifdef P4THRIFT
```
antoninbas: I think this is bad style to have a file like this in the root directory and to include unconditionally everything Thrift-related (i.e. in some cases you include more than you need). I think the best solution is still to have a `#ifdef P4THRIFT` switch in each source file that requires some Thrift headers, and only include what you need.
There would be another solution requiring to have autoconf process the file, but I think that would be an overkill at the point and would make the source code more complicated.

antoninbas:
When you are done with addressing the comments, please remember to squash your commits so that your branch is exactly one commit ahead of master.

pierce-m:
Replaced by pull request 90.

PR #90, opened 2016-03-16
Add support for P4Thrift.

pierce-m:
Replaces pull request 89 -- because of intermediate merge commits I found rebasing to be unnecessarily complicated, so I simply pulled the changes into this branch.


antoninbas:
Looks good to me, will merge after the regression has passed

PR #94, opened 2016-03-24
Add support for arbitrary data structures in MatchUnits

grodtron:
Overview
-----------------
The purpose of this pull request is to generalize the match units so that the user can provide an implementation for the actual data structures that are used. My particular use case is using the behavioral model within a SystemC project. In order to obtain timing information we need to provide our own data structures which interact with our memory model. Another example use case would be changing to a different Trie implementation which may perform better for particular patterns of lookups.

Additionally, this change reduces the duplication in `match_units.{cpp,h}`, factoring out the common logic into a templated class, and the actual lookup logic into the separate `lookup_structure.{cpp,h}` classes.

Details
----------
Users can provide their own data structure implementations by subclassing `LookupStructure<KeyType>`, and then by subclassing `LookupStructureFactory` and overriding the appropriate `create()` methods.

This factory can then be passed to a Switch instance before calling `init_objects()` or `init_from_command_line()` using the `set_lookup_factory()` method. If this is not done, then a default factory is used which creates lookup structures that use the same logic that was previously used in `match_units.{cpp,h}`.

Performance
--------------------------
The main negative performance impact is that for Ternary match units, match keys are unfortunately being copied when stored (similarly to the Exact match units). This increases the memory usage, but should not have a significant impact on speed.

Another much smaller impact is the additional virtual function calls required. This is an unavoidable consequence of allowing the user to specify the data structure used in lookups, and the performance impact should be minimal.

Coverage
--------------
   - No additional test cases were added, since this was
   essentially a refactoring, all new code was covered by existing cases.
   - Coverage either improved or stayed the same in all cases with one exception: `modules/bm_sim/include/bm_sim/handle_mgr.h`
      - Line 114 is no longer being hit. I do not have a complete understand of this module, but my guess is this is because previously Ternary matches were using the `HandleMgr` to iterate over all handles, and now this is being done by in the corresponding lookup structure.
      - Would it be possible to suggest a test-case which would restore coverage for this line?

### DIFF
--- a/modules/bm_sim/include/bm_sim/P4Objects.h
+++ b/modules/bm_sim/include/bm_sim/P4Objects.h
@@ -61,7 +61,9 @@ class P4Objects {
   explicit P4Objects(std::ostream &outstream = std::cout)
       : outstream(outstream) { }
 
-  int init_objects(std::istream *is, int device_id = 0, size_t cxt_id = 0,
+  int init_objects(std::istream *is,
+                   LookupStructureFactory * lookup_factory,
+                   int device_id = 0, size_t cxt_id = 0,
                    std::shared_ptr<TransportIface> transport = nullptr,
                    const std::set<header_field_pair> &required_fields =
                      std::set<header_field_pair>(),
--- a/modules/bm_sim/include/bm_sim/lookup_structures.h
+++ b/modules/bm_sim/include/bm_sim/lookup_structures.h
@@ -1,0 +1,186 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Gordon Bailey (gb@gordonbailey.net)
+ *
+ */
+
+//! @file lookup_structures.h
+//! This file contains 2 classes: bm::LookupStructure and
+//! bm::LookupStructureFactory. When implementing your target, you may wish
+//! to provide custom implementations of the data structures used to perform
+//! matches. This can be done by inheriting from one or more of
+//! bm::ExactLookupStructure, bm::LPMMatchStructure, and
+//! bm::TernaryMatchStructure. Each of these is a specialization of the
+//! bm::LookupStructure template for the corresponding match key type.
+//!
+//! Once the implementation of the new lookup structure is complete, you must
+//! extend the bm::LookupStructureFactory class to build instances of it. This
+//! is relatively simple. An (abridged) example of creating a new bm::Switch
+//! using a custom lookup structure is as follows
+//!
+//! @code
+//!
+//! // Extend the appropriate LookupStructure type
+//! class MyExactLookupStructure : public bm::ExactLookupStructure {
+//!  public:
+//!   virtual ~MyExactLookupStructure() = default;
+//!
+//!  bool lookup(const bm::ByteContainer &key_data,
+//!              bm::internal_handle_t *handle) const override {
+//!    // ...
+//!  }
+//!
+//!  bool entry_exists(const bm::ExactMatchKey &key) const override {
+//!    // ...
+//!  }
+//!
+//!  void add_entry(const bm::ExactMatchKey &key,
+//!                 bm::internal_handle_t handle) override {
+//!    // ...
+//!  }
+//!
+//!  void delete_entry(const bm::ExactMatchKey &key) override {
+//!    // ...
+//!  }
+//!
+//!  void clear() override {
+//!    // ...
+//!  }
+//!
+//!  private:
+//!   // implementation details ...
+//! };
+//!
+//! // Create a factory subclass which creates your new structure
+//! class MyFactory : public bm::LookupStructureFactory {
+//!  public:
+//!   virtual ~MyFactory() = default;
+//!
+//!   // Note that it's only necessary to override the create function for the
+//!   // particular match types you are interested in, the others will use the
+//!   // default implementations.
+//!   std::unique_ptr<bm::ExactLookupStructure>
+//!   create_for_exact(size_t size, size_t nbytes_key) override {
+//!     return std::unique_ptr<bm::ExactLookupStructure>(
+//!                new MyExactLookupStructure(/* parameters */));
+//!   }
+//! };
+//!
+//! // Override bm::Switch to create your target
+//! class MySwitch : public bm::Switch {
+//!    // ...
+//! };
+//!
+//! int main(int argc, const char *argv[]) {
+//!   // Create an instance of your switch object
+//!   auto switch = new MySwitch();
+//!
+//!   // Pass it an instance of your factory class.
+//!   switch->set_lookup_factory(std::shared_ptr<bm::LookupStructureFactory>(
+//!                                  new MyFactory()));
+//!
+//!   // Initialize the switch. Note that this must be done *after* calling
+//!   // `set_lookup_factory`, otherwise the default lookup structures will
+//!   // be used.
+//!   int status = switch->init_from_command_line(argc, argv);
+//!   // ...
+//! }
+//!
+//! @endcode
+
+#ifndef BM_SIM_INCLUDE_BM_SIM_LOOKUP_STRUCTURES_H_
+#define BM_SIM_INCLUDE_BM_SIM_LOOKUP_STRUCTURES_H_
+
+#include "match_key_types.h"
+#include "bytecontainer.h"
+
+namespace bm {
+
+//! This class defines an interface for all data structures used
+//! in Match Units to perform lookups. Custom data strucures can
+//! be created by implementing this interface, and creating a
+//! LookupStructureFactory subclass which will create them.
+template <typename K>
+class LookupStructure {
+ public:
+  virtual ~LookupStructure() = default;
+
+  //! Look up a given key in the data structure.
+  //! Return true and set handle to the found value if there is a match
+  //! Return false if there is no match (value in handle is undefined)
+  virtual bool lookup(const ByteContainer &key_data,
+                      internal_handle_t *handle) const = 0;
+
+  //! Check whether an entry exists. This is distinct from a lookup operation
+  //! in that this will also match against the prefix length in the case of
+  //! an LPM structure, and against the mask and priority in the case of a
+  //! Ternary structure.
+  virtual bool entry_exists(const K &key) const = 0;
+
+  //! Store an entry in the lookup structure. Associates the given handle
+  //! with the given entry.
+  virtual void add_entry(const K &key, internal_handle_t handle) = 0;
+
+  //! Remove a given entry from the structure. Has no effect if the entry
+  //! does not exist.
+  virtual void delete_entry(const K &key) = 0;
+
+  //! Completely remove all entries from the data structure.
+  virtual void clear() = 0;
+};
+
+// Convenience typedefs to simplify the code needed to override LookupStructure
+// and LookupStructureFactory
+typedef LookupStructure<ExactMatchKey>   ExactLookupStructure;
+typedef LookupStructure<LPMMatchKey>     LPMLookupStructure;
+typedef LookupStructure<TernaryMatchKey> TernaryLookupStructure;
+
+//! This class is used by match units to create instances of the appropriate
+//! LookupStructure implementation. In order to use custom data structures in
+//! a target, the data structures should be defined, and then a new subclass
+//! of LookupStructureFactory should be created, overriding the
+//! `create_for_<match type>` function or functions corresponding to the new
+//! data structure.
+class LookupStructureFactory {
+ public:
+  virtual ~LookupStructureFactory() = default;
+
+  //! This is a utility to call the correct `create_for_<type>` function based
+  //! on the bm::MatchKey subtype passed as the template parameter K. This is
+  //! used by bm::MatchUnitGeneric when creating its lookup structure.
+  template <typename K>
+  static std::unique_ptr<LookupStructure<K> > create(
+      LookupStructureFactory *f, size_t size, size_t nbytes_key);
+
+  //! Create a lookup structure for exact matches.
+  virtual std::unique_ptr<ExactLookupStructure>
+  create_for_exact(size_t size, size_t nbytes_key);
+
+  //! Create a lookup structure for LPM matches.
+  virtual std::unique_ptr<LPMLookupStructure>
+  create_for_LPM(size_t size, size_t nbytes_key);
+
+  //! Create a lookup structure for ternary matches.
+  virtual std::unique_ptr<TernaryLookupStructure>
+  create_for_ternary(size_t size, size_t nbytes_key);
+};
+
+
+}  // namespace bm
+
+
+#endif  // BM_SIM_INCLUDE_BM_SIM_LOOKUP_STRUCTURES_H_
--- a/modules/bm_sim/include/bm_sim/match_key_types.h
+++ b/modules/bm_sim/include/bm_sim/match_key_types.h
@@ -1,0 +1,81 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Gordon Bailey (gb@gordonbailey.net)
+ *
+ */
+
+#ifndef BM_SIM_INCLUDE_BM_SIM_MATCH_KEY_TYPES_H_
+#define BM_SIM_INCLUDE_BM_SIM_MATCH_KEY_TYPES_H_
+
+#include <limits>
+
+#include "bytecontainer.h"
+
+namespace bm {
+
+typedef uintptr_t internal_handle_t;
+typedef uint64_t entry_handle_t;
+
+enum class MatchUnitType {
+  EXACT, LPM, TERNARY
+};
+
+// Entry types.
+struct MatchKey {
+  MatchKey() {}
+  MatchKey(ByteContainer data, uint32_t version)
+    : data(std::move(data)), version(version) {}
+
+  ByteContainer data{};
+  uint32_t version{0};
+};
+
+struct ExactMatchKey : public MatchKey {
+  using MatchKey::MatchKey;
+
+  static constexpr MatchUnitType mut = MatchUnitType::EXACT;
+};
+
+struct LPMMatchKey : public MatchKey {
+  LPMMatchKey() {}
+  LPMMatchKey(ByteContainer data, int prefix_length, uint32_t version)
+    : MatchKey(data, version), prefix_length(prefix_length) {}
+
+  int prefix_length{0};
+
+  static constexpr MatchUnitType mut = MatchUnitType::LPM;
+};
+
+struct TernaryMatchKey : public MatchKey {
+  TernaryMatchKey() {}
+  TernaryMatchKey(ByteContainer data, ByteContainer mask, int priority,
+      uint32_t version)
+    : MatchKey(data, version), mask(std::move(mask)),
+    priority(priority) {}
+
+  ByteContainer mask{};
+  // This is initialized to `max` because lookups search for the matching
+  // key with the minimum priority, this ensures that default constructed
+  // TernaryMatchKey's will never be matched.
+  int priority{std::numeric_limits<decltype(priority)>::max()};
+
+  static constexpr MatchUnitType mut = MatchUnitType::TERNARY;
+};
+
+}  // namespace bm
+
+#endif  // BM_SIM_INCLUDE_BM_SIM_MATCH_KEY_TYPES_H_
--- a/modules/bm_sim/include/bm_sim/match_tables.h
+++ b/modules/bm_sim/include/bm_sim/match_tables.h
@@ -31,6 +31,7 @@
 #include "ras.h"
 #include "calculations.h"
 #include "control_flow.h"
+#include "lookup_structures.h"
 
 namespace bm {
 
@@ -232,6 +233,7 @@ class MatchTable : public MatchTableAbstract {
       const std::string &name,
       p4object_id_t id,
       size_t size, const MatchKeyBuilder &match_key_builder,
+      LookupStructureFactory *lookup_factory,
       bool with_counters, bool with_ageing);
 
  private:
@@ -391,6 +393,7 @@ class MatchTableIndirect : public MatchTableAbstract {
     const std::string &match_type,
     const std::string &name, p4object_id_t id,
     size_t size, const MatchKeyBuilder &match_key_builder,
+    LookupStructureFactory *lookup_factory,
     bool with_counters, bool with_ageing);
 
  protected:
@@ -483,6 +486,7 @@ class MatchTableIndirectWS : public MatchTableIndirect {
     const std::string &match_type,
     const std::string &name, p4object_id_t id,
     size_t size, const MatchKeyBuilder &match_key_builder,
+    LookupStructureFactory *lookup_factory,
     bool with_counters, bool with_ageing);
 
  protected:
--- a/modules/bm_sim/include/bm_sim/match_units.h
+++ b/modules/bm_sim/include/bm_sim/match_units.h
@@ -30,20 +30,19 @@
 #include <iostream>
 #include <atomic>
 #include <utility>  // for pair<>
+#include <memory>
 
+#include "match_key_types.h"
 #include "match_error_codes.h"
+#include "lookup_structures.h"
 #include "bytecontainer.h"
 #include "phv.h"
 #include "packet.h"
 #include "handle_mgr.h"
-#include "lpm_trie.h"
 #include "counters.h"
 #include "meters.h"
 
 namespace bm {
-
-typedef uintptr_t internal_handle_t;
-typedef uint64_t entry_handle_t;
 
 // using string and not ByteContainer for efficiency
 struct MatchKeyParam {
@@ -73,10 +72,6 @@ struct MatchKeyParam {
   std::string key;
   std::string mask{};  // optional
   int prefix_length{0};  // optional
-};
-
-enum class MatchUnitType {
-  EXACT, LPM, TERNARY
 };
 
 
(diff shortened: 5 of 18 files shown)


antoninbas:
Hi Gordon,

Thanks for the PR. It will take some time for me to review it and I have a busy week (I am moving), so expect my comments early next week. I'll probably have a bunch of questions :)
Could you remind me if you or your organization have signed the Apache CLA?

grodtron:
Hi Antonin,

Great! I did sign the CLA and send it to Prem, so hopefully all that should be fine. I'll look forward to your questions, and in the meantime good luck with the move :)

grodtron:
Hi @antoninbas , I just wanted to check if you've had the chance to review this PR? Thanks for your time!

modules/bm_sim/include/bm_sim/P4Objects.h
```diff
@@ -61,7 +61,9 @@ class P4Objects {
   explicit P4Objects(std::ostream &outstream = std::cout)
       : outstream(outstream) { }
 
-  int init_objects(std::istream *is, int device_id = 0, size_t cxt_id = 0,
+  int init_objects(std::istream *is,
+                   LookupStructureFactory * lookup_factory,
```
antoninbas: I'd rather have the lookup_factory be a static member of `MatchUnitGeneric`. The use case I see is that each bmv2 target will specify its own factory once (or use the default one) and passing this pointer around is just making things more complicated. Do you see any drawback with that?

modules/bm_sim/include/bm_sim/lpm_trie.h
```diff
@@ -1,91 +0,0 @@
-/* Copyright 2013-present Barefoot Networks, Inc.
```
antoninbas: Please leave this file as it is (it is just a C++ wrapper around some C code I wrote a while ago) and use LPMTrie in your LPM lookup structure implementation.

modules/bm_sim/include/bm_sim/match_tables.h
```diff
@@ -232,6 +233,7 @@ class MatchTable : public MatchTableAbstract {
       const std::string &name,
       p4object_id_t id,
       size_t size, const MatchKeyBuilder &match_key_builder,
+      LookupStructureFactory * lookup_factory,
```
antoninbas: I believe this will go away if the lookup_factory is made a static member of `MatchUnitGeneric` as I suggested above.

modules/bm_sim/include/bm_sim/match_unit_types.h
```diff
@@ -0,0 +1,76 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
```
antoninbas: This file should be called `match_key_types.h`

modules/bm_sim/include/bm_sim/match_unit_types.h
```diff
@@ -0,0 +1,76 @@
...
+typedef uintptr_t internal_handle_t;
+typedef uint64_t entry_handle_t;
+
+enum class MatchUnitType {
+  EXACT, LPM, TERNARY
+};
+
+// Entry types.
+struct MatchKey {
+  inline MatchKey() {}
```
antoninbas: Please do not use `inline`

modules/bm_sim/include/bm_sim/P4Objects.h
grodtron: I do agree it would be simpler that way, the rationale for doing it this way was to leave it open to the possibility of having multiple p4 targets in the same binary.

For example, we're planning to simulate multiple switches in one binary (a network containing both RMT and NPU architectures, for example), so for that use case we would not be able to use a static factory.

modules/bm_sim/include/bm_sim/lookup_structures.h
```diff
@@ -0,0 +1,90 @@
...
+  //! Remove a given entry from the structure. Has no effect if the entry
+  //! does not exist.
+  virtual void delete_entry(const K & key) = 0;
+
+  //! Completely remove all entries from the data structure.
+  virtual void clear() = 0;
+};
+
+template <class K>
+class LookupStructureFactoryPart {
```
antoninbas: I do not see the point of this helper class. All 3 `create` methods are implemented in `LookupStructureFactory`, so I do not believe this class serves any purpose.

On a related note, I am not a big fan of the signature of the `create` methods. I would much rather have them return a `std::unique_ptr<LookupStructure<...>>` rather than have them take a pointer to a `unique_ptr`. It's much more natural to me. However that would require all 3 functions to have a different name because there would be no overloading any more. Which would require a special dispatcher class when invoking the method from `MatchUnitGeneric`, so too much complexity...
In any case, let me know if you don't think we can get rid of `LookupStructureFactoryPart`

modules/bm_sim/include/bm_sim/lookup_structures.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "match_unit_types.h"
+#include "bytecontainer.h"
+
+namespace bm {
+
+//! This class defines an interface for all data structures used
+//! in Match Units to perform lookups. Custom data strucures can
+//! be created by implementing this interface, and creating a
+//! factory class which will create them.
```
antoninbas: An example for the doxygen documentation (maybe just overwriting one of the `create` methods) would be nice.

modules/bm_sim/include/bm_sim/lookup_structures.h
```diff
@@ -0,0 +1,90 @@
...
+};
+
+class LookupStructureFactory
+  // Inherits the virtual method to create each type of lookup structure
+  // from each of these base classes
+  : public LookupStructureFactoryPart<ExactMatchKey>
+  , public LookupStructureFactoryPart<LPMMatchKey>
+  , public LookupStructureFactoryPart<TernaryMatchKey> {
+ public:
+  void create(std::unique_ptr< LookupStructure<ExactMatchKey>> *,
```
antoninbas: be consistent with the rest of the code (I am putting this comment here but this applies to other places):
- `void *a` (notice the position of the star)
- when parameters / arguments overflow to the next line, align them with the previous line:
```
void my_fn(int x,
                  int y);
```
- no space after `std::unique_ptr<`

modules/bm_sim/include/bm_sim/lookup_structures.h
```diff
@@ -0,0 +1,90 @@
...
+
+  //! Check whether an entry exists. This is distinct from a lookup operation
+  //! in that this will also match against the prefix length in the case of
+  //! an LPM structure, and against the mask and priority in the case of a
+  //! Ternary structure.
+  virtual bool entry_exists(const K & key) = 0;
+
+  //! Store an entry in the lookup structure. Associates the given handle
+  //! with the given entry.
+  virtual void store_entry(const K & key, internal_handle_t handle) = 0;
```
antoninbas: can we rename this `add_entry` for consistency?

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,247 @@
...
+    entries_map.erase(key.data);
+  }
+
+  void clear() override {
+    entries_map.clear();
+  }
+
+ private:
+  // TODO(gordon) Should this be internal_handle_t ?
+  std::unordered_map<ByteContainer, entry_handle_t, ByteContainerKeyHash>
```
antoninbas: yes I believed this should be `internal_handle_t` and that there was a typo in my original code

modules/bm_sim/include/bm_sim/lookup_structures.h
```diff
@@ -0,0 +1,90 @@
...
+//! factory class which will create them.
+template <typename K>
+class LookupStructure {
+ public:
+  virtual ~LookupStructure() = default;
+
+  //! Look up a given key in the data structure.
+  //! Return true and set handle to the found value if there is a match
+  //! Return false if there is not match (value in handle is undefined)
+  virtual bool lookup(const ByteContainer & key_data,
```
antoninbas: `lookup` should be `const` or there is something wrong

modules/bm_sim/include/bm_sim/lookup_structures.h
```diff
@@ -0,0 +1,90 @@
...
+      internal_handle_t * handle) = 0;
+
+  //! Check whether an entry exists. This is distinct from a lookup operation
+  //! in that this will also match against the prefix length in the case of
+  //! an LPM structure, and against the mask and priority in the case of a
+  //! Ternary structure.
+  virtual bool entry_exists(const K & key) = 0;
```
antoninbas: should be `const` as well

modules/bm_sim/include/bm_sim/match_units.h
```diff
@@ -435,126 +421,21 @@ class MatchUnitExact : public MatchUnitAbstract<V> {
 
  private:
   std::vector<Entry> entries{};
-  std::unordered_map<ByteContainer, entry_handle_t, ByteContainerKeyHash>
-  entries_map{};
+  std::unique_ptr<LookupStructure<K>> lookupStructure;
```
antoninbas: initialize to `nullptr`

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,247 @@
...
+    auto it = find_handle(key);
+    this->handles.erase(it);
+  }
+
+  void clear() override {
+    this->handles.clear();
+  }
+
+ private:
+  std::vector<std::tuple<TernaryMatchKey, internal_handle_t>> handles;
```
antoninbas: the internal handle can be used as and index into the vector (this is actually the difference with the regular client-facing handle, which includes a version number); so I would just use that to make the code simpler. Don't forget to initialize your vector to `size` elements.
`handles` is a bad name for this (especially after removing the handles)

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,247 @@
...
+       std::unique_ptr<LookupStructure<ExactMatchKey>> * ls,
+       size_t size, size_t nbytes_key) {
+  (void) nbytes_key;
+  ls->reset(new ExactMap(size));
+}
+
+void LookupStructureFactory::create(
+       std::unique_ptr<LookupStructure<LPMMatchKey>> * ls,
+       size_t size, size_t nbytes_key) {
+  (void)size;
```
antoninbas: missing space

modules/bm_sim/src/match_units.cpp
```diff
@@ -296,11 +298,11 @@ format_ternary_key(ByteContainer *key, const ByteContainer &mask) {
...
 class MatchKeyBuilderHelper {
  public:
-  template <typename E,
-            typename std::enable_if<E::mut == MatchUnitType::EXACT, int>::type
+  template <typename K,
+            typename std::enable_if<K::mut == MatchUnitType::EXACT, int>::type
             = 0>
   static std::vector<MatchKeyParam>
-  entry_to_match_params(const MatchKeyBuilder &kb, const E &entry) {
+  entry_to_match_params(const MatchKeyBuilder &kb, const K & key) {
```
antoninbas: `K &key`

modules/bm_sim/include/bm_sim/switch.h
```diff
@@ -551,6 +552,10 @@ class SwitchWContexts : public DevMgr, public RuntimeInterface {
     return contexts.at(cxt_id).add_component<T>(ptr);
   }
 
+  void set_lookup_factory(LookupStructureFactory * new_factory) {
```
antoninbas: let's make this a smart pointer (shared or unique), since this is a client-facing API.

modules/bm_sim/include/bm_sim/P4Objects.h
antoninbas: That is a good point. But a bunch of code actually relies on the fact that there is only one `Switch` instance per binary. For example, action primitive registration, which lets you register target-specific primitives.
However, I do believe removing all static elements is a worthy goal. I think we can leave it as it is in your Pull Request. If more elements become customizable, we should probably bundle them in a `SwitchConfiguration` object or something (just made up that class), which would include your lookup factory. This way we would avoid ending up with huge signatures for methods like `init_objects`.

antoninbas:
Hi @grodtron 
Thanks a lot for the Pull Request and I apologize for the delay in reviewing it.
I wrote a few comments, please take them into account or let me know if you disagree.
There are a lot of tiny formatting issues, please address them as well for consistency sake. I am thinking about parameters overflowing to the next line, spacing when using `&` or `*` (the code base uses `int *x` and `int &x`),...
I will review the code again after you address my comments and before merging.
One last thing: after the review is over, I will ask you to squash all your commits into one, such that your branch is one commit ahead of master. This is to keep the repo history cleaner and easier to navigate.

modules/bm_sim/include/bm_sim/lpm_trie.h
grodtron: Done in 0a0908c75b1927f34e1f3e7c4e97d1a8c0cf2233

modules/bm_sim/include/bm_sim/lookup_structures.h
grodtron: Hmm, yes I see your point, as it stands there actually is no reason for `LookupStructureFactoryPart` to exist. I also agree with you that the signature of `create` is quite awkward. 

The reason I made `LookupStructureFactoryPart` was to allow a virtual templated member function, but then with the signature I chose there's no need for it to be a template function, a simple overload in `LookupStructureFactory` would do the job.

I think the best solution would be to change the signature of create in `LookupStructureFactoryPart` to:

``` cpp
virtual std::unique_ptr<LookupStructure<K>>
create(size_t size, size_t nbytes_key) = 0;
```

The advantage of this is that the interface is much more natural, but also that it remains simple at the call site, (`match_units.h:393`) where it becomes:

``` cpp
lookupStructure = lookup_factory->create<K>( ... );
```

In this case all of the complexity is contained in `lookup_structures.h`, users overriding `LookupStructureFactory` can do so without knowledge of `LookupStructureFactoryPart` and `MatchUnitGeneric` can also use `create` in a very natural way. (If we choose this option I'll be sure to add some additional documentation to `LookupStructureFactoryPart` explaining it.)

The alternative would be removing `LookupStructureFactoryPart` and keeping the signature the same, since in that case passing the `std::unique_ptr<LookupStructure<...>>` as a parameter is necessary in order to resolve the overload. Personally I think this is the worse option, as it is much less natural for users creating their own factories as well as at the call site itself.

What do you think?

modules/bm_sim/include/bm_sim/P4Objects.h
grodtron: Great! Commenting for myself to mark this as resolved :)

modules/bm_sim/include/bm_sim/match_tables.h
grodtron: (resolved above)

modules/bm_sim/include/bm_sim/lookup_structures.h
antoninbas: Yes, I would like a solution where each `create` method returns a `std::unique_ptr<...>` but which can still be called easily from `MatchUnitGeneric`. It is fine with me if the implementation details of `LookupStructureFactoryPart` become more complex. If you have something in mind, please update your pull request and I will look at it.

grodtron:
Hi @antoninbas ,

Thanks for your feedback. I'm still working on addressing all of it, especially correcting all of the style issues you highlighted. In the meantime I'd be interested in your feedback specifically on the approach I've settled on for `LookupStructureFactory`.

The solution that I came up with is unfortunately not as clean as I had hoped. The original signature of `create` was done the way it was because otherwise from the point of view of the compiler the signatures of the 3 `create` functions are the same.

The way I've chosen to circumvent this problem is by changing the signature in `LookupStructureFactoryPart` to be:

``` cpp
virtual std::unique_ptr<LookupStructure<K>> 
create (size_t size, size_t nbytes_key, LookupStructure<K> *dummy = nullptr) = 0
```

The addition on the `dummy` argument allows the compiler to see the 3 different overloads as having unique signatures, and is not unprecedented in c++ (`operator++()` and `operator++(int)` for example).

Since the static type of the factory at the call site is  `LookupStructureFactoryPart<K>`, the default value for `dummy` will be used, and from the caller's point of view nothing seems unusual. From the implementer's point of view, it is necessary to add `LookupStructure<K> *` to the end of the argument list, but that is the only major difference.

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,252 @@
...
+    //
+    auto c_it = find_handle(key);
+
+    // This advances a non-const iterator to the position pointed by the
+    // const iterator, effectively removing the const-ness of the iterator
+    auto it = this->handles.begin();
+    std::advance(it, std::distance(this->handles.cbegin(), c_it));
+#else
+    auto it = find_handle(key);
+#endif
```
grodtron: As the comment says, this is a workaround for an issue in `g++-4.8`. I've also tested it on 4.9 and it seems to work, but maybe it's worth also adding 4.9 to the travis build? As far as I can tell it's the default `g++` on the latest Debian (Jessie)

antoninbas:
Hi @grodtron 

Regarding your dummy pointer approach, not a big fan of the dummy pointer since it requires the target implementer to include it. We have been trying to design an approach where the name `create` is the same for all methods. However, I don't think this is necessary. Maybe it is even better to have different method names for different key types, _so long as there is a unified way of calling the methods from `MatchUnitGeneric`_. So I came up with the following:

```
#include <memory>

template <typename K>
class LookupStructure { };

struct ExactMatchKey { };
struct LPMMatchKey { };

class LookupStructureFactory {
 public:
  // do not forget virtual destructor as target designer can subclass
  virtual ~LookupStructureFactory() { }

  virtual std::unique_ptr<LookupStructure<ExactMatchKey> >
  create_for_exact(size_t size, size_t nbytes_key) {
    (void) size; (void) nbytes_key;
    return nullptr;
  }

  virtual std::unique_ptr<LookupStructure<LPMMatchKey> >
  create_for_LPM(size_t size, size_t nbytes_key) {
    (void) size; (void) nbytes_key;
    return nullptr;
  }
};

class LookupStructureFactoryDispatch {
 public:
  template <typename K>
  static std::unique_ptr<LookupStructure<K> > create(
      LookupStructureFactory *f, size_t size, size_t nbytes_key);
};

template <>
std::unique_ptr<LookupStructure<ExactMatchKey> >
LookupStructureFactoryDispatch::create<ExactMatchKey>(
    LookupStructureFactory *f, size_t size, size_t nbytes_key) {
  return f->create_for_exact(size, nbytes_key);
}

template <>
std::unique_ptr<LookupStructure<LPMMatchKey> >
LookupStructureFactoryDispatch::create<LPMMatchKey>(
    LookupStructureFactory *f, size_t size, size_t nbytes_key) {
  return f->create_for_LPM(size, nbytes_key);
}

class BadMatchKey { };

int main() {
  auto f = std::unique_ptr<LookupStructureFactory>(
      new LookupStructureFactory());
  auto p1 = LookupStructureFactoryDispatch::create<ExactMatchKey>(
      f.get(), 9u, 9u);
  auto p2 = LookupStructureFactoryDispatch::create<LPMMatchKey>(
      f.get(), 9u, 9u);
  // compile error
  // auto p3 = LookupStructureFactoryDispatch::create<BadMatchKey>(
  //     f.get(), 9u, 9u);
}
```
I omitted `TernaryMatchKey` on purpose to make it shorter. It is still pretty long so sorry about that. As you can tell from the `main` function, we can call create using the key type as the template parameter using the static methods in the dispatcher. The target implementer can override one or more of `create_for_exact`, `create_for_LPM`, `create_for_ternary`.
Let me know what you think.
You'll notice that i removed `LookupStructureFactoryPart`. I also made the `LookupStructureFactory` destructor virtual, which may be necessary if the target implementer wants to store some state in his custom factory.

grodtron:
Hi @antoninbas 

Hmm, yes you're right, I think your method is cleaner. I'll integrate that into the PR. In the meantime I believe I have addressed all of the other issues that you raised with my latest push. The big change is the implementation of `TernaryMap` I believe I understood what you meant in your comment, but maybe you could verify that you agree with my implementation.

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,244 @@
...
+    , nbytes_key(nbytes_key) {}
+
+  bool lookup(const ByteContainer &key_data,
+              internal_handle_t *handle) const override {
+    auto min_priority = std::numeric_limits<
+                          decltype(TernaryMatchKey::priority)>::max();
+    bool match;
+
+    const TernaryMatchKey *min_key = nullptr;
+    entry_handle_t min_handle = 0;
```
grodtron: should be `internal_handle_t`, I'll fix this in the next update

grodtron:
Hi @antoninbas 

I believe with my most recent push I have addressed all of your comments from the original PR. Looking forward to your feedback, thanks for your time!

modules/bm_sim/include/bm_sim/match_key_types.h
```diff
@@ -0,0 +1,78 @@
...
+
+struct TernaryMatchKey : public MatchKey {
+  TernaryMatchKey() {}
+  TernaryMatchKey(ByteContainer data, ByteContainer mask, int priority,
+      uint32_t version)
+    : MatchKey(data, version), mask(std::move(mask)),
+    priority(priority) {}
+
+  ByteContainer mask{};
+  int priority{std::numeric_limits<decltype(priority)>::max()};
```
antoninbas: this can just be 0, no?

grodtron: The reason for this is that in the lookup function it looks for the match with the minimum priority, starting from `numeric_limits<>::max()`, this ensures that default-constructed keys will never match. I can add a comment to that effect

modules/bm_sim/include/bm_sim/match_units.h
```diff
@@ -377,94 +372,29 @@ class MatchUnitAbstract : public MatchUnitAbstract_ {
...
-    ByteContainer key{};
-    int prefix_length{0};
-    V value{};
-    uint32_t version{0};
-
-    static constexpr MatchUnitType mut = MatchUnitType::LPM;
-  };
+  MatchUnitGeneric(size_t size, const MatchKeyBuilder &match_key_builder,
+                   LookupStructureFactory *lookup_factory)
+    : MatchUnitAbstract<V>(size, match_key_builder)
```
antoninbas: style (comma at the end of the line, arg alignment for call to `LookupStructureFactory::create<K>`)

modules/bm_sim/include/bm_sim/match_key_types.h
antoninbas: ok, works for me

modules/bm_sim/include/bm_sim/switch.h
```diff
@@ -568,13 +569,33 @@ class SwitchWContexts : public DevMgr, public RuntimeInterface {
...
+
  private:
   size_t nb_cxts{};
   // TODO(antonin)
   // Context is not-movable, but is default-constructible, so I can put it in a
   // std::vector
   std::vector<Context> contexts{};
 
+  LookupStructureFactory *get_lookup_factory() const {
+    if (lookup_factory.get()) {
```
antoninbas: no need for the `.get()` I believe

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,263 @@
...
+ private:
+  std::unordered_map<ByteContainer, internal_handle_t, ByteContainerKeyHash>
+    entries_map{};
+};
+
+class TernaryMap : public TernaryLookupStructure {
+ public:
+  explicit TernaryMap(size_t size, size_t nbytes_key)
+    : keys(size)
+    , nbytes_key(nbytes_key) {}
```
antoninbas: style:
`keys(size), nbytes_key(nbytes_key) {}`

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,263 @@
...
+    auto handle = find_handle(key);
+
+    return handle != keys.size();
+  }
+
+  void add_entry(const TernaryMatchKey &key,
+                 internal_handle_t handle) override {
+    assert(handle < keys.size());
+
+    keys[handle] = key;
```
antoninbas: replace with `keys.at(handle) = key`, remove the `assert` (`at` throws an out-of-bound exception)

modules/bm_sim/include/bm_sim/match_units.h
grodtron: Like this? (The `{}` on the following line is to make the above line < 80 chars)
```cppp
MatchUnitGeneric(size_t size, const MatchKeyBuilder &match_key_builder,
                   LookupStructureFactory *lookup_factory)
    : MatchUnitAbstract<V>(size, match_key_builder), entries(size),
      lookup_structure(
        LookupStructureFactory::create<K>(lookup_factory,
                                          size,
                                          match_key_builder.get_nbytes_key()))
  {}
```

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  void delete_entry(const TernaryMatchKey &key) override {
+    auto handle = find_handle(key);
+
```
antoninbas: can you put an `assert(handle != keys.end())` here; as well as a comment explaining that this cannot happen because this function is only called if the handle is valid

modules/bm_sim/src/lookup_structures.cpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  void add_entry(const TernaryMatchKey &key,
+                 internal_handle_t handle) override {
+    assert(handle < keys.size());
+
+    keys[handle] = key;
+  }
+
+  void delete_entry(const TernaryMatchKey &key) override {
```
antoninbas: It's a bit sad that we have to go through the whole array to find the handle, especially since we are holding a write lock; we can see if this is a problem later on

modules/bm_sim/src/switch.cpp
```diff
@@ -81,11 +83,13 @@ SwitchWContexts::init_objects(const std::string &json_path, int dev_id,
     notifications_transport = std::move(transport);
   }
 
+
```
antoninbas: remove extra line

modules/bm_sim/src/lookup_structures.cpp
grodtron: I agree - one thought I had as I was writing this was to keep the keys sorted by priority so that we can at least return on the first match instead of having to scan the entire table even after having matched. This would then require storing the handles along with the keys as they would no longer work as indices.

In any case, and as you said, I figured I'd leave it as it was for now.

modules/bm_sim/include/bm_sim/match_units.h
antoninbas: you can also put the args on the next line:
```
lookup_structure(
  LookupStructureFactory::create<K>(
    lookup_factory, size, match_key_builder.get_nbytes_key())) {}
```

antoninbas:
I am done with comments (for good this time :)). Thanks again for the great job, I had been willing to unify the match unit code for a while but lacked time, and now it is done :)
When you are done with addressing the comments, please squash everything into one commit (as explained in a previous comment). I will then run valgrind once locally on the tests (you may have missed the valgrind script under `tools/`?) and merge.

grodtron:
Great sounds good! I'll run valgrind on my side too (I had missed it), I'll push a squashed version of the whole thing in however long my current build + tests take to run :)

grodtron:
Hi @antoninbas 

I've fixed those last little issues and squashed everything together. Thanks for all the feedback and have a good weekend!

modules/bm_sim/src/lookup_structures.cpp
grodtron: Ended up using `.at` because `handle

PR #102, opened 2016-04-12
automatically masking first byte of each match key param

antoninbas:
- in case user input not correctly formatted
- would it be better to return an error to the client / application?

### DIFF
--- a/modules/bm_sim/src/match_units.cpp
+++ b/modules/bm_sim/src/match_units.cpp
@@ -294,6 +294,12 @@ format_ternary_key(ByteContainer *key, const ByteContainer &mask) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


vgurevich:
Hello Antonin,

I believe that returning an error would actually be a better approach in
the future. In general, there are a lot of malformed messages that can be
passed (accidentally or maliciously) into BMv2 PI interface, so it needs to
be "vigilant".  I also see that in the pd code that the compiler generates
you already use try/catch around the calls to pd_conn_mgr_client(), thus it
looks like we are lready set up to handle error returns from BMv2.

I would vote for strict checking on BMv2 side and returning errors, rather
than trying to guess what's going on.

Just my .02

Thanks,
Vladimir

Issue #92, opened 2016-03-20
simple_switch_CLI cannot be properly executed when a non-standard prefix is specified during ./configure


vgurevich:
I tried to run ./configure with a non-standard prefix, e.g.

./configure prefix=/tmp/my_install

While make, "make check" and "make install" succeeded, simple_switch_CLI could not find runtime_CLI python module. 

The reason is, runtime_CLY.py* got installed in $(prefix)/lib/python2.7/site-packages, but then this directory is not configured to become a part of PYTHONPATH

antoninbas:
Fixed by b77e968ab6dc1853d20182b90bd78e5430c9e89a

antoninbas:
I incorrectly labeled this as a bug.
This is not a bug. If you use a custom prefix, it is your responsibility to update the `PYTHONPATH`, just like it is your responsibility to update the `PATH`.

vgurevich:
Hi Antonin,

I do not think this should be the customer's responsibility. The customer
should not care which language a program is written and what are the
internal pre-requisites. If "make install" succeeded, the expectation is
that no matter where the program was executed from (it doesn't have to be
in the PATH, one can simply specify full PATH to run it), it will run
correctly and find all its pre-requisites.

Thanks,
Vladimir

antoninbas:
solved in #104

PR #104, opened 2016-04-14
added  to Python sys path in scripts (e.g. bm_CLI) so that they can f…

antoninbas:
…ind installed Python modules


antoninbas:
addresses #92

vgurevich:
+1

Thank you so much, Antonin!

PR #105, opened 2016-04-15
modified implementation of the ternary lookup structure

antoninbas:
- old implementation was looping over the whole array instead of only "occupied slots"
- no more copy of the entry in the lookup structure (using pointer instead)


antoninbas:
@grodtron FYI

PR #137, opened 2016-06-02
if a field alias and an actual field have the same name, give priorit…

antoninbas:
…y to alias


antoninbas:
Travis build still not working because of LLVM apt repo, so bypassing tests

PR #138, opened 2016-06-02
moved CLI tests to simple_switch tests, because they use simple_switch

antoninbas:
running `make check` without running `make` first was failing


antoninbas:
LLVM apt repo still down, bypassing Travis tests

PR #139, opened 2016-06-03
small fixes for compiling with clang

zarghul:
I tried to compile with clang 3.8, and it spotted some small problems:

* some unnecessary `std::move`
* missing headers


antoninbas:
Hi Yuri,

The `tools/check_style.sh` script is failing. Please run it locally and fix the error.

zarghul:
it should be ok now

PR #140, opened 2016-06-03
explicit python2 shebang in scripts

zarghul:
Explicitly calling python2 does not cause any harm to systems where python2 is default, and it makes the scripts work out of the box on systems where python3 is default (e.g. Arch Linux)

### DIFF
--- a/mininet/1sw_demo.py
+++ b/mininet/1sw_demo.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
 
 # Copyright 2013-present Barefoot Networks, Inc.
 #
--- a/mininet/stress_test_ipv4.py.in
+++ b/mininet/stress_test_ipv4.py.in
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
 
 # Copyright 2013-present Barefoot Networks, Inc.
 #
--- a/targets/simple_switch/simple_switch_CLI.in
+++ b/targets/simple_switch/simple_switch_CLI.in
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 # Copyright 2013-present Barefoot Networks, Inc.
 #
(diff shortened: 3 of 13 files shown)


antoninbas:
Hi Yuri,

Is there a reason why your Pull Request is limited to these scripts. When I run `git grep "/usr/bin/env python"`, I get the following list:
```
targets/simple_switch/simple_switch_CLI.in:#!/usr/bin/env python
targets/simple_switch/sswitch_CLI.py:#!/usr/bin/env python
tests/stress_tests/gen_udp_tcp_traffic.py:#!/usr/bin/env python
tools/bm_CLI.in:#!/usr/bin/env python
tools/bm_nanomsg_events.in:#!/usr/bin/env python
tools/bm_p4dbg.in:#!/usr/bin/env python
tools/bmpy_utils.py:#!/usr/bin/env python
tools/cpplint.py:#!/usr/bin/env python
tools/nanomsg_client.py:#!/usr/bin/env python
tools/p4dbg.py:#!/usr/bin/env python
tools/runtime_CLI.py:#!/usr/bin/env python
```

zarghul:
You are right, i will add them to the pull request.

The reason is that my python 3 never complained about them, probably
because some of them are still valid in python 3, and I did not think of
finding them all.

Thank you for the feedback!

antoninbas:
Most of them are wrappers that are only used when installed on your system. For example `tools/bm_CLI.in` is a wrapper around the runtime CLI. So you may not have tried to run it. It may be safer to replace the shebang everywhere and worry about python3 portability later. Except maybe `tools/cpplint.py` which is a 3rd party script from Google.

zarghul:
for `tools/cpplint.py` a solution may be to change `tools/check_style.sh` such that it explicitly calls `python2`.

antoninbas:
Cool,
Do you mind running the following:
```
git fetch --all
git rebase [whatever the bmv2 origin is in your repo] master
git push -f origin python2-fix
```
This is because I have just removed the clang build from the Travis regression (while the LLVM apt repo is down). Doing the above will trigger the Travis regression to run again. You will have to do the same in your other pull request.

PR #154, opened 2016-06-21
changes to travis deps scripts

antoninbas:
- switch to nanomsg production release 1.0.0; added `cmake` as a dependency as it is required by nanomsg
- renamed `build/travis` directory to `travis` (`build` was a bad name...)
- detecting if `libthrift` already installed in `install-thrift.sh`, although not sure how useful / reliable it is. Trying to do the same for `libnanomsg`, although the difference between `SOVERSION` and the package `VERSION` may make it even more brittle.
- packaging the `travis` directory (when running `make dist`)

- updating the `libnanomsg` version seems to have exposed a bug in the CLI tests crip `run_one_test.py`. When parallel tests are run, each switch needs to get its own device id, so that a different nanomsg IPC socket is used (otherwise multiple `bind` calls for same address)


antoninbas:
bypassing sporadic fail of simple-switch tests, it is an issue with the test code, not bmv2 (problem for another day)

PR #185, opened 2016-09-03
fixed race condition in parse vset shadow copies

antoninbas:
updated README to indicate that parse value sets are no longer missing


ckbf:
Antonin -

Can I do a mask on a parser value set when I match? This is not in the spec, but I think we would need this especially for Tofino.

Thx,
- cK

PR #196, opened 2016-09-21
trying to fix coverage

antoninbas:
recently coverage reports started including code under `third_party`, despite the use of `--relative-only` with `gcov`; so I'm trying to fix that by manually deleting the reports for these files before sending them to codecov


antoninbas:
it seems to be working so I am merging; I'm still unsure about why this is needed though

PR #203, opened 2016-09-28
-pthread flag needed for compilation

antoninbas:


antoninbas:
@cc10512 
After merging your pull request, the build fails for me on Linux (linker errors because of `pthread_*` symbols not found). This is not showing in the Travis regression (because for some reason `"-lssl -lcrypto -lrt -lpthread"` are added when linking; I think it is because of how Thrift is installed).
Anyway when using `AX_PTHREAD`, the `PTHREAD_CFLAGS` still have to be added to the automake `CFLAGS`, which I am doing in this pull request (see https://www.gnu.org/software/autoconf-archive/ax_pthread.html). I am curious to know if this breaks the build on your mac machine. Hopefully not, as the `AX_PTHREAD` macro is supposed to add the appropriate flag to disable the clang warning. I'll have to merge this pull request as soon as the Travis regressions pass, because I expect that right now the build is broken for a bunch of users.

cc10512:
please merge, i'll cherry pick your commit now and see if it works.

cc10512:
@antoninbas: +1 on the mac.
It adds '-Qunused-arguments -pthread' which removes the warning for unused arguments.

antoninbas:
@cc10512 great that's the expected behavior of the macro, I'm happy it's working

PR #225, opened 2016-11-03
Fix modify_field_with_hash_based_offset implementation

BlakeLawson:
According to the [spec](http://p4.org/wp-content/uploads/2016/03/p4_v1.1.pdf) for modify_field_with_hashed_based_offset, 

> If size is not zero, the hash value is used to generate a value between base and
> (base + size - 1) by calculating (base + (hash_value % size)). If size is 0 then the
> value used is (base + hash_value).

Previously, the implementation of modify_field_with_hash_based_offset did not handle the case where size is 0, which caused the switch to crash when size was set to zero.

This commit fixes this bug.

### DIFF
--- a/targets/simple_switch/primitives.cpp
+++ b/targets/simple_switch/primitives.cpp
@@ -264,8 +264,13 @@ class modify_field_with_hash_based_offset
                            const NamedCalculation &, const Data &> {
   void operator ()(Data &dst, const Data &base,
                    const NamedCalculation &hash, const Data &size) {
-    uint64_t v =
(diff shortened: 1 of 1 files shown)


antoninbas:
Hi,
In the latest version of the P4_14 spec, the special case where `size == 0` has been removed. The spec says that `size` needs to be larger to 0, and does not specify the behavior when `size` is 0. I therefore consider it to be runtime exception if `size` is 0. We don't really use C++ exceptions in bmv2, but it may be interesting to have a special `p4_runtime_exception` in the C++ code, which could be used for cases like this one.

BlakeLawson:
Hi Antonin,

Thanks for the clarification.

PR #236, opened 2016-12-02
Arbitrary names for externs.

pierce-m:
Allows you to register externs with names different from the respective class name. This is useful for externs like "register", because "register" is a reserved word in C++.

### DIFF
--- a/include/bm/bm_sim/extern.h
+++ b/include/bm/bm_sim/extern.h
@@ -66,20 +66,22 @@ class ExternFactoryMap {
 #define BM_REGISTER_EXTERN(extern_name) \
   BM_REGISTER_EXTERN_W_NAME(extern_name, extern_name)
 
-#define BM_REGISTER_EXTERN_W_NAME_METHOD(extern_name, extern__, extern_method_name, ...) \
-  template <typename... Args>                                                            \
-  struct _##extern_name##_##extern_method_name##_0                                       \
-      : public ::bm::ActionPrimitive<::bm::ExternType *, Args...> {                      \
-    void operator ()(::bm::ExternType *instance, Args... args) override {                \
-      auto lock = instance->_unique_lock();                                              \
-      instance->_set_packet_ptr(&this->get_packet());                                    \
-      dynamic_cast<extern__ *>(instance)->extern_method_name(args...);                   \
-    }                                                                                    \
-  };                                                                                     \
-  struct _##extern_name##_##extern_method_name                                           \
-      : public _##extern_name##_##extern_method_name##_0<__VA_ARGS__> {};                \
-  REGISTER_PRIMITIVE_W_NAME(_BM_EXTERN_TO_STRING(_##extern_name##_##extern_method_name), \
-                            _##extern_name##_##extern_method_name)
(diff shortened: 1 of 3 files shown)


include/bm/bm_sim/extern.h
```diff
@@ -52,6 +52,8 @@ class ExternFactoryMap {
   std::unordered_map<std::string, ExternFactoryFn> factory_map{};
 };
 
+#define TO_STRING(name) #name
```
antoninbas: Please rename to something less generic to avoid name collisions, maybe `_BM_EXTERN_TO_STRING`

include/bm/bm_sim/extern.h
```diff
@@ -61,20 +63,44 @@ class ExternFactoryMap {
            #extern_name,                                                \
            [](){ return std::unique_ptr<ExternType>(new extern_name()); });
 
-#define BM_REGISTER_EXTERN_METHOD(extern_name, extern_method_name, ...) \
-  template <typename... Args>                                           \
-  struct _##extern_name##_##extern_method_name##_0                      \
-      : public ::bm::ActionPrimitive<::bm::ExternType *, Args...> {     \
+#define BM_REGISTER_EXTERN_W_NAME(extern_name, extern__)                \
```
antoninbas: Can we use `BM_REGISTER_EXTERN_W_NAME` to implement `BM_REGISTER_EXTERN` to avoid duplicating the code, or is there something preventing us from doing that? (same for method below)

pierce-m:
The motivation is is the description of the PR.

include/bm/bm_sim/extern.h
pierce-m: Okay.

antoninbas:
Hmm great, but that does not say why it's needed. If the name of your extern is a valid C++ identifier, then you can just use that. Which is why I asked which case you ran into that made you want to have that feature, where the name you use in the implementation is not the same as the P4 name.

pierce-m:
"register" is not a valid c++ identifier, because it's a reserved keyword
in c++. The p4 name of the extern is "register", but I can't name a class
"register" in c++. Am I understanding your concern correctly?

antoninbas:
I had missed the description, I later edited my comment above

pierce-m:
Right on.

pierce-m:
Would it be acceptable to use an existing extern type in the test file,
then simply register it with a different name using the new macros?

antoninbas:
Sure, that should not be an issue. Just make sure you don't register it twice with the same name.
Then you should write a test like this:
```
auto check_one_name = [](const std::string &name) {
  auto extern_instance = ExternFactoryMap::get_instance()->get_extern_instance(name);
  auto typed_instance = dynamic_cast<TYPE_YOU_CHOOSE *>(extern_instance.get());
  ASSERT_NE(nullptr, typed_instance);
  // next check method
  auto primitive = get_extern_primitive(name, "<YOUR_METHOD_NAME>");
  ASSERT_NE(nullptr, primitive);
};
check_one_name("name_1");
check_one_name("name_2");
```

pierce-m:
Hey Antonin, can you approve this? It passes everything and I added the
test you asked for. I "dismissed" the review because I thought that's what
I had to do to merge it, but I think I was wrong. Thanks.

PR #238, opened 2016-12-05
Data increment

pierce-m:
Not sure if this is really desired but it makes incrementing a Data a little nicer. Useful for implementing a counter.


antoninbas:
I am not sure this is very useful. For counters, you may not always want to increment by 1 (e.g. if you are counting bytes).
I would recommend just writing a nice increment method in your counter extern implementation:
```
void increment(size_t amount = 1) {
    counter_value.add(counter_value, Data(amount));
}
```

PR #240, opened 2016-12-06
Revert "Add ability to subscribe to group member updates"

antoninbas:
Reverts p4lang/behavioral-model#239

Realized this is not very useful to targets, will propose a better way.


pierce-m:
???

PR #254, opened 2017-01-05
Allow ActionPrimitive_ access to p4objects that created it.

pierce-m:
This is useful for the "hash" extern function found here: https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4#L99

Typically a tuple literal is supplied as the "data" parameter to this function, which causes the compiler to create a field_list out of the tuple literal and replace the tuple literal with the field_list's id. The hash primitive must then resolve this field_list id to the FieldList object in order to create a Calculation using the "algo" parameter. In needs access to P4Objects to do this.

### DIFF
--- a/src/bm_sim/P4Objects.cpp
+++ b/src/bm_sim/P4Objects.cpp
@@ -648,6 +648,8 @@ P4Objects::init_objects(std::istream *is,
         return 1;
       }
 
+      primitive->_set_p4objects(this);
+
       for (const auto &cfg_parameter : cfg_primitive_parameters) {
         const string type = cfg_parameter["type"].asString();
 
--- a/src/bm_sim/P4Objects.cpp
+++ b/src/bm_sim/P4Objects.cpp
@@ -647,8 +647,6 @@ P4Objects::init_objects(std::istream *is,
         return 1;
       }
 
-      primitive->_set_p4objects(this);
-
       for (const auto &cfg_parameter : cfg_primitive_parameters) {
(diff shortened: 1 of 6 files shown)


src/bm_sim/P4Objects.cpp
```diff
@@ -648,6 +648,8 @@ P4Objects::init_objects(std::istream *is,
         return 1;
       }
 
+      primitive->_set_p4objects(this);
```
antoninbas: this implementation is not correct. Currently a single action primitive instance is created, but if you need a reference to p4objects, then you need to have one instance per p4objects. You can look at what is done for externs, where registering an extern type actually registers a lambda factory function. Unlike for externs though, you need to instantiate only **one** action primitive per p4objects (not a new instance every time the primitive is called in an action!!!).
I suggest that you start by adding a unit test to test_actions which exposes the issue I am trying to describe here: register one primitive action that leverages the new feature you are adding, create 2 different `P4Objects` instances using that primitive inside an action's body, execute the primitives. For example if your primitive action looks up a field list in p4objects, you need to check that both primitives end up with a different field list instance, each in the appropriate p4object for the primitive (but my guess is that with the implementation here you are going to end up with the same).

pierce-m: Okay, I'll work on it. Thanks.

pierce-m:
Thanks, but it looks like I can't merge because I'm not authorized.

PR #268, opened 2017-01-19
Fix minor memory leak in P4Objects

antoninbas:
Only happens when the JSON input is invalid. It was exposed when running
test_p4objects under valgrind.


antoninbas:
failed status check is due to temporary travis issue, so merging

PR #276, opened 2017-01-24
Constants in fieldlists

pierce-m:


pierce-m:
Please don't merge this until I add the field width information, thanks.

include/bm/bm_sim/field_lists.h
```diff
@@ -46,23 +47,42 @@ class FieldList {
...
+      return !(*this == other);
+    }
+  };
+
  public:
-  using iterator = std::vector<field_t>::iterator;
-  using const_iterator = std::vector<field_t>::const_iterator;
-  using reference = std::vector<field_t>::reference;
-  using const_reference = std::vector<field_t>::const_reference;
+  using field_list_member_t = boost::variant<field_t, constant_t>;
```
antoninbas: I think the implementation choice (`boost::variant`) should be hidden from the `FieldList` public "interface" here. Instead you could provide a method which lets the target programmer apply a functor of his choice to members of the list. Something like this:
```
struct FieldListVisitor : boost::static_visitor<> {
  void operator ()(const field_t &field) { } // do nothing on purpose
  void operator ()(const constant_t &constant) { } // do nothing on purpose
};
class FieldList {
  // ...
  // public method that can be used by targets
  template <typename T>
  visit(const T &visitor) {
      static_assert(std::is_base_of(FieldListVisitor, T), "invalid visitor, must inherit from FieldListVisitor");
      std::for_each(fields.begin(), fields.end(), boost::apply_visitor(visitor));
  }
};
```

For convenience I think we can also provide a `FieldList::copy_fields_between_phvs` method (implemented with `FieldList::visit`) that could be used by `simple_switch`.
This will hide `boost::variant` to the target and remove the need for ugly `typeid` calls.

include/bm/bm_sim/field_lists.h
```diff
@@ -82,24 +102,31 @@ class FieldList {
...
+      }
       return seed;
     }
   };
 
  private:
-  std::vector<field_t> fields{};
-  std::unordered_set<field_t, FieldKeyHash> fields_set{};
+  std::vector<field_list_member_t> fields{};
+  std::unordered_set<field_list_member_t, FieldKeyHash> fields_set{};
```
antoninbas: You can just stick with fields in this set (`field_t`) and maybe rename the `contains` method to `contains_field`

include/bm/bm_sim/field_lists.h
```diff
@@ -46,23 +47,42 @@ class FieldList {
...
       return header == other.header && offset == other.offset;
     }
-
     bool operator!=(const field_t& other) const {
       return !(*this == other);
     }
   };
 
+  struct constant_t {
+    int value;
```
antoninbas: `ByteContainer`

include/bm/bm_sim/field_lists.h
```diff
@@ -82,24 +102,31 @@ class FieldList {
...
 
  private:
   struct FieldKeyHash {
-    std::size_t operator()(const field_t& f) const {
+    std::size_t operator()(const field_list_member_t& flm) const {
       std::size_t seed = 0;
-      boost::hash_combine(seed, f.header);
-      boost::hash_combine(seed, f.offset);
-
+      if (flm.type() == typeid(field_t)) {
```
antoninbas: implement with visitor

include/bm/bm_sim/field_lists.h
pierce-m: Okay. But why not just make the visit function take a const reference to FieldListVisitor?

antoninbas: AFAIK, boost visitors don't use virtual polymorphism. I am just reproducing this here. You could use virtual polymorphism if you wanted here and accept a `FieldListVisitor`, but you would need to make both operators in `FieldListVistor` virtual. You would also incur the cost of a virtual function call for each member of the variant vector (which is why boost visitors don't use it probably).

include/bm/bm_sim/field_lists.h
```diff
@@ -24,12 +24,17 @@
...
 
 #include <utility>  // for pair<>
 #include <vector>
 #include <unordered_set>
+#include <string>
 
 #include "phv_forward.h"
+#include "bytecontainer.h"
+
+using bm::ByteContainer;
```
antoninbas: Not needed as you only manipulate `ByteContainer` in the `bm` namespace to start with. Also `using` declarations are bad practice in headers like this.

include/bm/bm_sim/field_lists.h
```diff
@@ -46,59 +51,86 @@ class FieldList {
...
 
- public:
-  using iterator = std::vector<field_t>::iterator;
-  using const_iterator = std::vector<field_t>::const_iterator;
-  using reference = std::vector<field_t>::reference;
-  using const_reference = std::vector<field_t>::const_reference;
-  using size_type = size_t;
+  struct constant_t {
+    ByteContainer value;
+
```
antoninbas: I'm confused, where did the bitwidth go?

include/bm/bm_sim/field_lists.h
pierce-m: will remove

include/bm/bm_sim/field_lists.h
pierce-m: Bitwidth isn't needed if we're representing with ByteContainer. The only reason I needed it in the first place was to construct a ByteContainer for BufBuilder.

pierce-m: I guess it's possible that we'll have a constant as part of a field list that has some not-divisible-by-eight bitwidth and want to perform a calculation on the field list as a contiguous buffer. I'll add it back.

pierce-m: Actually nevermind, ByteContainer seems to keep sizes in bytes, not bits, so I'll keep it the way it is.

antoninbas: You need to put it back, for exactly this reason: `ByteContainer` keeps the size in bytes, not bits, and serializing to a buffer requires knowing the number of bits.

pierce-m: Okay I see the push_back_constant method in bufbuilder now. Will put it back.

PR #278, opened 2017-01-28
Add a Dockerfile for behavioral-model

sethfowler:
This PR adds a Dockerfile for this repo. It's based on the p4lang/third-party image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM p4lang/third-party:latest
+MAINTAINER Seth Fowler <seth.fowler@barefootnetworks.com>
+
+# Install dependencies.
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+      libboost-filesystem-dev \
+      libboost-thread-dev
+
+# Copy entire repository.
+COPY . /behavioral-model/
+WORKDIR /behavioral-model/
+
+# Build.
+RUN ./autogen.sh && \
+    ./configure && \
```
antoninbas: Please run configure with `--with-pdfixed` or it limits what can be done with this docker image. You should probably also add `-with-stress-tests` if we are later going to run the tests in the docker.

sethfowler: I'll file another PR to improve the Dockerfile. There are some other issues I noticed as well.

Issue #291, opened 2017-02-17
std::out_of_range


Starl1ght:
Following code is produced by p4c-bm compiler, everything is OK. However, once I run it in bmv2, I get 

> terminate called after throwing an instance of 'std::out_of_range'
>   what():  _Map_base::at

I've attached produced JSON. Sorry for very big size, however, this problem suddenly appeared out of nowhere, and was not able to exactly pin-point it and produce minimal example
[proj.json.txt](https://github.com/p4lang/behavioral-model/files/782622/proj.json.txt)

antoninbas:
This is because of the use of `<header_stack>[last]` in a set statement in the parser.
I was able to reproduce it with this simple P4 program:
```
header_type stack_t {
    fields { f1 : 8; f2 : 8; }
}

header stack_t stack[2];

header_type meta_t {
    fields { offset : 8; }
}

metadata meta_t meta;

parser start {
    extract(stack[next]);
    set_metadata(meta.offset, meta.offset + stack[last].f2);
    return ingress;
}

control ingress { }
```

antoninbas:
This is strongly related to this know issue: https://github.com/p4lang/behavioral-model/issues/152

Starl1ght:
Any possible workaround here? I do really need something like 

>  set_metadata(meta.offset, meta.offset + stack[last].f2);
for Ipv6 offsets parsing and other stuff :(

antoninbas:
Fixed in #292

Issue #152, opened 2016-06-20
Support header stacks references in expressions


antoninbas:
- s[last]
- s[next] ?

cw299:
Is the support of keyword **last** still under development? `simple_switch` throws an error message if my p4 program tries to access the largest-index valid instance of a header stack (inside of an action):

> terminate called after throwing an instance of 'std::out_of_range'
>   what():  _Map_base::at

My code looks like: 

```
action my_action() {
    modify_field(stack[last].field, 1);
}
```

antoninbas:
This is still under development. It is likely that I will wait until the new P4 (P4-16) spec comes out so that I can make sure the bmv2 implementation is consistent with the spec.

cw299:
I try to access an element of a stack via an index variable. The index variable should be represented by a metadata field. But a metadata field is not an integer data type.  Accordingly, the following example won't work:
```
modify_field(my_metadata.idx, 0);
modify_field(stack[my_metadata.idx].field, 1);
```
Is there any way to cast an metadata field to an integer?

antoninbas:
It would be useful to see what kind of error you are getting.
However, AFAIK:
- this is disallowed in the P4 language
- bmv2 does not support it
In other words, you can only access a header stack element through a compile-time constant.

cw299:
The compiler gives the following error:

> parse error in file /[...]/simple_switch.p4 at line 151 : Syntax error while parsing at token my_metadata (ID)
> parse error in file /[...]/simple_switch.p4 at line 151 : Invalid index in array header reference

_Background:_ 
I try to read a sequence of MPLS Labels that are attached on an incomming packet. The number of MPLS Labels is unknown until the Parser counts them and stores the number of labels into a metadata field. Each Label determines an output port. The task of the p4 program is to replicate the packet (by `clone*` operations) to these ports.

antoninbas:
There is probably a workaround that does not involve using a metadata field as an index into the header stack. For example, if you have a reasonable upper bound to the stack depth, you can always hand-write different actions like this:
```
table mpls_count {
    reads {
        my_metadata.count : exact;
    }
    actions { count_is_1; count_is_2; count_is_3; count_is_4; ...}
}
```
Obviously this is very verbose, but can be reasonable for up to say 8 mpls labels.
[...]

PR #292, opened 2017-02-18
Add better support for header stacks in expression

antoninbas:
We add support for hs[last].f, hs.size and hs.lastIndex which are valid
header stack operations in P4_16. This is implemented by introducing new
type of operations in the bmv2 expression engine, which can now appear
in the JSON input. These operations are dereference_stack,
last_stack_index, size_stack and access_field. Header stacks also become
valid operands. The JSON documentation was updated accordingly and the
JSON format version number was incremented to 2.5.


antoninbas:
See issues #152 and #291

PR #295, opened 2017-02-23
Enable the debugger in the Docker image

sethfowler:
It'd be nice to be able to use the debugger in the Docker image.

It would be even better to be able to install the debugger globally, so we wouldn't need to run it out of the source tree. It might be that automake can take care of that easily; I haven't investigated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ WORKDIR /behavioral-model/
 RUN apt-get update && \
     apt-get install -y --no-install-recommends $BM_DEPS $BM_RUNTIME_DEPS && \
     ./autogen.sh && \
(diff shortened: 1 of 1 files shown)


antoninbas:
@sethfowler what do you mean "install the debugger globally"? It is already installed as `bm_p4dbg` (by default under `/usr/local/bin`).

sethfowler:
Oh, nice. I was just going by the README, which suggests to use `tools/p4dbg.py`.

antoninbas:
Yes the README doesn't mention the installed tools (debugger, CLI,...) because it isn't how we've been using them the most so far. We should probably update it.

PR #303, opened 2017-03-03
Add support for table initializers in JSON

antoninbas:
This is done using the new 'entries' attribute in the JSON description
of a table. When this attribute is present the table becomes immutable:
JSON-provided entries cannot be deleted / modified and new entries
cannot be added.

The JSON documentation was updated to reflect this, and the version
number was bumped up to 2.7.


antoninbas:
FYI @cc10512 in case you want to take a look

docs/JSON_format.md
```diff
@@ -419,12 +425,44 @@ attributes for these objects are:
...
+    `prefix_length` attribute which has to be an integer
+    - for `ternary`: we need a `key` and a `mask` attributes which both have to
+    be hexstrings
+    - for `range`: we need a `start` and a `end` attribute which both have to be
+    hexstrings
+- `action_entry`: a JSON object with the following attributes:
+  - `action_id`: the id of the default action
+  - `action_data`: a JSON array where each entry is the hexstring value for an
+  action argument. The size of the array needs to match the number of parameters
+  expected by the action function with id `action_id`.
```
cc10512: I think that regardless of what we decide on the language side, the compiler should assign a priority to the entry.

docs/JSON_format.md
```diff
@@ -419,12 +425,44 @@ attributes for these objects are:
...
+
+Note that the `match_key` and the `action_entry` need to be formatted
+correctly. In particular:
+- the number of match fields in the match key needs to match the table
+description
+- the match types of the match fields in the match key need to match the table
+description
+- the hexstrings need to have the correct byte-width
+- duplicate entries (as determined by the macth keys) will trigger an error
```
cc10512: s/macth/match

I assume that the compiler should trigger an error for this. If we wait until runtime, it is not ideal.

src/bm_sim/P4Objects.cpp
```diff
@@ -1002,19 +1002,127 @@ P4Objects::init_actions(const Json::Value &cfg_root) {
...
+      case MatchKeyParam::Type::RANGE:
+        match_key.emplace_back(match_type,
+                               transform_hexstr(cfg_f["start"].asString()),
+                               transform_hexstr(cfg_f["end"].asString()));
+        break;
+    }
+  }
+  return match_key;
+}
+
```
cc10512: are you checking anywhere that the format of the json entry is valid or are you relying on cfg_f[key] to throw an exception?

docs/JSON_format.md
antoninbas: ok, will add a priority

src/bm_sim/P4Objects.cpp
antoninbas: not sure I understand this, I do check that the format of the match key is correct

PR #326, opened 2017-04-06
Update the Dockerfile to use third-party:stable and Ubuntu 16.04

sethfowler:
This PR updates the `Dockerfile` to:

- Build from `p4lang/third-party:stable`, instead of `:latest`.
- Build with Ubuntu 16.04, which we're using as the base image in `third-party` now.

It also adds a `.dockerignore` file to reduce the size of the generated image a little. I didn't try to get too aggressive with it, since upcoming Docker features will make this kind of thing much less of an issue.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+**/.git
+**/.gitmodules
+**/.travis.yml
+**/Dockerfile
+**/*.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM p4lang/third-party:latest
+FROM p4lang/third-party:stable
 MAINTAINER Seth Fowler <seth.fowler@barefootnetworks.com>
 
 # Select the type of image we're building. Use `build` for a normal build, which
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,5 @@
+**/.git
+**/.gitmodules
+**/.travis.yml
+**/Dockerfile
+**/*.md
```
antoninbas: you can probably exclude other stuff here:
- the travis folder
- I added a docker file called Dockerfile.grpc
- ...
but that's up to you

sethfowler: Yeah, there's probably a lot more stuff that could be excluded. Next month's release of Docker includes a feature called "multi-stage builds" (check it out if you haven't; it's cool!) which will let us perform a build and then throw away the entire source tree, leaving only the built artifacts. Since that's coming so soon and it'll totally solve this problem, I figured it's not worth the effort of optimizing this too much now.

On the other there, now that I think about it, even with multi-stage builds there *is* some value, since reducing the size of the context reduces the amount of data that Docker needs to copy into the image and could eliminate some spurious rebuilds. So perhaps a followup PR that added a more thorough .dockerignore would be worthwhile. Better to land this first, though, since any changes would require retesting of the downstream Docker images.

PR #333, opened 2017-04-17
Strip binaries in Dockerfile

sethfowler:
Stripping the binaries when we install them halves the size of the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ RUN apt-get update && \
     ./autogen.sh && \
     ./configure --enable-debugger --with-pdfixed --with-stress-tests && \
     make && \
-    make install && \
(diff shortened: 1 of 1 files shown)


cc10512:
isn't this going to limit our ability to debug?

antoninbas:
@cc10512 not debug P4 programs or the compiler, as this should be done using the bmv2 debugger and logs

sethfowler:
Yeah, the only use case it would interfere with would be if we want to debug BMV2 itself inside the Docker image. That may come up in the future, but we haven't had a need for it so far. (To my knowledge, at least.)

sethfowler:
I'm going to go ahead and merge this, but we can definitely revisit the issue if we start wanting to run gdb inside a container.

PR #335, opened 2017-04-18
Revert "Check for duplicate names / ids in JSON input"

antoninbas:
Reverts p4lang/behavioral-model#334

This change broke p4c regressions. Reverting this while we figure out if it is a problem with p4c which needs to be fixed, or if these checks are too strict.


jafingerhut:
I'll bet you $1 it is wrong output from p4c :-)

antoninbas:
I don't know if you're going to find someone willing to take that bet :)

PR #346, opened 2017-04-29
Extra logging when evaluating conditions, with source info

jafingerhut:
@antoninbas This is probably not in a form that you would like to merge, but I was hoping to get your thoughts on what approach you would like to see when using the new source_info added with the latest p4c commit earlier today.  The changes in this PR only use this information to add extra debug tracing when evaluating conditions, but of course this information could be used in many other places, too.

Here is a sample of a new log output line when evaluating a condition:

    [18:56:44.690] [bmv2] [T] [thread 8299] [0.0] [cxt 0] temp.p4(134): Condition "hdr.ethernet.dstAddr == 0xFFFFFFFFFFFF" is false

### DIFF
--- a/include/bm/bm_sim/named_p4object.h
+++ b/include/bm/bm_sim/named_p4object.h
@@ -25,6 +25,7 @@
 
 #include <bm/bm_sim/source_info.h>
 #include <string>
+#include <memory>
 
 namespace bm {
 
@@ -37,10 +38,10 @@ using p4object_id_t = int;
 class NamedP4Object {
  public:
   NamedP4Object(const std::string &name, p4object_id_t id)
-    : name(name), id(id), source_info(nullptr) {}
+    : name(name), id(id) {}
   NamedP4Object(const std::string &name, p4object_id_t id,
-                const SourceInfo *source_info)
-    : name(name), id(id), source_info(source_info) {}
+                std::unique_ptr<SourceInfo> source_info)
+    : name(name), id(id), source_info(std::move(source_info)) {}
 
   virtual ~NamedP4Object() { }
 
@@ -65,12 +66,12 @@ class NamedP4Object {
   unsigned get_column() const { return column; }
   const std::string &get_source_fragment() const { return source_fragment; }
   bool has_source_info() const { return (source_info != nullptr); }
-  const SourceInfo *get_source_info() const { return source_info; }
+  const SourceInfo *get_source_info() const { return source_info.get(); }
 
  protected:
   const std::string name;
   p4object_id_t id;
-  const SourceInfo *source_info;
+  std::unique_ptr<SourceInfo> source_info;
 };
 
 }  // namespace bm
--- a/include/bm/bm_sim/named_p4object.h
+++ b/include/bm/bm_sim/named_p4object.h
@@ -37,8 +37,12 @@ class NamedP4Object {
  public:
   NamedP4Object(const std::string &name, p4object_id_t id)
       : name(name), id(id), has_source_info(false) {}
-  NamedP4Object(const std::string &name, p4object_id_t id, const std::string &filename, unsigned line, unsigned column, const std::string &source_fragment)
-      : name(name), id(id), has_source_info(true), filename(filename), line(line), column(column), source_fragment(source_fragment) {}
+  NamedP4Object(const std::string &name, p4object_id_t id,
+                const std::string &filename, unsigned line,
+                unsigned column, const std::string &source_fragment)
+      : name(name), id(id), has_source_info(true),
+        filename(filename), line(line),
+        column(column), source_fragment(source_fragment) {}
 
   virtual ~NamedP4Object() { }
 
--- a/include/bm/bm_sim/named_p4object.h
+++ b/include/bm/bm_sim/named_p4object.h
@@ -43,11 +43,6 @@ class NamedP4Object {
                 std::unique_ptr<SourceInfo> source_info)
     : name(name), id(id), source_info(std::move(source_info)) {}
 
-  NamedP4Object(const NamedP4Object& rhs)
-      : name(rhs.name), id(rhs.id),
-        source_info(rhs.source_info ? new SourceInfo(*rhs.source_info)
-                    : nullptr) { }
-
   virtual ~NamedP4Object() { }
 
   //! Get the name of the P4 instance
--- a/include/bm/bm_sim/named_p4object.h
+++ b/include/bm/bm_sim/named_p4object.h
@@ -24,6 +24,8 @@
 #define BM_BM_SIM_NAMED_P4OBJECT_H_
 
 #include <string>
+#include <sstream>
+#include <bm/bm_sim/source_info.h>
 
 namespace bm {
 
@@ -36,13 +38,10 @@ using p4object_id_t = int;
 class NamedP4Object {
  public:
   NamedP4Object(const std::string &name, p4object_id_t id)
-      : name(name), id(id), has_source_info(false) {}
+    : name(name), id(id), source_info(nullptr) {}
   NamedP4Object(const std::string &name, p4object_id_t id,
-                const std::string &filename, unsigned line,
-                unsigned column, const std::string &source_fragment)
-      : name(name), id(id), has_source_info(true),
-        filename(filename), line(line),
-        column(column), source_fragment(source_fragment) {}
+                const SourceInfo *source_info)
+    : name(name), id(id), source_info(source_info) {}
 
   virtual ~NamedP4Object() { }
 
@@ -66,15 +65,13 @@ class NamedP4Object {
   unsigned get_line() const { return line; }
   unsigned get_column() const { return column; }
   const std::string &get_source_fragment() const { return source_fragment; }
+  bool has_source_info() const { return (source_info != nullptr); }
+  const SourceInfo *get_source_info() const { return source_info; }
 
  protected:
   const std::string name;
   p4object_id_t id;
-  bool has_source_info;
-  const std::string filename;
-  unsigned line;
-  unsigned column;
-  const std::string source_fragment;
+  const SourceInfo *source_info;
 };
 
 }  // namespace bm
--- a/include/bm/bm_sim/named_p4object.h
+++ b/include/bm/bm_sim/named_p4object.h
@@ -61,11 +61,6 @@ class NamedP4Object {
   //! Default assignment operator
   NamedP4Object &operator=(NamedP4Object &&other) = default;
 
-  const std::string &get_filename() const { return filename; }
-  unsigned get_line() const { return line; }
-  unsigned get_column() const { return column; }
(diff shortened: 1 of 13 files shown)


jafingerhut:
"Clutter the logs" and "include vital information that speeds up debugging rates by 5x" are two sides of the same coin, IMO, and I lean towards the latter myself.  Interactive debuggers are not everyone's cup of tea -- they *must* be interactive.  Logs can be generated the first time a long test is run, and potentially provide all the details needed to figure out what went wrong quickly.

I have pushed another commit that makes a new SourceInfo class to hold the source info, and uses it, but still only for conditionals.

Where in the bmv2 code would you recommend looking to add tracing of source info when primitive actions within compound actions are executed?  I have looked at that a bit before, but it wasn't obvious to me exactly how the primitive actions are stored.

include/bm/bm_sim/source_info.h
```diff
@@ -0,0 +1,85 @@
...
+
+#ifndef BM_BM_SIM_SOURCE_INFO_H_
+#define BM_BM_SIM_SOURCE_INFO_H_
+
+//! @file source_info.h
+//! Stores source location information about some objects, as read from
+//! the JSON file produced by the compiler.
+
+#include <string>
+#include "jsoncpp/json.h"
```
antoninbas: It would be easy to avoid introducing this dependency in the header file. You could forward declare `Json::Value` and move the implementation of `newFromJson` to an implementation file. Or better still, I think the json parsing code should stay in `P4Objects.cpp`; it would be easier to do sanity checking on the json if you left it there (because of the code already defined there), and would keep the json parsing in a single place.

include/bm/bm_sim/source_info.h
```diff
@@ -0,0 +1,85 @@
...
+    if (!cfg_source_info["line"].isNull()) {
+      line = cfg_source_info["line"].asInt();
+    }
+    if (!cfg_source_info["column"].isNull()) {
+      column = cfg_source_info["column"].asInt();
+    }
+    if (!cfg_source_info["source_fragment"].isNull()) {
+      source_fragment = cfg_source_info["source_fragment"].asString();
+    }
+    return new SourceInfo(filename, line, column, source_fragment);
```
antoninbas: As mentioned above, you should use a `unique_ptr` to express memory ownership. It is also unclear whether you are releasing this memory anywhere right now (I'm leaning towards no), which you wouldn't have to worry about with a managed pointer.

include/bm/bm_sim/source_info.h
```diff
@@ -0,0 +1,85 @@
...
+  SourceInfo()
+    : filename(""), line(0), column(0), source_fragment("") { }
+  SourceInfo(std::string filename, unsigned line, unsigned column,
+             std::string source_fragment)
+    : filename(filename),
+      line(line),
+      column(column),
+      source_fragment(source_fragment) { }
+
+  static SourceInfo *newFromJson(const Json::Value &cfg_source_info) {
```
antoninbas: method name not consistent with the rest of the code (we use underscores)

src/bm_sim/P4Objects.cpp
```diff
@@ -1513,8 +1513,11 @@ P4Objects::init_pipelines(const Json::Value &cfg_root,
     for (const auto &cfg_conditional : cfg_conditionals) {
       const string conditional_name = cfg_conditional["name"].asString();
       p4object_id_t conditional_id = cfg_conditional["id"].asInt();
-      Conditional *conditional = new Conditional(conditional_name,
-                                                 conditional_id);
+      const Json::Value &cfg_source_info = cfg_conditional["source_info"];
+      Conditional *conditional =
```
antoninbas:
```
auto conditional = new Conditional(
  conditional_name, conditional_id, ...)
```
would be a better formatting IMO

include/bm/bm_sim/source_info.h
```diff
@@ -0,0 +1,85 @@
...
+    std::stringstream result;
+    result << filename << "(" << line << ":" << column << ")";
+    return result.str();
+  }
+
+  std::string get_source_fragment() const { return source_fragment; }
+
+ private:
+  std::string filename;
+  unsigned line;
```
antoninbas: use `unsigned int` for consistency with the rest of the code

include/bm/bm_sim/named_p4object.h
```diff
@@ -46,9 +51,13 @@ class NamedP4Object {
   //! Get the compiler-provided id
   p4object_id_t get_id() const { return id; }
 
+  bool has_source_info() const { return (source_info != nullptr); }
+  const SourceInfo *get_source_info() const { return source_info; }
+
  protected:
   const std::string name;
   p4object_id_t id;
+  const SourceInfo *source_info;
```
jafingerhut: I have attempted to change the SourceInfo member of class NamedP4Object to use unique_ptr, but I freely admit that C++'s rules for these things have stymied me.

I have checked in some code that attempts to use unique_ptr for this, on a branch that is not part of this PR, here: https://github.com/jafingerhut/behavioral-model/commit/3c7f2411d480cf984ce0fd7b4b4c48c51f222cac

I have added a comment to my own diff there with some questions, in case anyone is interested in taking a look -- if you know C++ default copy constructors and how they interact with unique_ptr fields, you may know what I am missing here.

include/bm/bm_sim/named_p4object.h
```diff
@@ -24,6 +24,8 @@
 #define BM_BM_SIM_NAMED_P4OBJECT_H_
 
 #include <string>
+#include <sstream>
```
jafingerhut: It used to be there because of a use of stringstream in that file.  I moved it to another file where stringstream is currently being used.

include/bm/bm_sim/named_p4object.h
jafingerhut: UPDATE: I figure out one way to make unique_ptr's work, which is now part of this PR.  You may still have comments about how I have used unique_ptr's, if I have done so incorrectly or in an odd way.

include/bm/bm_sim/source_info.h
jafingerhut: I have moved newFromJson to P4Object.cpp as you suggested.

include/bm/bm_sim/source_info.h
jafingerhut: I changed it to new_from_json, but now that it is outside of class SourceInfo that name isn't so specific any more, and could probably use a better one.

include/bm/bm_sim/source_info.h
jafingerhut: Done.

include/bm/bm_sim/source_info.h
jafingerhut: Done

src/bm_sim/P4Objects.cpp
jafingerhut: Done

antoninbas:
@jafingerhut Your troubles with `unique_ptr` are party my bad. The copy constructor and copy assignment operator in the `NamedP4Object` class should have been explicitly deleted a long time ago. Because they were not explicitly deleted, I relied on them for `HeaderUnionType`. I will explicitly delete them and make the necessary changes. There are very few case where you should write an explicit copy constructor.

antoninbas:
I opened a PR to explicitly delete the copy constructor and copy assignment operator for NamedP4Object: https://github.com/p4lang/behavioral-model/pull/349. I suggest you rebase it into your branch once I have merged it into master. After that I can do another review pass on your PR.

jafingerhut:
Thanks.  I will wait for the commit of your other change and rebase my changes based on those, removing my explicit copy constructor.

include/bm/bm_sim/source_info.h
```diff
@@ -0,0 +1,61 @@
...
+  SourceInfo()
+    : filename(""), line(0), column(0), source_fragment("") { }
+  SourceInfo(std::string filename, unsigned int line, unsigned int column,
+             std::string source_fragment)
+    : filename(filename),
+      line(line),
+      column(column),
+      source_fragment(source_fragment) { }
+
+  std::string toString() const {
```
antoninbas: my recommendation is to pre-compute that string in the constructor and store it as a class member as this is not such a cheap operation
also could this happen in an implementation file, to avoid the dependency on `sstream` in the header file

src/bm_sim/actions.cpp
```diff
@@ -289,6 +290,8 @@ ActionFnEntry::execute(Packet *pkt) const {
   auto &primitives = action_fn->primitives;
   size_t param_offset = 0;
   // primitives is a vector of pointers
+  const SourceInfo *si = action_fn->get_source_info();
```
antoninbas: put it in the macro below so there is no penalty when logging macros are disabled

src/bm_sim/conditionals.cpp
```diff
@@ -36,6 +37,16 @@ Conditional::operator()(Packet *pkt) const {
   PHV *phv = pkt->get_phv();
   bool result = eval(*phv);
   BMELOG(condition_eval, *pkt, *this, result);
+
+  // It would be nicer to see the following additional info in the log:
+  //
+  // + The full expression even if it spans multiple lines.
+  // + The current values of all variables involved in evaluating the
+  //   expression.
+  const SourceInfo *si = get_source_info();
```
antoninbas: same comment as above, call `get_source_info` in the macro

include/bm/bm_sim/source_info.h
jafingerhut: Done

src/bm_sim/actions.cpp
jafingerhut: Done

src/bm_sim/conditionals.cpp
jafingerhut: Done

antoninbas:
I actually noticed 2 additional things:
- you forgot to add `source_info.h` to the Makefile, so it will not get installed properly
- you should rename `toString` to `to_string` for consistency

jafingerhut:
Fixed toString -> to_string and added source_info.h to Makefile.

I have added some code to the test_p4objects.cpp function you pointed out, but I don't know if those changes by themselves will increase code coverage until I see the test results (or learn to run code coverage results on my personal Ubuntu VM, but I'd rather let Travis give me the results if that is OK).

tests/test_p4objects.cpp
```diff
@@ -517,6 +517,12 @@ class JsonBuilder {
     Json::Value cond(Json::objectValue);
     cond["name"] = name;
     cond["id"] = 0;
+    Json::Value source_info(Json::objectValue);
```
antoninbas: this is not really what I had in mind, but we can add it as is and I will add a proper test after it is merged;,I will assign you for the review

jafingerhut: Sounds good.  Thanks for all the help in getting these changes into committable shape.

PR #352, opened 2017-05-04
Add better test for SouceInfo

antoninbas:


antoninbas:
@jafingerhut you may want to review this

PR #355, opened 2017-05-07
Trace primitive action execution, with source info

jafingerhut:


include/bm/bm_sim/actions.h
```diff
@@ -636,6 +636,7 @@ class ActionFn :  public NamedP4Object {
   void parameter_push_back_string(const std::string &str);
 
   void push_back_primitive(ActionPrimitive_ *primitive);
+  void push_back_source_info(std::unique_ptr<SourceInfo> source_info);
```
antoninbas: I would make it an optional parameter to `push_back_primitive` instead, it makes more sense IMO

src/bm_sim/P4Objects.cpp
```diff
@@ -381,8 +381,10 @@ P4Objects::add_primitive_to_action(const Json::Value &cfg_primitive,
         EFormat() << "Unknown primitive action: " << primitive_name);
 
   action_fn->push_back_primitive(primitive);
+  action_fn->push_back_source_info(object_source_info(cfg_primitive));
 
   const auto &cfg_primitive_parameters = cfg_primitive["parameters"];
+  std::vector<ActionPrimitive_ *> primitives{};
```
antoninbas: is this actually needed?

tests/test_actions.cpp
```diff
@@ -254,6 +254,7 @@ TEST_F(ActionsTest, SetFromConst) {
   Data value(0xaba);
   SetField primitive;
   testActionFn.push_back_primitive(&primitive);
+  testActionFn.push_back_source_info(nullptr);
```
antoninbas: if you make the source info an optional parameter to `push_back_primitive` as suggested above, you will not need to modify all these test files

src/bm_sim/actions.cpp
```diff
@@ -289,12 +291,18 @@ ActionFnEntry::execute(Packet *pkt) const {
 
   auto &primitives = action_fn->primitives;
   size_t param_offset = 0;
+  size_t si_offset = 0;
   // primitives is a vector of pointers
   BMLOG_TRACE_SI_PKT(*pkt, action_fn->get_source_info(),
-                     "Executing action {}", action_fn->get_name());
+                     "Action {}", action_fn->get_name());
   for (auto primitive : primitives) {
+    BMLOG_TRACE_SI_PKT(*pkt, action_fn->source_infos[si_offset],
```
antoninbas: That is really going to lead to a huge log file for people who have to debug switch.p4 applications...

jafingerhut: They can be disabled with '-L debug' instead of the current default '-L trace'.

Would you be interested in changing the default to '-L debug' so only people who explicitly set '-L debug' will see these new messages?

Or creating a new level higher than '-L trace' (although that might require changes to spdlog, which might not be desired).

jafingerhut: Oh, and I have to debug even larger P4 applications, and they are a pain to debug *without* this kind of detail, IMHO.

jafingerhut:
I guess I need to try running with g++-6.0 if I want to reproduce the failing test case locally on my machine?

src/bm_sim/actions.cpp
jafingerhut: @antoninbas Any thoughts on my questions about possible changes to the default log level, or creating a new log level more detailed than the current '-L trace' ?

jafingerhut:
Closing in favor of a later PR that merges cleanly with latest master: https://github.com/p4lang/behavioral-model/pull/381

PR #356, opened 2017-05-07
Read JSON source_info line and column values as 32-bit unsigned ints

jafingerhut:
They were being interpreted as 32-bit signed ints, which causes an out
of range error if the line number is in range 2^31 through 2^32-1.
Due to a bug in p4c-bm2-ss right now, 2^32-1 is possible, which is how
I discovered this issue.


jafingerhut:
I plan to fix the incorrect 2^32-1 line numbers being generated in some cases by p4c-bm2-ss, too, but this seems like the right thing to do on the bmv2 side, regardless of that bug, since p4c-bm2-ss is generating this value as an unsigned 32-bit int.

PR #357, opened 2017-05-08
Replace some asserts with error messages

antoninbas:
in JSON parsing code


vgurevich:
It would be really nice if these error messages would also state the name of a particular array/attribute/etc for the ease of JSON debugging. Line number would be ideal, but that it probably more difficult.

PR #358, opened 2017-05-08
Remove the OverwritingHeader parser error

antoninbas:
As per an update to the P4_16 spec


antoninbas:
This cannot be merged until the corresponding changes have been done in p4lang/p4c. See https://github.com/p4lang/p4c/issues/578

PR #379, opened 2017-05-23
Add support for conditional jump in action bodies

antoninbas:
We tentatively add support for 2 "core" action primitives for absolute
jumps: `_jump` and `_jump_if_zero`. `_jump` takes one parameter which
must resolve to an integral value `offset`. When it is executed, we jump
to the primitive call at index `offset` in the enclosing
action. `_jump_if_zero` is a conditional jump and takes two parameters
which must both resolve to integral values, which we call respectively
`cond` and `offset`. If `cond` is 0, the primitive behaves exactly like
`_jump` and we jump to the primitive call at index `offset`; otherwise
the primitive is a no-op.

This was implemented at no additional cost compared to the existing
implementation. In other words, if an input does not use these 2
primitives, performance will not be affected.

This can be used by a P4_16 compiler to support conditions in actions.

version number was bumped up to 2.13

### DIFF
--- a/src/bm_sim/actions.cpp
+++ b/src/bm_sim/actions.cpp
@@ -25,7 +25,6 @@
 #include <bm/bm_sim/packet.h>
 #include <bm/bm_sim/logger.h>
 
-#include <iostream>
 #include <string>
 #include <vector>
 
@@ -219,8 +218,14 @@ ActionFn::parameter_push_back_string(const std::string &str) {
 }
 
 void
-ActionFn::push_back_primitive(ActionPrimitive_ *primitive) {
-  primitives.push_back(primitive);
+ActionFn::push_back_primitive(ActionPrimitive_ *primitive,
+                              std::unique_ptr<SourceInfo> source_info) {
+  size_t param_offset = 0;
+  if (!primitives.empty()) {
+    param_offset = primitives.back().get_param_offset();
+    param_offset += primitives.back().get_num_params();
+  }
+  primitives.emplace_back(primitive, param_offset, std::move(source_info));
 }
 
 void
@@ -268,6 +273,24 @@ ActionOpcodesMap::get_instance() {
   return &instance;
 }
 
+ActionPrimitiveCall::ActionPrimitiveCall(
+    ActionPrimitive_ *primitive, size_t param_offset,
+    std::unique_ptr<SourceInfo> source_info)
+    : primitive(primitive), param_offset(param_offset),
+      source_info(std::move(source_info)) { }
+
+void
+ActionPrimitiveCall::execute(ActionEngineState *state,
+                             const ActionParam *args) const {
+  // TODO(unknown): log source info?
+  primitive->execute(state, args);
+}
+
+size_t
+ActionPrimitiveCall::get_num_params() const {
+  return primitive->get_num_params();
+}
+
 void
 ActionFnEntry::push_back_action_data(const Data &data) {
   action_data.push_back_action_data(data);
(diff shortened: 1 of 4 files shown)


antoninbas:
@jafingerhut this means that your PR #355 will need to be updated but I think the code will look nicer. I created the `ActionPrimitiveCall` class, which can include source information.
Once the p4c bmv2 backend is updated to leverage the jump primitives, I think you can attach source information to these primitive calls which will be valuable for debugging.

jafingerhut:
@antoninbas If we get a little generalization in bmv2 for the P4-16 programs that can be compiled, I am more than happy to update that PR to adjust.  Thanks.  I did have a question on that PR regarding log levels that may address your concern about log file growth there.

mbudiu-vmw:
Wouldn't it be simple to introduce a `label` primitive whose semantics is a no-op, and to have jumps use labels? It will certainly make code generation and debugging easier; I expect understanding the execution of a nested if conditional will be hairy with the jumps as proposed.

antoninbas:
The label primitive would be a pain to support; it would require some special handling wrt other primitives for backward-jumps (which I agree are not needed for if/else support), which I don't think is a good idea. Supporting string labels would also not be ideal.

One thing I could do is try to replace relative jumps with absolute jumps to a primitive operation. In a way this is the same as supporting labels, instead the label would be the index of the primitive operation in the sequence and there would be no need for a special label primitive as each primitive operation would be implicitly labelled with its index. @mbudiu-vmw would you like this better?

mbudiu-vmw:
That would work; so the label is just a number denoting the position within the action.
We could even emit that as part of json to make it easy to read, but it would not be interpreted by the runtime.

antoninbas:
To make sure we are on the same page, I'll give some more details.
Let's look at the following code:
```
if (hdr1.f8)
    hdr1.f32 = 0xaa;
else
    hdr1.f32 = 0xbb;
```

Right now this pull request would require some json that looks like this for the primitive list in the action:
```
[op0(_jump_if_zero, [hdr1.f8, 2]), op1(assign, [hdr1.f32, 0xaa]), op2(_jump, [1]), op3(assign, [hdr1.f32, 0xbb])]
```
Notice how the jumps use relative offsets. When `op0` is executed, if `hdr1.f8` is 0, we skip the next `2` operations and execute `op3`.

What I'm proposing is to make the jumps absolute. The new json would look like this:
```
[op0(_jump_if_zero, [hdr1.f8, 3]), op1(assign, [hdr1.f32, 0xaa]), op2(_jump, [4]), op3(assign, [hdr1.f32, 0xbb])]
```
The offsets are now absolute. When `op0` is executed, if `hdr1.f8` is 0, we jump to the operation at index `3`, which is `op3`.

The change I'm offering is a very small one. But using absolute offsets make things a little more readable and maybe a little easier to generate in p4c.

As you correctly mention, the offsets can be explicitly added in the JSON for the sake for readability, the bmv2 json parser will ignore them. For example:
```
{
    "op": "assign",
    "parameters": [...],
    offset: 0
},
{
    "op": "assign",
    "parameters": [...],
    offset: 1
}
```

@mbudiu-vmw let me know if we are on the same page and I'll make the change this afternoon.

mbudiu-vmw:
Yes, that looks good. Thank you.

antoninbas:
I updated the PR to support absolute jumps.

src/bm_sim/actions.cpp
```diff
@@ -289,12 +312,13 @@ ActionFnEntry::execute(Packet *pkt) const {
...
   BMLOG_TRACE_SI_PKT(*pkt, action_fn->get_source_info(),
                      "Executing action {}", action_fn->get_name());
-  for (auto primitive : primitives) {
-    primitive->execute(&state, &(action_fn->params[param_offset]));
-    param_offset += primitive->get_num_params();
+  for (size_t idx = 0; idx < primitives.size();) {
+    const auto &primitive = primitives[idx];
+    param_offset = primitive.get_param_offset();
+    primitive.execute(&state, &(action_fn->params[param_offset]));
+    idx = primitive.get_jump_offset(idx);
```
mbudiu-vmw: Would you like to check here that idx is greater than the previous value?

antoninbas: it does not have to be, we could imagine supporting backward jumps (for loops)
what would make sense is to check the jump indices in the bmv2 json parser but that would require some special handling for the jump primitives, which I am trying to avoid

hesingh:
@antoninbas I like what you did for code for this issue.

However, I wonder, if one could use keywords in JSON such as if_cond, else_cond, else_if_cond to support if-statements?  This way, the format is also consistent with the if_cond used in checksum.  Further, whether the if-statement exists in the apply block of a control or inside an action, we could generate the same JSON, couldn't we?

PR #385, opened 2017-06-01
More advanced support for CRCs

antoninbas:
We now support a long list of pre-defined crc algorithms, as well as
some custom string format to define new algorithms (string format lets
the user defined polynomial, initial remainder, etc).

### DIFF
--- a/src/bm_sim/P4Objects.cpp
+++ b/src/bm_sim/P4Objects.cpp
@@ -30,6 +30,7 @@
 #include <exception>
 
 #include "jsoncpp/json.h"
+#include "crc_map.h"
 
 namespace bm {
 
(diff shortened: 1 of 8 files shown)


src/bm_sim/P4Objects.cpp
```diff
@@ -2211,9 +2212,20 @@ P4Objects::check_required_fields(
...
+  {
+    auto h = CalculationsMap::get_instance()->get_copy(name);
+    if (h) return h;
+  }
+  {
+    auto h = CrcMap::get_instance()->get_copy(name);
+    if (h) return h;
+  }
+  {
+    auto h = CrcMap::get_instance()->get_copy_from_custom_str(name);
```
cc10512: Since you're not doing any additional checking before calling the name, is there a particular reason that you expose both get_copy and get_copy_form_custom_str? That is vs. get_copy checking the map and if there is a miss, directly calling get_copy_from_custom_str as a private method?

antoninbas: No particular reason, but I don't see a compelling reason to have a single public method as well, especially given that there is a single call site.

PR #395, opened 2017-06-20
Ability to query object id from name with thrift

antoninbas:
As an alternative to querying the entire JSON config and parsing it in
the client.

### DIFF
--- a/src/bm_runtime/Standard_server.cpp
+++ b/src/bm_runtime/Standard_server.cpp
@@ -1124,6 +1124,51 @@ public:
     _return.append(switch_->get_config_md5());
   }
 
+  int32_t bm_get_id_from_name(const int32_t cxt_id, const BmResourceType::type resource_type, const std::string& resource_name) {
+    Logger::get()->trace("bm_get_id_from_name");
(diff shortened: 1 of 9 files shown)


src/bm_runtime/Standard_server.cpp
```diff
@@ -1124,6 +1124,48 @@ class StandardHandler : virtual public StandardIf {
...
+    p4object_id_t resource_id;
+    auto rc = switch_->p4objects_id_from_name(
+        cxt_id, map_type(resource_type), resource_name, &resource_id);
+    if (rc != P4Objects::IdLookupErrorCode::SUCCESS) {
+      InvalidIdLookup error;
+      switch (rc) {
+        case P4Objects::IdLookupErrorCode::INVALID_RESOURCE_NAME:
+          error.code = IdLookupErrorCode::INVALID_RESOURCE_NAME;
+          break;
+        default:
```
sethfowler: Nit: you could probably just use an `if` instead of a `switch` here.

antoninbas: this was done on purpose, to support additional error codes in the future; I have actually added a new one already.

PR #436, opened 2017-09-08
Necessary changes in simple_switch_grpc tests

antoninbas:
To support P4Runtime controller arbitration.


antoninbas:
See https://github.com/p4lang/PI/pull/199

PR #452, opened 2017-10-28
handle implicit fallthrough in switch statement

toanju:
this mitigates -Werror=implicit-fallthrough for gcc7


toanju:
Great, I'll update this then.

PR #458, opened 2017-11-02
Use new {Set|Get}ForwardingPipelineConfig RPC fields

antoninbas:
p4runtime.proto was updated and the format of the
{Set|Get}ForwardingPipelineConfig RPCs was modified, with some fields
marked as deprecated and some new fields introduced. This commit makes
sure that only the new fields are used.


antoninbas:
Merged at the same time as https://github.com/p4lang/PI/pull/224

PR #461, opened 2017-11-04
Enhance veth_setup.sh to disable Linux kernel sending IPv6 pkts, enab…

jafingerhut:
…le jumbo frames

I have been using this modified script on Ubuntu 16.04 Linux VMs for a while now, and it has some useful enhancements over the existing one.

### DIFF
--- a/tools/veth_setup.sh
+++ b/tools/veth_setup.sh
@@ -6,5 +6,24 @@ for idx in 0 1 2 3 4 5 6 7 8; do
         ip link add name $intf0 type veth peer name $intf1
         ip link set dev $intf0 up
         ip link set dev $intf1 up
+
+        # Set the MTU of these interfaces to be larger than default of
+        # 1500 bytes, so that P4 behavioral-model testing can be done
+        # on jumbo frames.
+        ifconfig $intf0 mtu 9500 up
+        ifconfig $intf1 mtu 9500 up
+
+        # Disable IPv6 on the interfaces, so that the Linux kernel
+        # will not automatically send IPv6 MDNS, Router Solicitation,
+        # and Multicast Listener Report packets on the interface,
+        # which can make P4 program debugging more confusing.
+        #
+        # Testing indicates that we can still send IPv6 packets across
+        # such interfaces, both from scapy to simple_switch, and from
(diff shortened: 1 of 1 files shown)


tools/veth_setup.sh
```diff
@@ -6,5 +6,24 @@ for idx in 0 1 2 3 4 5 6 7 8; do
...
+        # and Multicast Listener Report packets on the interface,
+        # which can make P4 program debugging more confusing.
+        #
+        # Testing indicates that we can still send IPv6 packets across
+        # such interfaces, both from scapy to simple_switch, and from
+        # simple_switch out to scapy sniffing.
+        #
+        # https://superuser.com/questions/356286/how-can-i-switch-off-ipv6-nd-ra-transmissions-in-linux
+        sysctl net.ipv6.conf.${intf0}.disable_ipv6=1
+        sysctl net.ipv6.conf.${intf1}.disable_ipv6=1
```
vgurevich: Actually, if we are on that topic, another very important thing is to disable all HW acceleration. Probably not that a big deal if you use VETHs, but a very important thing on most modern ethernet cards. 

```
    TOE_OPTIONS="rx tx sg tso ufo gso gro lro rxvlan txvlan rxhash"
    for TOE_OPTION in $TOE_OPTIONS; do
       /sbin/ethtool --offload $intf0 "$TOE_OPTION" off &> /dev/null
       /sbin/ethtool --offload $intf1 "$TOE_OPTION" off &> /dev/null
    done
```

jafingerhut: So this script is called veth_setup.sh and is dedicated for the purpose of creating veth interfaces.

If those commands are harmless for veth's, I don't see a problem adding them to this script, but they aren't helpful, either, are they?

If there is a separate script somewhere for configuring physical ethernet interfaces, perhaps a separate script in this same tools directory, that would seem to make more sense than adding such commands to veth_setup.sh

vgurevich: @jafingerhut ,

I haven't tried in a while to run this script without these lines (i.e. they have been present in the script I use for the past two years or so). They are definitely not harmful, but you are right, perhaps it makes sense to retest and see if they are helpful for veths.

PR #464, opened 2017-11-16
Do not crash in CLI if match field format is bad

antoninbas:
For example, if a prefix is not in the format <prefix>/<length>.


tools/runtime_CLI.py
```diff
@@ -523,7 +533,12 @@ def parse_param_(field, bw):
             param = BmMatchParam(type = param_type,
                                  valid = BmMatchParamValid(key))
         elif param_type == BmMatchParamType.RANGE:
-            start, end = field.split("->")
+            try:
+                start, end = field.split("->")
```
vgurevich: Can we use ..  or ... for the range? It is way more natural, in my view. We can add this in addition to ->, but I think that if we replace it, few people will notice. It is also easy to confuse -> with =>

PR #473, opened 2017-11-28
Check for duplicate names / ids in JSON input

antoninbas:
Duplicate names or ids are forbidden for objects of the same type
(e.g. header type, parser...). We enforce this restriction in the JSON
parser.

### DIFF
--- a/src/bm_sim/P4Objects.cpp
+++ b/src/bm_sim/P4Objects.cpp
@@ -28,6 +28,7 @@
 #include <tuple>
 #include <vector>
 #include <set>
+#include <unordered_set>
 #include <exception>
 
 #include "jsoncpp/json.h"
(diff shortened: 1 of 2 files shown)


antoninbas:
@jafingerhut please take a look

jafingerhut:
@antoninbas Looks good to me.  I compiled with these changes locally, and tried running the resulting simple_switch binary on all bmv2 JSON files generated by the 2017-Nov-27 version of p4c-bm2-ss on programs in the p4c/testdata/p4_14_samples and p4c/testdata/p4_16_samples directories, and the only error messages caused by these new checks in this PR were for the programs listed below.

Except for the newer one, issue986-1-bmv2.json, this is exactly the list I expected to see.  The error message for that newer program makes sense, too, given that it is Mihai's attempt to allow a bmv2 JSON file to invoke the same table last in the then and else branch of an if statement, and the table does show up as the same-named control node as a result.

```
----------------------------------------------------------------------
Try it on the bmv2 JSON files produced by compiling all programs that
can be successfully compiled using 2017-Nov-27 version of p4c-bm2-ss
in the directory:

    p4c/testdata/p4_16_samples

Below are the ones that simple_switch gives a duplicate name or id
error when loading the JSON file using this command:

    simple_switch --use-files 0 -i 0@0 <bmv_json_file_name>

----------------------------------------
issue986-1-bmv2.json
Thrift port was not specified, will use 9090
Duplicate objects of type 'control node' with name 't1'

This one was added during Nov 2017.  It does have two control nodes
with table t1, intentionally, because Mihai Budiu was attempting to
allow as many P4_16 programs that the bmv2 JSON file format can
accept, without having different next_table values for the same table
in the source code.

The easiest thing is probably to disallow this source program to be
run on bmv2, as this behavioral-model PR would do.

----------------------------------------
x-bmv2.json
Thrift port was not specified, will use 9090
Duplicate objects of type 'header' with name 's'

This one happened before, too, and from looking at the source code and
bmv2 JSON file, it should cause a duplicate name error (until the p4c
bug is fixed).

----------------------------------------------------------------------
Try it on the bmv2 JSON files produced by compiling all programs that
can be successfully compiled using 2017-Nov-27 version of p4c-bm2-ss
in the directory:

    p4c/testdata/p4_14_samples

Below are the ones that simple_switch gives a duplicate name or id
error when loading the JSON file using this command:

    simple_switch --use-files 0 -i 0@0 <bmv_json_file_name>

----------------------------------------
16-TwoReferences.json
Thrift port was not specified, will use 9090
Duplicate objects of type 'control node' with name 'C'

This one happened before, too, and from looking at the source code and
bmv2 JSON file, it should cause a duplicate name error (until the p4c
bug is fixed).
```

antoninbas:
@jafingerhut even if the table is applied at the end of each if / else branch, my understanding is that it is still the same table in bmv2, so it should only appear once in the JSON.

jafingerhut:
The file issue986-1-bmv2.json generated via the 2017-Nov-27 version of p4c-bm2-ss does have a table "t1" represented twice in the bmv2 JSON file, so it does look like a bmv2 JSON file that p4c-bm2-ss is generating that it should not, and the changes in this PR are correctly giving an error message about it.

PR #475, opened 2017-11-30
 Added stop_and_return API to simple switch to enable clean exit

dushyantarora13:
simple switch starts multiple worker threads and detaches them in start_and_return. This does not allow a clean exit on program termination.

### DIFF
--- a/targets/simple_switch/simple_switch.cpp
+++ b/targets/simple_switch/simple_switch.cpp
@@ -162,25 +162,15 @@ SimpleSwitch::receive_(int port_num, const char *buffer, int len) {
 void
 SimpleSwitch::start_and_return_() {
   check_queueing_metadata();
-  thread_exit_ = false;
-  thread_count_ = 0;
 
-  std::thread t1(&SimpleSwitch::ingress_thread, this);
-  t1.detach();
-  thread_count_ += 1;
+  threads_.push_back(std::thread(&SimpleSwitch::ingress_thread, this));
   for (size_t i = 0; i < nb_egress_threads; i++) {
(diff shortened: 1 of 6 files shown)


antoninbas:
Could you provide an example of a situation where exiting with detached threads still running is an issue?

I think an approach where the threads would not be detached at all would be better. You could push all the thread objects to a vector (class member). In your stop method, you would simply push back the `nullptr` signal to every queue and join all the threads in the vector. The threads would see the `nullptr` signal and break. You would not need a separate atomic `thread_exit_` signal or a `thread_cnt_` variable. What do you think?

dushyantarora13:
Terminating with detached threads causes memory checker to complain because of unclean exit. Here is the trace I get:
RAW: Memory was written to after being freed.  MallocBlock: 0x7f16be39d800, user ptr: 0x7f16be39d820, size: 1480. 
Stack trace:
  Write of size 8 at 0x7b6800004cb8 by main thread:
    #0 pthread_cond_destroy 
    #1 ~Queue behavioral_model/include/bm/bm_sim/queue.h
    #2 SimpleSwitch::~SimpleSwitch() behavioral_model/targets/simple_switch/simple_switch.h
    #3 SimpleSwitch::~SimpleSwitch() behavioral_model/targets/simple_switch/simple_switch.h
    #4 operator()     
    #5 ~unique_ptr 
    #6 sswitch_grpc::SimpleSwitchGrpcRunner::~SimpleSwitchGrpcRunner() behavioral_model/targets/simple_switch_grpc/switch_runner.cpp
    #7 at_exit_wrapper(void*) 
  Mutex M772221547724164240 is already destroyed.
I made changes like you said. Thanks for suggesting a much cleaner approach.

PR #484, opened 2017-12-07
Adding mirroring_mapping_add API to switch runner

dushyantarora:
Tests can use this API to mirroring session id to a physical port

### DIFF
--- a/targets/simple_switch_grpc/switch_runner.h
+++ b/targets/simple_switch_grpc/switch_runner.h
@@ -65,6 +65,9 @@ class SimpleSwitchGrpcRunner {
   int get_dp_grpc_server_port() {
     return dp_grpc_server_port;
   }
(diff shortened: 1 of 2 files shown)


targets/simple_switch_grpc/switch_runner.h
```diff
@@ -65,6 +65,7 @@ class SimpleSwitchGrpcRunner {
   int get_dp_grpc_server_port() {
     return dp_grpc_server_port;
   }
+  int mirroring_mapping_add(int mirror_id, int egress_port);
```
antoninbas: Please add a comment explaining that is temporary and only until P4 Runtime supports configuring mirroring sessions.

antoninbas:
@dushyantarora style check may fail if you don't write `TODO(unknown)` :)

PR #491, opened 2017-12-12
Don't set priority in test_grpc_counter

dushyantarora:
The table has an exact match field.


dushyantarora:
Sounds good! Thanks Antonin!

PR #499, opened 2017-12-18
Ensure all received packets have been processed

dushyantarora:
before returning from PacketStream rpc. The current implementation returns immediately if the client calls stream->WritesDone() even if there are packets being processed by simple_switch (we set active to false at the end of rpc and check for active in transmit_fn). While if the client does not call WritesDone() and waits in stream->Read() it might be blocked forever (say if the packet is dropped). Another corner case which might happen is if the client calls WritesDone() and before the packet is processed a new client connection sets active to true and resets this->stream. Now the transmit_fn will send old reply to new stream request.
In order to fix these issues we need to somehow ensure all the packets have been processed by simple_switch before returning from PacketStream rpc. For this end I added a utility function: SwitchWContexts::block_until_no_more_packets() which we call now via SimpleSwitchGrpcRunner instance before PacketStream rpc returns.

### DIFF
--- a/src/bm_sim/switch.cpp
+++ b/src/bm_sim/switch.cpp
@@ -425,17 +425,29 @@ SwitchWContexts::swap_requested() {
   return false;
 }
 
+void
+SwitchWContexts::block_until_no_more_packets() {
+  boost::unique_lock<boost::shared_mutex> lock(process_packet_mutex);
+  for (cxt_id_t cxt_id = 0; cxt_id < nb_cxts; cxt_id++) {
+    // Wait until no more packets exist for this context
+    while (phv_source->phvs_in_use(cxt_id) > 0) {
+      std::this_thread::yield();
+    }
+  }
+}
+
 int
 SwitchWContexts::do_swap() {
   int rc = 1;
   if (!enable_swap || !swap_requested()) return rc;
-  boost::unique_lock<boost::shared_mutex> lock(ongoing_swap_mutex);
+  boost::unique_lock<boost::shared_mutex> lock(process_packet_mutex);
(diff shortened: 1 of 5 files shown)


jafingerhut:
A given P4 program might drop every type of packet you can send to a device (not necessarily a useful P4 program, but a valid P4 program).

Even more useful P4 programs that drop some packets, and not others, without knowing something about the P4 program's behavior, you don't know which packets it will drop and which it will not.

Do you have some way to guarantee that the "dummy packet" you mention will not be dropped?

Does your proposed change rely on that dummy packet not being dropped?

dushyantarora:
@jafingerhut Thanks for the question. I don't actually process the dummy packet in the code. I just move it from one buffer queue to the next. The special thing about the dummy packet is that its ingress port = -1. I use this fact to identify it. 
The change does rely on the fact that it will not be dropped which I make sure by checking for the dummy packet after a new packet is dequeued.

dushyantarora:
Thanks for proposing a simpler alternative. I wasn't aware of the guarantees provided in 4. Glad to take a stab at this.

src/bm_sim/switch.cpp
```diff
@@ -425,17 +425,27 @@ SwitchWContexts::swap_requested() {
   return false;
 }
 
+void
+SwitchWContexts::block_until_no_more_packets(cxt_id_t cxt_id) {
```
antoninbas: You should stick with a function that does it for all contexts
- I cannot think of a use case where one would want to purge packets in a single context
- Because of the mutex, this function couldn't be used to implement  `block_until_no_more_packets_in_all_contexts`

dushyantarora: Yeah you are right. If we grab the mutex no new packets can be created in any context. Fixed.

PR #501, opened 2017-12-19
Support providing match key element name in JSON

### LINKED ISSUE #500: Support the ability to specify key names
The name of the key should be decoupled from the "target" field. This is necessary to support @name annotations on key fields in P4-16. 

The JSON representation could be something like:

```
"key" : [{
              "match_type" : "exact",
              "name" : "hdr.x.add",
              "target" : ["x_0", "add"],
              "mask" : null
            }]
```

antoninbas:
Introduced a new optional JSON attribute for match key fields,
'name'. In P4_16 a table can match on arbitary expressions and while the
compiler can evaluate these expressions upstream and match on a
temporary field holding the expression result, it is desirable to
preserve the orginal name.

As far as bmv2 itself is concerned, this information is only useful for
debugging, but this onformation could be required by control-planes.

version number was bumped up to 2.15

Fixes #500


antoninbas:
Merging this as this change is not going to break anything in itself.

PR #504, opened 2017-12-22
show log if register_read out of range

### LINKED ISSUE #503: better log for register index out of range
I defined `register<bit<8>>(1) r;`  and I access it by `r.read(x,999);`.
I will get following line in log file.
```
simple_switch: ../../include/bm/bm_sim/stateful.h:117: const bm::Register& bm::RegisterArray::operator[](size_t) const: Assertion `idx < size()' failed.
```

https://github.com/p4lang/behavioral-model/blob/d742940270595c28c5caec5f711a63a2979e1b43/include/bm/bm_sim/stateful.h#L115-L119

qin-nz:
I am new to C++, and I want to fix #503 

This PR prevent bmv2 quit because of access register_array out of range.
Instead, I just log critical.

### DIFF
--- a/include/bm/bm_sim/stateful.h
+++ b/include/bm/bm_sim/stateful.h
@@ -113,6 +113,7 @@ class RegisterArray : public NamedP4Object {
       bm::Logger::get()->critical(
         "Access register `{}` out of range, size={},index={}",
         name, size(), idx);
+      assert(idx < size());
     }
     return registers[idx];
   }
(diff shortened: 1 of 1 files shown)


qin-nz:
logs looks like:
```
[15:38:20.693] [bmv2] [C] [thread 14398] Access register `if_mac` out of range, size=3 index=3
```

without packet infomation.

qin-nz:
@antoninbas 
I add assert back. So this PR just log before terminate.

In my case, the index of register bmv2 read is from packet headers.
So I think in compile-time, compiler (almost) can't give a warning in my case.

antoninbas:
Superseded by https://github.com/p4lang/behavioral-model/pull/637

PR #511, opened 2017-12-28
We can now build against thrift-0.11.

liujed:
thrift-0.11 uses std::shared_ptr when built with C++11, and falls back
on boost otherwise. This now builds against the same shared_ptr provider
as Thrift.


antoninbas:
@liujed you may also want to update the README to list supported versions

liujed:
@antoninbas I updated the README  to just say that we support thrift 0.9.2 and later. Similarly for nanomsg. I've only tested on the latest (thrift 0.11.0 and nanomsg 1.1.2), however. Please let me know if you think we should be more precise.

krishnamohanelluru:
I still see compilation errors in server.cpp with this change. I have used 0.11.0 but didn't change nanomsg to 1.1.2
getting below error
git/src/bm_runtime/server.cpp:130: undefined reference to `apache::thrift::server::TThreadedServer::TThreadedServer(std::shared_ptr<apache::thrift::TProcessor> const&, std::shared_ptr<apache::thrift::transport::TServerTransport> const&, std::shared_ptr<apache::thrift::transport::TTransportFactory> const&, std::shared_ptr<apache::thrift::protocol::TProtocolFactory> const&, std::shared_ptr<apache::thrift::concurrency::ThreadFactory> const&)'
| collect2: error: ld returned 1 exit status
| Makefile:377: recipe for target 'simple_router' failed

liujed:
@krishnamohanelluru That's a linking error, probably caused by a mismatch between the Thrift headers you're compiling against and the Thrift library you're linking against. Can't really say much else without steps for reproducing the error.

PR #515, opened 2017-12-31
Implement P4_16 conformant stack support

antoninbas:
For header stacks and header union stacks. We support both behaviors at
the moment (legacy and P4_16). The legacy implementation is the default
one and will be the default for the 1.10 release. The P4_16 conformant
behavior can be toggled by using the `--enable-WP4-16-stacks` configure
flag. The plan is to have the new implementation be the default after
the 1.10 release.

The new implementation differs from the legacy one as follows:
push_front and pop_front shift the entire stack (not just until the next
index), and pushed headers are marked as invalid instead of valid.

We have tried to minimize the amount of code guarded by ifdef
directives. We used the CRTP to implement both behaviors and both
implementations are always compiled. We have unit tests for both
implementations as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -112,6 +112,14 @@ AC_ARG_ENABLE([Werror],
     AS_HELP_STRING([--enable-Werror], [Make all compiler warnings fatal]),
     [enable_Werror="$enableval"], [enable_Werror=no])
 
+AC_ARG_ENABLE([WP4-16-stacks],
+    AS_HELP_STRING([--enable-WP4-16-stacks],
+                   [Implement stacks strictly as per the P4_16 specification instead of legacy behavior]),
+    [enable_WP4_16_stacks="$enableval"], [enable_WP4_16_stacks=no])
+
+AS_IF([test "$enable_WP4_16_stacks" = "yes"],
+      [MY_CPPFLAGS="$MY_CPPFLAGS -DBM_WP4_16_STACKS"])
+
 # Checks for programs.
 AC_PROG_CXX
 AC_PROG_CC
--- a/include/bm/bm_sim/stacks.h
+++ b/include/bm/bm_sim/stacks.h
@@ -47,13 +47,13 @@ class StackIface {
   virtual ~StackIface() { }
 
   //! Removes the first element of the stack. Returns the number of elements
-  //! removed, which is `0` if the stack is empty and `1` otherwise. The second
-  //! element of the stack becomes the first element, and so on...
+  //! removed. The second element of the stack becomes the first element, and so
+  //! on...
   virtual size_t pop_front() = 0;
 
   //! Removes the first \p num element of the stack. Returns the number of
-  //! elements removed, which is `0` if the stack is empty. Calling this
-  //! function is more efficient than calling pop_front() multiple times.
+  //! elements removed. Calling this function is more efficient than calling
+  //! pop_front() multiple times.
   virtual size_t pop_front(size_t num) = 0;
 
   //! Pushes an element to the front of the stack. If the stack is already full,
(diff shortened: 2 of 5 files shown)


antoninbas:
@jafingerhut you may want to try out this patch before I merge it if you have some test P4 programs.

jafingerhut:
@antoninbas From this comment "The plan is to have the new implementation be the default after
the 1.10 release." do you mean that the plan is that this new behavior will become how it works for P4_14 programs compiled and run using bmv2, also?  I ask because the new behavior of push_front with these changes makes the newly-inserted headers invalid, whereas the P4_14 v1.0.x specs up to v1.0.4 currently say that the newly-inserted headers should be valid.  If the plan is to modify P4_14 behavior that existing programs may rely upon, then I won't stop you, but it seems worth asking more people about who may be relying upon the existing behavior.

antoninbas:
It means that starting with release 1.11, bmv2 users that wish to have the old P4_14 behavior will need to configure bmv2 with `--enable-WP4-16-stacks=no`.

jafingerhut:
Understood.  Thanks.

antoninbas:
To give some more insight in my reasoning:
- supporting both behaviors simultaneously in a dynamic fashion would require including additional information in the JSON, which I am reluctant to do.
- it seems to me that we are trying to promote p4lang/p4c as the de-facto compiler for bmv2 (thus deprecating the old one). When compiling P4_14 programs, my understanding is that p4c will insert the necessary `setValid` calls when pushing new headers which means that the new behavior (`enable-WP4-16-stacks=yes`) will work for both P4_14 and P4_16 programs as long as they are compiled with p4c. Please let me know if this is not correct. Starting with 1.11, if one wishes to use the old p4c-bm compiler, one will need to run `configure` with `--enable-WP4-16-stacks=no` (I will update the README before the release). This may serve as an incentive for people to transition to p4c instead.

antoninbas:
Actually I didn't know about this PR: https://github.com/p4lang/p4-spec/pull/501. Your question makes more sense to me. If the P4_14 spec changes I would expect everyone to update their P4 programs.

I am not sure if p4lang/p4c is expected to support both versions of P4_14 (v1.0.4 and v1.0.5). If it does support both, I imagine it would do so through a command-line flag. When doing the P4_14 -> P4_16 translation, p4c will have to insert `setValid` calls for v1.0.4 (and do nothing for v1.0.5). I may do something similar for the legacy p4c-bm compiler (add a flag and insert primitive calls for v1.0.4).

A few remarks:
- We cannot infer from the source the P4_14 version, so if a compiler wants to support both versions, we need a flag / a pragma / ...
- I don't think it is good practice to tie the compiler version to the language version, i.e. I don't think we should have compiler version _X_ support only v1.0.4 and compiler version _Y_ support only v1.0.5. Ideally compiler version _Y_ should support both.

jafingerhut:
Thanks again for explanations.  My guess was that there would be a desire _not_ to change the behavior of push and pop on header stacks in P4_14 from v1.0.4 to v1.0.5, only to change the P4_14 v1.0.5 spec so that add_header and remove_header are spec'd to behave as behavioral-model already does (i.e. eliminate the push-in-middle-of-stack / pop-in-middle-of-stack behavior from the spec that they currently describe).

jafingerhut:
I have created a P4_16 program with STF tests for the p4lang/p4c repo here: https://github.com/p4lang/p4c/pull/1129

While writing and trying out those new tests, I found a p4c bug there in how header stacks are copied between controls and sub-controls, at least in some P4_16 programs, mentioned but without example code yet, here: https://github.com/p4lang/p4c/issues/1128  That bug only exhibits itself with the legacy implementation of push/pop in P4_16 programs, I believe.

jafingerhut:
@jnfoster For changes to P4_14 language spec, changing definition of add/remove header to match what the open source tools have been doing for years, and leaving definition of push/pop alone, seems like the right thing to me.

With the changes in this PR for p4lang/behavioral-model, before these changes become the default behavior, it would be good to change p4lang/p4c so that when compiling P4_14 source code, it generates BMv2 JSON files that do the new 'push' primitive (which makes the N newly pushed headers invalid), followed by N primitive calls that make the newly-pushed headers become valid, so that P4_14 programs continue to behave as they have in the past.  The same goes for invoking p4lang/p4c to auto-translate P4_14 to P4_16 source code.

PR #518, opened 2018-01-05
Copy the field lists between cloned and original packet after parsing

dushyantarora:
has been done on cloned packet. This is needed because the copied field_list fields can get overwritten during parsing. 
To give an example this fix will prevent something like this:

```
struct M {
  bit<9> port;
};

parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t smeta) {
    state start { 
      meta = {0};
      transition accept;
    }
}

control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply {
        meta.port = smeta.egress_spec;
        clone3(CloneType.I2E, 64, {meta.port});
    }
};
```
The cloned pack will have metadata value meta.port = 0 in current implementation instead of the value of smeta.egress_spec

### DIFF
--- a/targets/simple_switch/simple_switch.cpp
+++ b/targets/simple_switch/simple_switch.cpp
@@ -282,17 +282,16 @@ SimpleSwitch::enqueue(int egress_port, std::unique_ptr<Packet> &&packet) {
 }
 
 // used for ingress cloning, resubmit
-std::unique_ptr<Packet>
-SimpleSwitch::copy_ingress_pkt(
+void
+SimpleSwitch::copy_field_list_and_set_type(
     const std::unique_ptr<Packet> &packet,
+    const std::unique_ptr<Packet> &packet_copy,
     PktInstanceType copy_type, p4object_id_t field_list_id) {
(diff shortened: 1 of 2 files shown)


targets/simple_switch/simple_switch.cpp
```diff
@@ -378,12 +378,17 @@ SimpleSwitch::ingress_thread() {
             packet->save_buffer_state();
         packet->restore_buffer_state(packet_in_state);
         p4object_id_t field_list_id = clone_spec >> 16;
-        auto packet_copy = copy_ingress_pkt(
```
antoninbas: The comment attached to the `copy_ingress_pkt` definition is not valid any more with this change.
Maybe you should remove `copy_ingress_pkt` altogether now that it is only used for resubmit, or try to find a smaller common denominator for ingress cloning & resubmit to avoid duplicating code.

dushyantarora: Thanks for the comment. Found a smaller common denominator.

PR #522, opened 2018-01-09
Add new core assignment primitives for stacks

antoninbas:
Added assign_header_stack for header stack assignment and
assign_union_stack for header union stack assignment. Stack copies are
valid in P4_16. It is not possible for the compiler to generate
equivalent code by copying each stack element individually as stack have
internal state (the next "pointer"). Having dedicated bmv2 primitives is
also more efficient.

version number was bumped up to 2.17

### DIFF
--- a/docs/JSON_format.md
+++ b/docs/JSON_format.md
@@ -10,7 +10,7 @@ on each attribute.
 
 ## Current bmv2 JSON format version
 
-The version described in this document is *2.16*.
+The version described in this document is *2.17*.
 
 The major version number will be increased by the compiler only when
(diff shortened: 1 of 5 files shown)


antoninbas:
@jafingerhut @mbudiu-vmw maybe you want to take a look
this should help with https://github.com/p4lang/p4c/issues/1128#issuecomment-356170823

jafingerhut:
Looks like what I would expect.  I haven't tested it yet, but the test cases I've written and included on some p4c issues should exercise the one for header stacks (the non-union one, at least), once the p4c-bm2-ss bmv2 backend has been modified to take advantage of these new bmv2 primitives.

PR #527, opened 2018-01-15
Call DeviceMgr::init in SwitchRunner constructor

antoninbas:
This is necessary now that we call DeviceMgr::destroy in the
SwitchRunner destructor. Otherwise, we could have a segfault if
init_and_start returns prematurely. The alternative would be to have yet
another flag... but I see no harm in calling DeviceMgr::init in the
constructor.


antoninbas:
@dushyantarora do you see any issue with this change?

dushyantarora:
Sorry for the late reply. Yesterday was a holiday for us. Change LGTM.

PR #533, opened 2018-01-20
Fix simple_switch_grpc direct counter GTest

antoninbas:
Because PSA mandates that a given table can have at most one of each
direct resource, we have remove the id from the DirectCounterEntry &
DirectMeterEntry messages in P4Runtime (the id can be inferred from the
table id, included in the TableEntry field). As a result one of the
simple_switch_grpc unit tests had to be updated to fix a compialtion
issue.


antoninbas:
See https://github.com/p4lang/PI/issues/266

PR #546, opened 2018-02-05
Remove dependency on empty.proto in simple_switch_grpc

### LINKED ISSUE #545: Build simple_switch_grpc error.
Compiler:  `gcc version 7.2.0`
protoc: `3.5.0`
grpc: `v1.8.x`

After I compiler and install PI and bmv2 with guideline
( install `libyang` and `sysrepo`  for `PI` , and using `./configure --with-bmv2 --with-proto --with-sysrepo` to configure, and in `bmv2` using `./configure --enable-debugger --with-pi` )
Then I go to [`targets/simple_switch_grpc`](https://github.com/p4lang/behavioral-model/tree/master/targets/simple_switch_grpc) to build `simple_switch_grpc`

Build with command `make` after specified `./configure --with-sysrepo`
I found there have an error:

```
/home/kevin/tools/bmv2/targets/simple_switch_grpc/.libs/libbm_grpc_dataplane.so: undefined reference to `google::protobuf::Empty::~Empty()'
/home/kevin/tools/bmv2/targets/simple_switch_grpc/.libs/libbm_grpc_dataplane.so: undefined reference to `google::protobuf::Empty::Empty()'
/home/kevin/tools/bmv2/targets/simple_switch_grpc/.libs/libbm_grpc_dataplane.so: undefined reference to `protobuf_google_2fprotobuf_2fempty_2eproto::AddDescriptors()'
collect2: error: ld returned 1 exit status
make[2]: *** [simple_switch_grpc] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```
Does this problem occur because of my grpc or `protobuf` dependencies? 
I found in `PI` said need to use `grpc: v1.3.2` and `protobuf: v3.2.0`.

Thanks for any help.

antoninbas:
Fixes #545.
I do not observe a linking error on my system, where
google::protobuf::Empty is part of libprotobuf. It is possible that it
is not the case on other systems or with different versions (more
recent?) of protobuf.

### DIFF
--- a/targets/simple_switch_grpc/p4/bm/dataplane_interface.proto
+++ b/targets/simple_switch_grpc/p4/bm/dataplane_interface.proto
@@ -16,8 +16,6 @@ syntax = "proto3";
 
 // This package and its contents are a work-in-progress.
 
-import "google/protobuf/empty.proto";
-
 package p4.bm;
 
 service DataplaneInterface {
@@ -29,7 +27,7 @@ service DataplaneInterface {
   // ports are valid and "UP"). This is useful in the context of testing, to
   // simulate port-up / port-down events.
   rpc SetPortOperStatus(SetPortOperStatusRequest)
-      returns (google.protobuf.Empty) {
+      returns (SetPortOperStatusResponse) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


antoninbas:
@dushyantarora any issue with this?

dushyantarora:
LGTM

PR #551, opened 2018-02-09
Support "default-default" entry for tables with ability to reset

antoninbas:
This commit introduces the following:
  * support for compile-time default entry for indirect tables; until
  now only direct tables could have compile-time default entries. Unlike
  a control-plane provided default entry, the compile-time default entry
  for indirect tables uses action + action data (not member / group
  handle)
  * ability to reset default entry to initial compile-time default entry
  (aka default-default entry). The Thrift IDL has been updated as well
  as the runtime_CLI, with support in P4 Runtime to follow.

These changes should be able to accomodate both P4_14 & P4_16 and both
bmv2 compilers.

I believe I also found a possible race condition when indirect match
tables share the same action profile and it should be fixed now.

### DIFF
--- a/docs/JSON_format.md
+++ b/docs/JSON_format.md
@@ -546,7 +546,7 @@ attributes for these objects are:
     - `action_id`: the id of the default action
     - `action_const`: an optional boolean value which is `true` iff the control
     plane is not allowed to change the default action function. Default value is
-    `false`.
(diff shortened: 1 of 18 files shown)


antoninbas:
I found more race conditions in the action profile implementation which for some reason I have completely missed in the last year. So this PR ended-up being much more than supporting default-default entry...

PR #559, opened 2018-03-06
Include direct resources in pi_table_fetch result

antoninbas:
We leverage the new format for the fetch result to include direct
resources' (counters & meters) configs.

We also include a GTest unit test for simple_switch_grpc, which reads a
direct counter value using TableEntry.


antoninbas:
Depends on https://github.com/p4lang/PI/pull/321

PR #565, opened 2018-03-16
Changed port data type to uint32_t everywhere in bmv2

dushyantarora:
Using port_t (uint32_t) everywhere where port is used.


targets/simple_switch_grpc/switch_runner.h
```diff
@@ -49,9 +49,9 @@ class SimpleSwitchGrpcRunner {
   // there is no real need for a singleton here, except for the fact that we use
   // PIGrpcServerRunAddr, ... which uses static state
   static SimpleSwitchGrpcRunner &get_instance(
-      int max_port = 512, bool enable_swap = false,
+      bm::DevMgrIface::port_t max_port = 512, bool enable_swap = false,
       std::string grpc_server_addr = "0.0.0.0:50051",
-      int cpu_port = -1,
+      bm::DevMgrIface::port_t cpu_port = -1,
```
antoninbas: You kept -1 as the default value here, but changed it to 0 below?

targets/simple_switch_grpc/switch_runner.cpp
```diff
@@ -294,15 +295,16 @@ SimpleSwitchGrpcRunner::init_and_start(const bm::OptionsParser &parser) {
 
   // check if CPU port number is also used by --interface
   // TODO(antonin): ports added dynamically?
-  if (cpu_port >= 0) {
+  if (cpu_port > 0) {
```
antoninbas: it seems that 0 is still a valid value for dataplane ports (at least I hope it is the case because many bmv2 users rely on this), but that it is no longer a valid value for the cpu_port for simple_switch_grpc
in this case I believe you should update main.cpp to 1) indicate in the help string for --cpu-port that 0 is not valid, 2) return an error if the user has passed a value of 0, and 3) probably add a static_cast when `SimpleSwitchGrpcRunner::get_instance` to make the intent clear.

targets/simple_switch/simple_switch.h
```diff
@@ -95,14 +95,14 @@ class SimpleSwitch : public Switch {
     return mirroring_map.erase(mirror_id);
   }
 
-  int mirroring_mapping_get(mirror_id_t mirror_id) const {
-    return get_mirroring_mapping(mirror_id);
+  bool mirroring_mapping_get(mirror_id_t mirror_id, port_t* port) {
```
antoninbas: style consistency: s/`port_t* port`/`port_t *port`

targets/simple_switch/simple_switch.h
```diff
@@ -142,16 +142,19 @@ class SimpleSwitch : public Switch {
   void egress_thread(size_t worker_id);
   void transmit_thread();
 
-  int get_mirroring_mapping(mirror_id_t mirror_id) const {
+  bool get_mirroring_mapping(mirror_id_t mirror_id, port_t* port) {
```
antoninbas: style consistency: s/`port_t* port`/`port_t *port`

targets/simple_switch_grpc/switch_runner.cpp
antoninbas: alternatively you can use `boost::optional<uint32_t>` in the `SimpleSwitchGrpcRunner` interface if you think `boost::optional` is an acceptable dependency and keep 0 as a valid value

dushyantarora: 0 is still valid for data plane ports. In fact it is the default value for incoming/outgoing port in dataplane_interface.proto. This means that we cannot use 0 as a cpu_port. I updated main.cpp accordingly.

targets/simple_switch_grpc/switch_runner.h
dushyantarora: I missed this one. Fixed.

targets/simple_switch/simple_switch.h
dushyantarora: Fixed

targets/simple_switch/simple_switch.h
dushyantarora: Fixed

dushyantarora:
I understand. I can wait until the release it cut.

include/bm/bm_sim/target_parser.h
```diff
@@ -108,6 +113,9 @@ class TargetParserBasic final : public TargetParserIface {
   //! Retrieve the value of command-line option `--<name>`. The option needs to
   //! have been previously registered with add_int_option().
   ReturnCode get_int_option(const std::string &name, int *v) const;
+  //! Retrieve the value of command-line option `--<name>`. The option needs to
+  //! have been previously registered with add_uint_option().
+  ReturnCode get_uint_option(const std::string &name, uint32_t *v) const;
```
antoninbas: shouldn't this be `unsigned int`? I know you are using `uint32_t` for the port, but using it here doesn't seem very consistent with `[get|add]_int_option` or even with your comment for `add_uint_option`. If you really want to use `uint32_t` here, maybe you should call the methods `add_uint32_option` and `get_uint32_option`.

targets/simple_switch/simple_switch.h
```diff
@@ -95,14 +95,14 @@ class SimpleSwitch : public Switch {
     return mirroring_map.erase(mirror_id);
   }
 
-  int mirroring_mapping_get(mirror_id_t mirror_id) const {
-    return get_mirroring_mapping(mirror_id);
+  bool mirroring_mapping_get(mirror_id_t mirror_id, port_t *port) {
```
antoninbas: by the way it doesn't seem right to be dropping the `const` here?

targets/simple_switch/simple_switch.h
```diff
@@ -142,16 +142,19 @@ class SimpleSwitch : public Switch {
   void egress_thread(size_t worker_id);
   void transmit_thread();
 
-  int get_mirroring_mapping(mirror_id_t mirror_id) const {
+  bool get_mirroring_mapping(mirror_id_t mirror_id, port_t *port) {
```
antoninbas: same for this `const`?

dushyantarora: Fixed

targets/simple_switch/simple_switch.h
dushyantarora: Fixed

include/bm/bm_sim/target_parser.h
dushyantarora: Made it unsigned it.

Issue #541, opened 2018-01-26
std::out_of_range() exception thrown when egress_spec over 255


4tXJ7f:
When a packet has egress_spec > 255 after ingress, an `std::out_of_range()` exception gets thrown. It looks like this is because `egress_buffers` use a `max_port` of 256. In [`SimpleSwitch::enqueue()`](https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L276-L280) `push_front()` is called, which seems to lead to an out-of-bound access [here](https://github.com/p4lang/behavioral-model/blob/327a4029a0dc7ae505973d990bc4a29c9cc4c054/include/bm/bm_sim/queueing.h#L251). The special port 511 gets filtered out [here](https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L447). While this is not a critical issue, the [PSA Section 6.2](https://p4.org/p4-spec/docs/PSA.html#sec-after-ingress) does seem to say that packets with an invalid port should just get dropped (last else branch in pseudo code).

jafingerhut:
And if simple_switch is enhanced to drop such packets, it would probably be nice to have a distinct message in the log, when logging is enabled, that gives a message similar to: "egress_spec 256 does not correspond to any port.  Dropping packet."

antoninbas:
Fixed by #571

PR #571, opened 2018-03-24
Cleanup code in simple_switch and simple_switch_grpc targets

dushyantarora:
1) Create one extra queue for cpu_port if it is greater than max_port: queue id max_port.
2) Added check for invalid port in SimpleSwitch::enqueue()
3) Cleaned up unused function in simple_switch_grpc/tests/utils.cpp

### DIFF
--- a/targets/simple_switch/simple_switch.cpp
+++ b/targets/simple_switch/simple_switch.cpp
@@ -177,7 +177,7 @@ SimpleSwitch::reset_target_state_() {
 }
 
 int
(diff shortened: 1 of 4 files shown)


dushyantarora:
I understand your concern. I think I will wait for P4Runtime port translation support.
If I read it correctly #541 is not advocating for being able to choose arbitrary integer values for ports; instead just saying that SimpleSwitch should drop packets with invalid egress_port instead of throwing std::out_of_range() exception. This commit will fix the issue.

PR #572, opened 2018-03-28
Fix crash in PI implementation when reading default meter rates

antoninbas:
If a meter entry was not initialized, bmv2 returns an empty vector of
rates. We handle this case gracefully by setting information rates and
burst sizes to -1, which are the default values as per the P4Runtime
spec.


antoninbas:
@dushyantarora this fixes the issue you reported, please take a look

dushyantarora:
Can we add a change in PI/proto/frontend/src/device_mgr.cpp after this to not send meter config instead of sending -1? It should be an easy change I would think and we will maintain read/write symmetry as well. Also, p4runtime.proto says:
  // Table read:
  //   - If the meter_config field has default value, it will not be set in the
  //     read table entry.
  MeterConfig meter_config = 6;

targets/simple_switch_grpc/tests/test_meter.cpp
```diff
@@ -0,0 +1,117 @@
...
+    EXPECT_TRUE(status.ok());
+    EXPECT_EQ(read_response.entities_size(), 1);
+    const auto &read_entry = read_response.entities(0).table_entry();
+    ASSERT_TRUE(read_entry.has_meter_config());
+    const auto &meter_config = read_entry.meter_config();
+    EXPECT_EQ(meter_config.pir(), -1);
+    EXPECT_EQ(meter_config.pburst(), -1);
+    EXPECT_EQ(meter_config.cir(), -1);
+    EXPECT_EQ(meter_config.cburst(), -1);
+  }
```
dushyantarora: Can we also test reading of DirectMeterEntry here?

antoninbas:
@dushyantarora 
p4runtime.proto does say:
```
// Table read:
// - If the meter_config field has default value, it will not be set in the
// read table entry.
```
However my understanding is that it means that if the `meter_config` field is not set in the `TableEntry` used in the `ReadRequest`, then the `TableEntry` in the `ReadResponse` will use a default `meter_config` field (i.e. unset) even if the meter was configured by the controller (i.e. it's the controller's way of saying "I'm not interested in the meter configuration value" for this Read).

Basically I interpret "default value" in "meter_config field has default value" as meaning the protobuf field is unset.

From a conversation with Waqar and Samar (https://github.com/p4lang/PI/pull/186#discussion_r133998203): "when an entry is added to a table with a direct meter array and no configuration is provided, the default configuration [-1 for rates and burst sizes] will be used to program the hardware and it will be the configuration returned when doing a Read operation."

Maybe you should work with Samar to clarify the comment. As far as I can tell there is nothing preventing us from having both:
- if the `meter_config` field is unset in the request, the response will not include a `meter_config` field (i.e. it will also be unset in the response)
- if the `meter_config` field is set in the request but the hardware meter configuration value is the P4Runtime default value for a meter configuration (i.e. always green), then the response will not include a `meter_config` field (i.e. it will be unset in the response)

For counters:
- if the `counter_data` field is unset in the request, the response will not include a `counter_data` field (i.e. it will also be unset in the response)
- if the `counter_data` field is set in the request, the response will ALWAYS include a `counter_data` field.

Of course this assumes that the table actually has a direct counter and / or meter.

Please review these statements with Samar. If necessary we can have a quick call tomorrow.

dushyantarora:
Thanks for the detailed explanation Antonin. Really appreciate it! Lets submit this PR for now so I can make progress. Changes(if any) would be required in p4lang/PI anyway.

antoninbas:
Sounds good

dushyantarora:
Thanks for fixing the bug!

PR #594, opened 2018-04-24
Use correct PI ID prefix for direct counters & meters

antoninbas:
After update to PI and p4c. The JSON and p4info files used for testing
simple_switch_grpc have also been updated.


antoninbas:
Depends on https://github.com/p4lang/PI/pull/347

PR #604, opened 2018-05-21
Add macros to easily register extern functions

omeranson:
Add macros BM_REGISTER_EXTERN_FUNCTION and BM_REGISTER_EXTERN_W_NAME_FUNCTION
(similar to BM_REGISTER_EXTERN_METHOD) to facilitate the registration
of standalone functions as extern functions.

### DIFF
--- a/tests/test_extern.cpp
+++ b/tests/test_extern.cpp
@@ -24,6 +24,7 @@
 #include <bm/bm_sim/P4Objects.h>
 
 #include <string>
+#include <utility>
 #include <vector>
 #include <cassert>
 
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
 BM_REGISTER_EXTERN_METHOD(ExternExpression, set_var2, const Data &);
 BM_REGISTER_EXTERN_METHOD(ExternExpression, execute, Data &);
 
+void set_field1(Field &v) {
+  v.set(1);
+}
+
+void set_field2(Field &v) {
+  v.set(2);
+}
+
+void set_field3(Field &v) {
+  v.set(3);
+}
+
+BM_REGISTER_EXTERN_FUNCTION(set_field1, Field &);
+BM_REGISTER_EXTERN_FUNCTION_W_NAME(set_field2_test, set_field2, Field &);
+BM_REGISTER_EXTERN_FUNCTION_W_NAME(BM_EXTERN_ADD_EXTERN_PREFIX(set_field3),
+                                   set_field3, Field &);
+
 namespace {
 
 // Google Test fixture for extern tests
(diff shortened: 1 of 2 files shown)


omeranson:
> When would one want to use a standalone function instead of defining a functor?

I found this syntax more convenient, and more in line with the other macros in extern.h. This way, I can only define a single function.

But I agree, it is only a convenience.

omeranson:
>    use the actual name passed to BM_REGISTER_EXTERN_W_NAME_FUNCTION or BM_REGISTER_EXTERN_FUNCTION. Prepending __extern_ is fine for the functor definition, but the name should be the one provided by the person calling the macro IMO

Done

>    add a unit test in tests/test_actions.cpp or tests/test_externs.cpp to make sure that the macro work as expected and to ensure it doesn't break in the future

Done

tests/test_extern.cpp
```diff
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
 BM_REGISTER_EXTERN_METHOD(ExternExpression, set_var2, const Data &);
 BM_REGISTER_EXTERN_METHOD(ExternExpression, execute, Data &);
 
+void set_field1(Field & v) {
```
antoninbas: formatting: s/`Field & v`/`Field &v`

tests/test_extern.cpp
```diff
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
...
+  v.set(1);
+}
+
+void set_field2(Field & v) {
```
antoninbas: same

tests/test_extern.cpp
```diff
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
...
+  v.set(2);
+}
+
+void set_field3(Field & v) {
```
antoninbas: same

tests/test_extern.cpp
```diff
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
...
+  v.set(3);
+}
+
+BM_REGISTER_EXTERN_FUNCTION(set_field1, Field&);
```
antoninbas: formatting: s/`Field&`/`Field &`

tests/test_extern.cpp
```diff
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
...
+BM_REGISTER_EXTERN_FUNCTION_W_NAME(set_field2_test, set_field2, Field&);
```
antoninbas: same

tests/test_extern.cpp
```diff
@@ -168,6 +169,23 @@ BM_REGISTER_EXTERN(ExternExpression);
...
+BM_REGISTER_EXTERN_FUNCTION_W_NAME(BM_EXTERN_ADD_EXTERN_PREFIX(set_field3),
+                                   set_field3, Field&);
```
antoninbas: same

PR #605, opened 2018-05-21
Add option --load-modules to simple_switch

omeranson:
Add option --load-modules to allow loading external code as .so files.
This allows externs to be registered, e.g. using REGISTER_PRIMITIVE or
BM_REGISTER_EXTERN_METHOD from external code, with changing
behavioral-model's code.

The .so files are comma-delimited, e.g. -- --load-modules
file1.so,file2.so .

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,6 +213,32 @@ AC_LINK_IFELSE(
 
 AM_CONDITIONAL([WITH_PCAP_FIX], [test "$pcap_fix" = "yes"])
 
+# Check for dlopen and dlfcn.h
+AC_ARG_ENABLE(modules,
+    AS_HELP_STRING([--enable-modules],
+                   [Allow loading third-party modules at runtime]))
+modules_enabled=no
+AS_IF([test "x$enable_modules" != "xno"], [
+  AC_MSG_CHECKING(for dlopen())
+  AC_CHECK_HEADERS(dlfcn.h, [
+    AC_SEARCH_LIBS([dlopen], [dl], [
+      MY_CPPFLAGS="$MY_CPPFLAGS -DBM_HAVE_DLOPEN -DBM_ENABLE_MODULES"
+      modules_enabled=yes
+    ], [
+      AC_MSG_RESULT(no)
+      AS_IF([test "x$enable_modules" == "xyes"], [
+        AC_MSG_ERROR([Cannot enable modules without dlopen])
+      ])
+    ])
+  ],[
+    AC_MSG_RESULT(no)
+    AS_IF([test "x$enable_modules" == "xyes"], [
+      AC_MSG_ERROR([Cannot enable modules without dlfcn.h])
+    ])
+  ])
+])
+AM_CONDITIONAL([ENABLE_MODULES], [test "x$modules_enabled" == "xyes"])
+
 # C++ libraries are harder (http://nerdland.net/2009/07/detecting-c-libraries-with-autotools/),
 # so use headers to check
 AC_CHECK_HEADER([boost/thread.hpp], [], [AC_MSG_ERROR([Boost threading headers not found])])
--- a/targets/simple_switch/main.cpp
+++ b/targets/simple_switch/main.cpp
@@ -20,7 +20,9 @@
 
 /* Switch instance */
 
-#include <dlfcn.h>
+#ifdef BM_HAVE_DLOPEN
+#  include <dlfcn.h>
+#endif  // BM_HAVE_DLOPEN
 #include <bm/SimpleSwitch.h>
 #include <bm/bm_runtime/bm_runtime.h>
 #include <bm/bm_sim/target_parser.h>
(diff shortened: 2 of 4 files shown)


omeranson:
> it lets you change the set of supported externs at runtime, which can be confusing as it violates the assumption that one bmv2 binary == one P4_16 architecture

I chose this behaviour to allow developers to write new externs without having to recompile bmv2. e.g. a simulation of an extern provided by hardware, or testing an extern design before implementing it in a more expensive environment. In effect, changing the set of supported externs at runtime is the goal.

I think this will facilitate development. It will also allow third-party developers to provide externs for experimentation, testing, and validation on bmv2.

> this code is very platform dependent, yet no precaution was taken in the code to gracefully handle the case where dlfcn.h / dlopen is not available

I'll wrap it in an #ifdef. I'll see if I can get it to be set automatically by the autoconf tools.

targets/simple_switch/main.cpp
```diff
@@ -20,15 +20,59 @@
...
+
 #include "simple_switch.h"
 
 namespace {
 SimpleSwitch *simple_switch;
-bm::TargetParserBasic *simple_switch_parser;
+
+std::string load_modules_option = "load-modules";
+
+class TargetParserModules : public bm::TargetParserBasic {
```
antoninbas: I'm not sure why inheritance is needed here. It seems you can just use `TargetParserBasic` as it was intended:
```
simple_switch_parser->add_string_option(load_modules_option, "load the given .so files as modules");
std::string modules;
auto retval = get_string_option(load_modules_option, &modules);
...
load_modules(modules);
```
And if you do think it is better to create a new object that inherits from `TargetParserBasic`, I think it would make sense to take care of ALL target-specific options there.

targets/simple_switch/main.cpp
```diff
@@ -20,15 +20,59 @@
...
+
+  int parse(const std::vector<std::string> &more_options,
+                    std::ostream *errstream) override {
+    int result = ::bm::TargetParserBasic::parse(more_options, errstream);
+    load_modules();
+    return result;
+  }
+
+ protected:
+  virtual int load_modules() {
```
antoninbas: I may be missing something but why is this `virtual`?

targets/simple_switch/main.cpp
omeranson: > I'm not sure why inheritance is needed here. It seems you can just use `TargetParserBasic` as it was intended:

The modules have to be loaded before the switch is initialized. simple_switch verifies the loaded json file. If it has external calls that are provided by the module, sinple_switch will exit unless the module is loaded first.

By overriding `TargetParserBasic`, I am able to inject code after the command line arguments are parsed, but before the switch is initialized.

> And if you do think it is better to create a new object that inherits from TargetParserBasic, I think it would make sense to take care of ALL target-specific options there.

I'll do that.

targets/simple_switch/main.cpp
omeranson: In case anyone wants to override this class. If you prefer, I can remove it, and if anyone needs it, they can add it later.

omeranson:
> If you want to contribute to this repo, please confirm that you work for an organization that is a p4.org member.

I work for HuaWei. I can see they are a member in the P4 member's page.

configure.ac
```diff
@@ -213,6 +213,27 @@ AC_LINK_IFELSE(
 
 AM_CONDITIONAL([WITH_PCAP_FIX], [test "$pcap_fix" = "yes"])
 
+# Check for dlopen and dlfcn.h
+AC_ARG_ENABLE(modules,
+    AS_HELP_STRING([--enable-modules],
+                   [Allow loading third-party modules at runtime]))
+AS_IF([test "x$enable_modules" == "xyes"], [
+  MY_CPPFLAGS="$MY_CPPFLAGS -DBM_ENABLE_MODULES"
```
antoninbas: I believe "dynamic module loading" should be enabled by default if the dlopen dependencies are met. I don't see any drawback in doing this and for most users it may mean one less configure knob that they have to worry about it. The code which is guarded by the preprocessor flags would also be less likely to break if it's built by default. Thoughts?

targets/simple_switch/main.cpp
```diff
@@ -20,15 +20,81 @@
...
+
 #include "simple_switch.h"
 
 namespace {
 SimpleSwitch *simple_switch;
-bm::TargetParserBasic *simple_switch_parser;
+
+std::string load_modules_option = "load-modules";
+
+class TargetParserModules : public bm::TargetParserBasic {
```
antoninbas: Suggest renaming this to `SimpleSwitchParser` now that this class is meant to regroup all of simple_switch command-line options.

targets/simple_switch/main.cpp
```diff
@@ -20,15 +20,81 @@
 
 /* Switch instance */
 
+#ifdef BM_HAVE_DLOPEN
```
antoninbas: is there actually a need for 2 separate processor flags here, or based on the configure `BM_ENABLE_MODULES` => `BM_HAVE_DLOPEN` (otherwise configure would fail).

configure.ac
omeranson: I definitely prefer it on by default as well. I was afraid you'd oppose the dlopen requirement,  but setting the default to be enabled if dlopen is already available sounds like the right way to go. Will change.

targets/simple_switch/main.cpp
omeranson: Will do.

targets/simple_switch/main.cpp
omeranson: This was in preparation in case another method other than dlopen would allow this same behaviour, e.g. Windows' LoadLibrary.

I can simplify if you prefer.

antoninbas: no need, sounds good

targets/simple_switch/main.cpp
```diff
@@ -20,15 +20,81 @@
...
+    }
+    if (retval != ReturnCode::SUCCESS) {
+      return -1;  // Unexpected error
+    }
+    std::istringstream iss(modules);
+    std::string module;
+    while (std::getline(iss, module, ',')) {
+#  ifdef BM_HAVE_DLOPEN
+      if (!dlopen(module.c_str(), RTLD_NOW | RTLD_GLOBAL)) {
+        std::clog << "WARNING: Skipping module: " << module << ": "
```
antoninbas: Just one more thing. Is it possible to print this to `errstream` (as passed to the `parse` function) to be consistent with the rest of the option parsing methods?

omeranson: Sure. Will do.

PR #608, opened 2018-05-25
Update simple_switch_grpc tests to use new P4Runtime proto package names

antoninbas:
p4 and p4.config have been renamed p4.v1 and p4.config.v1 to follow best
practices in deploying protobuf / gRPC cloud services (package name
should include major version number if non-backward compatible changes
are likely to be introduced in the future).

To be consistent with the P4Runtime server implementation code, we use
namespace aliases p4v1 and p4configv1 in implementation files.


antoninbas:
Depends on https://github.com/p4lang/PI/pull/356

PR #617, opened 2018-06-11
Updated install_deps.sh for Ubuntu 18.04

BenRLewis:
Updated dependencies to compile on 18.04

Thrift will not compile with libssl-dev as provided in 18.04 which is version 1.1.0. By adding a check into install_deps.sh, the right version is installed. 

Tests pass when running with make check.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -1,5 +1,13 @@
 #!/bin/sh
 set -e
+# Check whether running Ubuntu >= 18.04
+if test `lsb_release -rs` = '18.04';
+then
+    echo "Detected Ubuntu 18.04, using libssl1.0-dev"
+    LIBSSL="libssl1.0-dev"
+else
+    LIBSSL="libssl-dev"
+fi
 sudo apt-get install -y \
     automake \
     cmake \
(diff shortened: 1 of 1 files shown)


BenRLewis:
I have sent off my signed membership form/agreement to membership@p4.org but haven't had confirmation yet.

install_deps.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/sh
 set -e
+# Check whether running Ubuntu >= 18.04
+if test `lsb_release -rs` = '18.04';
```
jafingerhut: Just a minor comment that when Ubuntu 18.10 and 19.04 etc. are released, most likely they will continue to work with this new value for LIBSSL, but this script will use the older package name for those.  Changing the 'if' condition to something more like this might be good:
```
if [[ `lsb_release -rs` > "18" ]]
```

jafingerhut:
@BenRLewis I get an error trying to compile thrift version 0.9.2 on Ubuntu 18.04, with or without this change.  Version 0.9.2 is the version of thrift currently used by the travis/install-thrift.sh script in the repo.

Are you proposing also changing that script?  Or perhaps you do not get an error as I am?

antoninbas:
That's fine I think. I didn't get an update on the membership agreement and it's probably good to fix this.

PR #619, opened 2018-06-13
Implement PI multicast functions

antoninbas:
This enables P4Runtime multicast support in simple_switch_grpc. We
include a GTest unit test which verifies dataplane behavior after
programming the bmv2 PRE through P4Runtime.


antoninbas:
Depends on https://github.com/p4lang/PI/pull/377

PR #624, opened 2018-06-27
Change simple_switch_grpc priority zero test

antoninbas:
As per the P4Runtime spec, zero is not a valid priority any more for
ternary matches.


antoninbas:
Depends on https://github.com/p4lang/PI/pull/385

Issue #609, opened 2018-05-29
Counter index validation is missing from method count


avinash23482:
Hi,
I am trying counters in a P4 program.
Following is a snippet from the program,       

```
control Ingress(inout Headers hdr, inout MetaData meta,     
 inout standard_metadata_t smeta) {
    
    counter(1024,CounterType.packets) c1;

    action forward( MACAddr_t dstAddr, Port_t port){
        hdr.ethernet.dstAddr = dstAddr;
        smeta.egress_port = port;
        c1.count(2000);
    }
}
```

The index which is getting increment is greater than the maximum supported size for the counter c1.
Still the JSON output is getting generated without any validation.

jafingerhut:
Did you run this compiled P4 program in simple_switch?

If so, was there was no error or warning message about the out-of-bounds index when you used the `--log-console` or `--log-file` options?

A run-time check and warning in the log seems reasonable to me for out-of-bounds indices for counters, meters, and registers, but going to the trouble of checking statically when loading the JSON file for constant indices that are out of range seems like something that would seldom find a problem.

jafingerhut:
I tested myself and found that the behavioral model is not currently doing any run-time bounds checking when updating indexed counters, meters, and register arrays.

I created a proposed PR that will cause useful messages to be printed to the log (when you use the `--log-console` or `--log-file` command line options to simple_switch) to help you detect if your program is accessing out of bounds indices.

avinash23482:
Andy,I just tried generating JSON output for bmv2. I didn't run the compiled program.

-Avinash

jafingerhut:
@avinash23482 Now that PR #637 has been merged in, if you update to the latest version of source for p4lang/behavioral-model, compile, and install it again, if you run the JSON output with simple_switch and `--log-console` or `--log-file` enabled, you will get an error message in the log when c1.count(2000) is executed.  As far as the P4 program behavior goes, such out of bounds accesses will be treated as a no op.

At least in my opinion, it seems like too much work for too little gain to try to catch such things at compile time.

antoninbas:
I agree with @jafingerhut that that kind of static verification seems like an overkill to do in bmv2 or in the compiler. It would be appropriate in a static code analyzer tool - which would need to be architecture-aware to analyze extern method calls.

avinash23482:
Thanks Antonin and Andy !!!

PR #637, opened 2018-08-11
Add bounds checking for indexed counters, meters, and register arrays

jafingerhut:
with error message printed to log if --log-console or --log-file are
enabled.

Also log messages with possibly useful details on correct accesses to
these stateful objects, for use by P4 programmers when debugging their
code.

### DIFF
--- a/targets/simple_switch/primitives.cpp
+++ b/targets/simple_switch/primitives.cpp
@@ -25,6 +25,7 @@
 #include <bm/bm_sim/meters.h>
 #include <bm/bm_sim/packet.h>
 #include <bm/bm_sim/phv.h>
+#include <bm/bm_sim/logger.h>
 
 #include <random>
 #include <thread>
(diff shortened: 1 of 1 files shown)


targets/simple_switch/primitives.cpp
```diff
@@ -282,15 +283,42 @@ REGISTER_PRIMITIVE(no_op);
 class execute_meter
   : public ActionPrimitive<MeterArray &, const Data &, Field &> {
   void operator ()(MeterArray &meter_array, const Data &idx, Field &dst) {
-    dst.set(meter_array.execute_meter(get_packet(), idx.get_uint()));
+    unsigned int i = idx.get_uint();
```
antoninbas: `auto`

targets/simple_switch/primitives.cpp
```diff
@@ -282,15 +283,42 @@ REGISTER_PRIMITIVE(no_op);
...
+    if (i < meter_array.size()) {
+        unsigned int color = meter_array.execute_meter(get_packet(), i);
```
antoninbas: `auto`

targets/simple_switch/primitives.cpp
```diff
@@ -282,15 +283,42 @@ REGISTER_PRIMITIVE(no_op);
...
+        dst.set(color);
+        BMLOG_TRACE_PKT(get_packet(),
+                        "Updated meter {} at index {},"
```
antoninbas: Pleas add single quotes around names (`'{}'`)

antoninbas:
Fixes #609

jafingerhut:
When you say "This should probably go inside bm_sim so that all targets can benefit from it", is there a particular file or files you have in mind, that by changing existing methods inside of them will have a similar effect?

Or do you mean that the functions in the file targets/simple_switch/primitives.cpp should be moved to a file inside bm_sim?

Happy to make changes like that if I know what you are looking for.

targets/simple_switch/primitives.cpp
jafingerhut: Done

targets/simple_switch/primitives.cpp
jafingerhut: Done

targets/simple_switch/primitives.cpp
jafingerhut: Done

jafingerhut:
From compiling simple_switch after `./configure 'CXXFLAGS=-O0 -g'`, it gives no error or assert messages when attempting out-of-bounds updates of indexed counters or meters.  It does fail with an assert when attempting to read a register array at an out of bounds index.

I may be drawing the wrong conclusion here, but this seems to suggest that the current code (without this PR's changes) is silently corrupting out-of-bounds memory for such wrong counter and meter updates?  If so, adding a run-time if check would make that check the only one, for counters and meters at least.

jafingerhut:
I have added the requested `#ifndef NDEBUG` around the new out-of-bounds index condition checks and error logging.  Let me know if you see anything else you'd like changed.

PR #638, opened 2018-08-14
Add similar bounds checking to psa_switch counter/etc accesses ...

jafingerhut:
... as was recently added for simple_switch.


jafingerhut:
I checked all other source files for l2_switch, simple_router, etc. and they do not have code like this.  psa_switch's primitives.cpp appears to have been directly copied from simple_switch's as a start, and has diverged only in comments from simple_switch's version so far.

jafingerhut:
Some docker command appears to have failed in the Travis tests, likely unrelated to the source code changes in this PR.  I don't see a button I can click to attempt a repeat of those tests, else I'd click it.

PR #657, opened 2018-09-18
Fix for ubuntu18.04

jafingerhut:


jafingerhut:
This is similar to https://github.com/p4lang/behavioral-model/pull/617 but the shell script change has been tested so that it should continue to work for Ubuntu 18.10, 19.04, meaning that the string comparison is not testing for Ubuntu version exactly 18.04, but 18.x or later.  Of course, it may need further tweaking once those Ubuntu releases actually come out, to see what packages they need.

PR #660, opened 2018-09-21
Add checksum_error field and behavior to simple_switch docs

jafingerhut:
I have verified with a small test program that it is behaving as
documented here.


jafingerhut:
@jnfoster @antoninbas I think this is correct from a quick test, but please see if it matches what you expect here.  I will create a separate PR to remove the 'deprecated' comment about checksum_error field in p4lang/p4c/p4include/v1model.p4

PR #665, opened 2018-09-28
Periodically-called externs

isaac-ped:
This adds functionality to bm_sim and the simple_switch target to register an function to be called periodically with a given interval.

A function declared outside of a primitive will always be called in that interval:

```cpp
static void run_always() {
    BMLOG_DEBUG("This function will run once a second");
}

REGISTER_PERIODIC_ACTION(run_always, std::chrono::seconds(1));
```

A function declared inside of a primitive will only be called if that primitive is used as an extern somewhere in the compiled P4 file:

```cpp
class timeout_extern : public ActionPrimitive<> {

    void operator() () {
        BMLOG_DEBUG("Calling timeout_extern");
    }


    static void periodic() {
        BMLOG_DEBUG("This will be called 4 times a second if timeout_extern() is used in P4");
    }

    REGISTER_PERIODIC_ACTION(periodic, std::chrono::milliseconds(250));
};
```


include/bm/bm_sim/actions.h
```diff
@@ -96,6 +97,42 @@ class NamedCalculation;
...
+                         std::chrono::milliseconds interval);
+
+  // To be called continuously : waits for the next interval to
+  // elapse then calls the next registered function or set of functions.
+  // Returns `false` if no actions are registered
+  bool next();
+
+ private:
+  unsigned int tick;
+  std::vector<PeriodicAction> list_{};
```
antoninbas: Wouldn't a priority queue be more appropriate for this? The actions can be ordered based on the next execution timestamp (last + period) and you can sleep until the next action (first element of the queue) is ready to be executed. You wouldn't need a min interval and you wouldn't need to iterate through the whole list of actions. I think you would get O(logN) complexity every time an action is executed.

include/bm/bm_sim/actions.h
```diff
@@ -96,6 +97,42 @@ class NamedCalculation;
...
+// periodically at a given interval
+struct PeriodicAction {
+  const char *name;
+  std::function<void()> fn;
+  std::chrono::milliseconds interval_ms;
+  unsigned int interval_ticks;
+};
+
+// Contains the list of the periodically executed actions
+class PeriodicActionsList {
```
antoninbas: This class is not thread safe. Have you given some thought to what happens in case of P4 program swap?

targets/simple_switch/simple_switch.cpp
```diff
@@ -188,6 +188,7 @@ SimpleSwitch::start_and_return_() {
     threads_.push_back(std::thread(&SimpleSwitch::egress_thread, this, i));
   }
   threads_.push_back(std::thread(&SimpleSwitch::transmit_thread, this));
+  threads_.push_back(std::thread(&SimpleSwitch::periodic_thread, this));
```
antoninbas: I'm not sure why we need any change in simple_switch at all. Shouldn't the periodic thread be managed entirely by the `SwitchWContexts` base class?

targets/simple_switch/simple_switch.cpp
```diff
@@ -267,6 +269,16 @@ SimpleSwitch::get_time_since_epoch_us() const {
   return duration_cast<ts_res>(tp.time_since_epoch()).count();
 }
 
+void
+SimpleSwitch::periodic_thread() {
+  auto actions = bm::PeriodicActionsList::get_instance();
+  while (!exiting) {
+    if (actions->next() != true) {
```
antoninbas: It seems that if there is no periodic action registered, the thread exits right away. But what if I load a new P4 program that uses an extern with a periodic action?

include/bm/bm_sim/actions.h
isaac-ped: Absolutely true. Will update to use priority queue shortly.

include/bm/bm_sim/actions.h
isaac-ped: I'm unsure how the program swap occurs, and why it would cause concurrent access of this structure. Could you elaborate a bit on what causes this class to be less safe than ActionsOpcodesMap, or point me to documentation on the program swap procedure?

targets/simple_switch/simple_switch.cpp
isaac-ped: Documentation for in `SwitchWContexts` states that the `start_and_return()` function of an individual target should start all processing threads. Is it acceptable to do processing in `SwitchWContexts` as well?

include/bm/bm_sim/actions.h
antoninbas: Swapping the P4 program is done by loading a new JSON file (generated by the compiler). Unfortunately I do not believe there is any centralized documentation fore this, only comments scattered in the switch.h source code. It is very similar to the P4Runtime ForwardingPipelineConfig swap if you're familiar with P4Runtime. It is a 2 step operation:
* first, `load_new_config` is called (https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/switch.h#L789) to load a new JSON (sent through RPC). All packet processing still uses the old JSON, but all subsequent runtime operations (e.g. table entry adds) are done on the new JSON
* when the application is ready, it calls `swap_configs` (https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/switch.h#L792) which lets bmv2 know that the new JSON & associated runtime state are ready to be used for packet processing
There is actually a 3rd step. `swap_configs` only signals to bmv2 that configurations should be swapped. But the target (e.g. simple_switch) has final say and controls exactly when the swap occurs. This is because the target may have to update some internal state, e.g. pointers that may be invalidated by the swap. This is detailed here: https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/switch.h#L49

Now back to your PR. `REGISTER_PERIODIC_ACTION` is actually executed when the first instance of a given primitive (extern function) is encountered while loading the JSON configuration. You have a single global instance of `PeriodicActionsList` with a thread periodically executing `next()`. What if `next()` is called while the new config is loaded and `REGISTER_PERIODIC_ACTION`? So that's why you need to make `PeriodicActionsList` thread safe.

When a new configuration is enabled, you need to cancel all the tasks for the previous P4 and start tasks for the new P4 IMO (the 2 programs may not be using the same extern functions). The only way to make that work is to tie the lifetime of `PeriodicActionsList` with that of `P4Objects`, by letting the `P4Objects` instance own the `PeriodicActionsList` instance just like it owns the `LearnEngine` instance. When configs are swapped, the `PeriodicActionsList` instances will be swapped as well. Now there is one issue with this approach: what about the periodic tasks registered outside of primitive / extern objects?

That leads us to what I think is the right approach. Let a `PeriodicActionsList` instance manage its own thread (like `LearnEngine`). Keep using the singleton pattern; even though it would make more sense to give ownership to `SwitchWContexts` IMO, it may be unpractical to achieve. But modify the `REGISTER_PERIODIC_ACTION` mechanics. Maybe it could look a lot like starting a thread actually.
```
// should not be copyable
// this is merely a skeleton
struct PeriodicTask {
    PeriodicTask(...) { PeriodicActionsList::get_instance()->register(...); }
    ~PeriodicTask() { cancel(); }
    void cancel() { PeriodicActionsList::get_instance()->unregister(...); }
};

class timeout_extern : public ActionPrimitive<> {
    timeout_extern() {
        // the task will be unregistered when the object is destroyed
        task = PeriodicTask(periodic, interval, this);
     }

    void operator() () {
        BMLOG_DEBUG("Calling timeout_extern");
    }

    // it is not a good idea to make this static, as it is not a good idea to have any shared static state in primitive instances
    // static void periodic() {
    void periodic() {
        BMLOG_DEBUG("This will be called 4 times a second if timeout_extern() is used in P4");
    }

    PeriodicTask task;
};
```
For a global static task, you could do `new PeriodicTask(...)` and never destroy it, or you could give ownership to the target class (e.g. `SimpleSwitch`).

This is still more convenient and efficient than having to manage a thread yourself for every periodic task you want to execute. You will also not have to worry about P4 program swap, as its `P4Objects` instance should have its own instance of each primitive in use.

targets/simple_switch/simple_switch.cpp
antoninbas: There are many threads that are already managed by `SwitchWContexts`. All targets can benefit from this kind of periodic task registration and you can hide thread management from the target (it's already done for learning for example).

include/bm/bm_sim/actions.h
antoninbas: If you do agree to go this route, I believe you should put the new classes in their own source files, not in actions.h / actions.cpp

isaac-ped:
Thank you for all of the detailed comments. I believe I have fixed the thread-safety problems, and the periodic tasks should be registered automatically as the externs they are tied to are loaded and unloaded by new configs.

Similar to before, 

```cpp
void tick() {
    std::cout << "Runs once a second" << std::endl;
}

REGISTER_GLOBAL_PERIODIC(tick, std::chrono::seconds(1));

class test_extern : public ActionPrimitive<>{
    void periodic() {
        std::cout << "Runs 4 times a second, if test_extern() is used in P4" << std::endl;
    }

    REGISTER_EXTERN_PERIODIC(test_extern, periodic, std::chrono::milliseconds(250));
}
```

I have also changed the underlying structure to be a priority_queue as you suggested.

I am still having trouble, however, figuring out where in SwitchWContexts to start the thread. All that needs to be added is a `PeriodicTaskList::get_instance()->start()` somewhere.
I could also make the PeriodicTaskList start its thread automatically upon creation.

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,92 @@
...
+  std::chrono::system_clock::time_point next;
+
+  PeriodicTask(const std::string &name,
+               std::function<void()> fn,
+               std::chrono::milliseconds interval);
+
+  ~PeriodicTask();
+};
+
+// This smells a bit (binding 'this' in inline member initialization),
```
antoninbas: This exactly why I suggested doing the registration in the constructor, I'm not sure why you are insisting on keeping macros.

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,92 @@
...
+// but it compiles and I haven't found documentation against it
+#define REGISTER_EXTERN_PERIODIC(extrn_name, fn, interval) \
```
antoninbas: Even though `extern` is a reserved work in C, is there a problem with using `extern_name`?

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,92 @@
...
+  PeriodicTask fn## __task = \
+      PeriodicTask(#extrn_name #fn, \
+                   std::bind(&extrn_name::fn, this), \
+                   interval);
+
+#define REGISTER_GLOBAL_PERIODIC(fn, interval) \
+  static PeriodicTask fn## __task = PeriodicTask(#fn, fn, interval);
```
antoninbas: This is not great (and again why I suggested not using these macros). Even though you may be able to guarantee that the `PeriodicTaskList` instance will be destroyed after all the `PeriodicTask` objects (because you may be able to guarantee that the `PeriodicTaskList` instance is constructed first), can't you just avoid static global variables?

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,92 @@
...
+
+class PeriodicTaskList {
+  std::mutex queue_mutex;
```
antoninbas: can you order you data members and function members according to the Google style guide?

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,92 @@
...
+    }
+  };
+
+  using TaskQueue = std::priority_queue<PeriodicTask*,
+                                        std::vector<PeriodicTask*>,
+                                        PeriodCompare>;
+
+  TaskQueue task_queue;
+
+  std::atomic<bool> running;
```
antoninbas: Not sure why you need 2 variables (`running` and `exiting`)
Also since you have a mutex anyway, I don't think there is a big advantage in making them atomic rather than just using the mutex.

include/bm/bm_sim/periodic_task.h
isaac-ped: Did not realize that you were suggesting removing the macros altogether. Will remove macros in favor of the registration method you suggested above.

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,170 @@
...
+  reset_next();
+}
+
+constexpr std::chrono::milliseconds PeriodicTaskList::kDefaultTimeout;
+
+bool
+PeriodicTaskList::contains_task(PeriodicTask *task) {
+  std::lock_guard<std::mutex> lock(queue_mutex);
+
+  TaskQueue tmp;
```
antoninbas: This is why it probably makes sense to differentiate between a *registered task* and the *next execution of a task*
You could create a new `struct` for *next execution of task* and a new data structure mapping task pointer to *next execution of task* pointer. This would let you check in constant time if a task exists and would also let you do constant-time "lazy" removal of a task.
However, performance is not really a concern for task registration & removal so I won't block the PR for this.

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,170 @@
...
+    BMLOG_DEBUG("Warning: Task {} does not exist in periodic task queue",
+                task->name);
+    return false;
+  }
+  return true;
+}
+
+void
+PeriodicTaskList::start() {
+  if (running) {
```
antoninbas: needs to be protected by locking the mutex

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,170 @@
...
+  while (true) {
+    std::unique_lock<std::mutex> lk(queue_mutex);
+    if (!running) {
+      break;
+    }
+    std::chrono::system_clock::time_point next;
+    if (!task_queue.empty()) {
+      next = task_queue.top()->next;
+    } else {
+      next = std::chrono::system_clock::now() + kDefaultTimeout;
```
antoninbas: sorry if I'm missing something but why do we need a default timeout, given that we use a condition variable?

src/bm_sim/switch.cpp
```diff
@@ -72,6 +73,8 @@ SwitchWContexts::receive(port_t port_num, const char *buffer, int len) {
 
 void
 SwitchWContexts::start_and_return() {
+  // Starts any registered periodically-executing externs
+  PeriodicTaskList::get_instance().start();
```
antoninbas: Ir would probably beet to put this with the `start` call underneath, unless you have a good reason to start execution before a P4 program is loaded?

tests/test_periodic_tasks.cpp
```diff
@@ -0,0 +1,108 @@
...
+static constexpr std::chrono::milliseconds kInterval{kSleepTimeMs / kPeriods};
+static constexpr std::chrono::milliseconds kSleepTime{kSleepTimeMs};
+
+// Tasks registered before the start of the thread execute on start
+TEST(PeriodicExtern, PreStartRegistration) {
+  int i = 0;
+  PeriodicIncrementor incrementor(kInterval, &i);
+  std::this_thread::sleep_for(kSleepTime);
+
+  ASSERT_EQ(i, 0);
```
antoninbas: I didn't know that when I started writing GTests, but unless a violation of the expectation causes the rest to crash, I believe `EXPECT_*` macros are preferred.

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,120 @@
...
+//!   }
+//!
+//!   void periodic_fn() {
+//!     // This will execute once a second
+//!   }
+//!
+//!   PeriodicTask task;
+//! }
+//! @endcode
+class PeriodicTask {
```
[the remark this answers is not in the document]
isaac-ped: Makes sense. I will remove the constructors.

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,120 @@
...
+#include <chrono>
+#include <functional>
+#include <condition_variable>
+#include <mutex>
+
+namespace bm {
+
+//! Initializes and registers a task that is to be executed periodically
+//! at a fixed interval. Task will execute during the lifetime of the object.
+//! Make a member of an ActionPrimitive class to tie the execution to the
```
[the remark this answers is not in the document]
isaac-ped: Was trying to explain how making an PeriodicTask a member of an ActionPrimitive object will ensure that the task is only executed if the extern represented by the ActionPrimitive is used in P4. Will clarify in the comment.

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+bool
+PeriodicTaskList::register_task(PeriodicTask *task) {
+  if (contains_task(task)) {
+    BMLOG_DEBUG("Warning: Task {} already exists in periodic task queue",
+                task->name);
+    return false;
+  }
+
+  std::lock_guard<std::mutex> lock(queue_mutex);
```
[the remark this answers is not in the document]
isaac-ped: Right you are! Missed that. Will fix.

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,120 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
```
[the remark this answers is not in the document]
isaac-ped: I am waiting to hear back about using my organizations name in the copyright. Will you need a CLA to accept the changes, and will it be worded as in: https://p4.org/assets/P4_Language_Consortium_Membership_Agreement.pdf ?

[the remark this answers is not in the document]
antoninbas: If your organization is not a p4.org member, you are making this contribution as an individual and you should use your name in the copyright.
If your organization is not a member(which is usually the most convenient way to contribute) and not interested in becoming one, you can sign the CLA as an individual. I think you can fill / sign the PDF and send it to membership@p4.org.

isaac-ped:
My apologies for the long delay.
I am still waiting for a decision regarding the copyright. For the moment, I have fixed the other issues.

I understand your comments regarding the different data structures for a registered task vs. task execution. Because (as you said) task registration performance is not a concern, I did not make that update at this time.

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,127 @@
...
+//!
+//!   // Note: If `task` is not the last declared member, periodic_fn
+//!   // may be called before all members have been initialized
+//!   PeriodicTask task;
+//! }
+//! @endcode
+class PeriodicTask {
+ public:
+  // Friend so that PeriodicTaskList can call execute()
+  friend class PeriodicTaskList;
```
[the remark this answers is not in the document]
antoninbas: Not an issue

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,127 @@
...
+
+  PeriodicTask(const std::string &name,
+               std::function<void()> fn,
+               std::chrono::milliseconds interval);
+  ~PeriodicTask();
+
+  // Deleting copy-constructor and copy-assignment
+  PeriodicTask(const PeriodicTask&) = delete;
+  PeriodicTask& operator= (const PeriodicTask&) = delete;
```
antoninbas: you actually must make your class non-movable as well

include/bm/bm_sim/periodic_task.h
```diff
@@ -0,0 +1,127 @@
...
+
+  void loop();
+
+  // The queue of PeriodicTasks, ordered by next execution time
+  TaskQueue task_queue;
+
+  std::thread periodic_thread;
+  bool running;
+  std::mutex queue_mutex;
+  std::condition_variable cv;
```
antoninbas: I recommend making `mutex` and `cv` mutable for future-proofness

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,165 @@
...
+      contains = true;
+    }
+    tmp.push(task_queue.top());
+    task_queue.pop();
+  }
+  task_queue.swap(tmp);
+
+  // Do not add a task twice
+  if (contains) {
+    BMLOG_DEBUG("Warning: Task {} already exists in periodic task queue",
```
antoninbas: this deserves to be a warning IMO

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,165 @@
...
+    } else {
+      tmp.push(task_queue.top());
+    }
+    task_queue.pop();
+  }
+  task_queue.swap(tmp);
+
+  cv.notify_all();
+  if (!removed) {
+    BMLOG_DEBUG("Warning: Task {} does not exist in periodic task queue",
```
antoninbas: this deserves to be a warning as well

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,165 @@
...
+    return false;
+  }
+  return true;
+}
+
+void
+PeriodicTaskList::start() {
+  std::lock_guard<std::mutex> lock(queue_mutex);
+  if (running) {
+    BMLOG_DEBUG("Warning: Tried to start already-running periodic task loop");
```
antoninbas: make it a warning and remove "Warning" from the message string...

src/bm_sim/periodic_task.cpp
```diff
@@ -0,0 +1,165 @@
...
+  running = true;
+  periodic_thread = std::thread(&PeriodicTaskList::loop, this);
+}
+
+void
+PeriodicTaskList::join() {
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex);
+    if (!running) {
+      BMLOG_DEBUG("Warning: Tried to join non-running periodic thread");
```
antoninbas: make it a warning and remove "Warning" from the message string...

isaac-ped:
Sorry for all the delays. Still waiting to hear from higher-ups about copyright.

include/bm/bm_sim/periodic_task.h
isaac-ped: I have taken your suggestion, but am not sure what exactly you mean by future-proofness.
Do you mind clarifying a little or providing a link on the topic, if only for my edification? I have read up on the interactions between const functions and mutable members, but am unsure how it applies.
By this, do you mean that the public functions of the PeriodicTaskList should have been `const`, and as such the mutex and cv should be mutable? Or is it simply because of the guarantee that `mutex` and `cv` are already thread safe, and thus do not need the protection that `const` offers?

isaac-ped:
Still waiting to hear about copyright (I had never expected the answer to take this long). Will (hopefully) hear back in the next couple of days. Thanks for your patience, and for all of your help and explanation!

isaac-ped:
I have updated the copyright messages and merged the current master branch into my fork. Do let me know if any other updates are needed.

isaac-ped:
I am so sorry for the delay. The membership agreement had not been executed as of my last message, but has now been completed.

PR #666, opened 2018-10-02
Packet & headers serialization functions

isaac-ped:
This adds functionality to externs which require acting on the packet in a fully serialized form, including headers.

Three functions are added to the packet/phv interfaces:

```cpp
std::vector<Header *> headers;
phv->get_valid_headers(&headers);
```
Places in the `headers` vector all of the headers of the headers of the given packet which are valid, and will be deparsed at the end of the pipeline.
Note: Each header listed in the headers iterator in the phv appears to be listed twice -- one with the `hdr_` prefix, and one without. This ignores those headers prefixed with `hdr_` so as not to duplicate.

```cpp
size_t buff_size;
char *buff = packet.deparsed_data(hdrs, &buff_size);
// buff now contains the serialized packet prefixed by the headers in hdrs
// hdrs can be filled by get_valid_headers() or from headers manually
// passed to an extern
... // Use/modify serialized headers and packet
packet.reparse_headers(buff, hdrs);
```

### DIFF
--- a/src/bm_sim/packet.cpp
+++ b/src/bm_sim/packet.cpp
@@ -23,6 +23,7 @@
 
 #include <algorithm>  // for swap
 #include <atomic>
+#include <vector>
 
 #include "xxhash.h"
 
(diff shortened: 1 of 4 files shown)


src/bm_sim/packet.cpp
```diff
@@ -119,6 +120,37 @@ Packet::~Packet() {
...
 }
 
+char *
+Packet::deparsed_data(const std::vector<Header *>&hdrs, size_t *size_p) {
+  size_t payload_size = get_data_size();
+  size_t size = payload_size;
+  for (auto hdr : hdrs) {
+     size += hdr->get_nbytes_packet();
+  }
+  char *buff = new char[size];
```
antoninbas: in C++11, you want to use a smart pointer or even a vector

isaac-ped:
Certainly reasonable complaints. For our specific use it was most convenient to have a deparsed copy of the valid headers. However, a copy of the original packet as it was received before control execution would work as well.
If accessing either the original packet or the current deparser is feasible, then this request is certainly not necessary. Given that, I will close this PR and ask for further advice regarding packet serialization on the mailing list.

PR #668, opened 2018-10-08
p4dbg.py: don't use thrift if socket and json are specified

theojepsen:
This lets you use `p4dbg.py` with `simple_switch_grpc` without thrift. You must specify both the `--json` and `--socket` path.

### DIFF
--- a/tools/p4dbg.py
+++ b/tools/p4dbg.py
@@ -562,7 +562,7 @@ class DebuggerAPI(cmd.Cmd):
     prompt = 'P4DBG: '
     intro = "Prototype for Debugger UI"
 
-    def __init__(self, addr, standard_client=None):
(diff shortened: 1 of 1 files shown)


theojepsen:
@antoninbas I've made the changes you proposed in your review.

tools/p4dbg.py
```diff
@@ -676,6 +678,8 @@ def handle_config_change(self):
         print "or we will exit."
         v = prompt_yes_no("Do you want to request the new config?", True)
         if v:
+            if not self.standard_client:
```
antoninbas: There is no sense in even asking the user, since you won't be able to load the new config. You should just print something and exit with no prompt.

theojepsen: Good point. Updated. It now prints that it cannot connect via thrift to reload the config, and exits.

PR #679, opened 2018-10-31
Add support for sending digests through P4Runtime

antoninbas:
For the simple_switch_grpc target. This is done by registering a custom
TransportIface implementation with the Switch base class, to capture all
the notifications which are usually broadcast on a Nanomsg PUBSUB
socket.

The P4Runtime implementation in p4lang/PI requires that no digest
notification be sent while a forwarding pipeline config swap is
ongoing. To ensure that we respect this assumption, we now generate a
new kind of notifications for "swap" events. These notifications are
captured by our custom TransportIface implementation for
simple_switch_grpc, to ensure that all digest notifications are ignored
during the swap.

### DIFF
--- a/PI/src/pi_learn_imp.cpp
+++ b/PI/src/pi_learn_imp.cpp
@@ -19,25 +19,85 @@
  */
 
 #include <bm/bm_sim/_assert.h>
+#include <bm/bm_sim/learning.h>
+#include <bm/bm_sim/switch.h>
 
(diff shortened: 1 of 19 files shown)


jafingerhut:
👍

Note: this is a comment on the feature being added.   I haven't read and reviewed the code, and I'm unlikely to find problems with it that you wouldn't from code review :-)

PR #688, opened 2018-11-23
Add notes on the different targets

jafingerhut:


jafingerhut:
If you have any recommendations for how to describe l2_switch and simple_router targets here, I am open to suggestions.  I do not know enough about how they differ from simple_switch and/or their goals to accurately describe them.

jafingerhut:
@antoninbas - I added your description of l2_switch and simple_router, with only minor editorial changes.

PR #699, opened 2019-01-14
Match kind docs

jafingerhut:


jafingerhut:
The part of this I think is most likely in need of editing is that I said that it is the maximum numeric priority table entry that wins if there are multiple matching entries in a range or ternary table.

Perhaps it would be better to say what it is in the P4Runtime specification, and that whether the winner is determined by the minimum or the maximum numeric priority may be different for other control plane APIs, and one should check their documentation?

jafingerhut:
Pretty weird that coverage went down so much for a change in a single documentation file :-)

jafingerhut:
Added a note about maximum numeric priority for P4Runtime API, and check your control plane API's documentation if you use something other than that, since some reverse the convention so that minimum numeric priority is the winner.  It also gives Codecov another chance to do something sane.

jafingerhut:
I could create a new document in the docs directory and add these details there if you think it would be better for stuff like this that is common to all BMv2 executables.

antoninbas:
I think it's fine for now. We can worry about re-organizing some things when psa_switch is in a better shape. Maybe some of that information could even go into the PSA spec? Not the bmv2 implementation-specific stuff, but how match kinds within a table "interact" with each other. What do you think?

jafingerhut:
Given that PSA defines a `range` match_kind, and the P4_16 language spec doesn't explicitly define what `lpm` match behavior is, it does seem reasonable for PSA to define this kind of thing.  I will create an issue on the PSA spec for that.

PR #701, opened 2019-01-17
Add some documentation specifically for P4_16 plus the v1model archit…

jafingerhut:
…ecture


jafingerhut:
@hesingh See if the additions proposed in this PR are the kind of documentation you wish there was more of for the P4_16 plus v1model architecture.

PR #705, opened 2019-01-22
Add support for P4Runtime idle timeout to simple_switch_grpc

antoninbas:
This enables idle timeout support by implementing the new PI target
functions and calling pi_table_idle_timeout_notify appropriately when a
match entry expires.

A few things to take into consideration:
 * _pi_table_idle_timeout_config_set is currently not doing anything. In
   bmv2 the sweep interval is global and not per table. We could set the
   sweep interval to the minimum of all the values provided by
   _pi_table_idle_timeout_config_set, but that would require storing
   some state in the PI implementation, and that did not seem worth the
   trouble at this stage.
 * bmv2 generates batched notifications for idle timeout (one per table
   per sweep interval). It seems a bit of a shame that we do not support
   batching in PI, therefore generating more function calls than we
   actually need.
 * the P4Runtime specification states that after generating a
   notification, the TTL should be reset to its initial value and a new
   notification should be generated if the match entry expires
   again. Currently, this is not exactly what bmv2 is doing. The bmv2
   "retimer" is based on the sweep interval and is not per entry. The
   sweeper thread will not generate notifications for the same entry in
   2 successive sweeps (by default the sweeper thread executes every
   1s), but if the entry has still not been hit or removed by the 3rd
   sweep, a new notification will be sent. We should be able to change
   this behavior easily in the future.

### DIFF
--- a/PI/bm/PI/pi.h
+++ b/PI/bm/PI/pi.h
@@ -21,6 +21,8 @@
 #ifndef BM_PI_PI_H_
 #define BM_PI_PI_H_
 
+#include <PI/pi.h>
+
 #include <cstdint>
 
 namespace bm {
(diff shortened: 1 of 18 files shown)


antoninbas:
Depends on https://github.com/p4lang/PI/pull/438

PR #710, opened 2019-02-03
Fix bashisms in configure

ayekat:
POSIX does not mandate support for the `==` syntax for the equality check in `test`. Generating `configure` (which has a `#!/bin/sh` shebang) with lines like `test foo == bar` breaks in all environments where `test` does not support the `==` syntax (e.g. the version provided by dash).

This leads to `configure` output like this:

    ./configure: 18025: test: xyes: unexpected operator

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -226,18 +226,18 @@ AS_IF([test "x$enable_modules" != "xno"], [
       modules_enabled=yes
     ], [
       AC_MSG_RESULT(no)
-      AS_IF([test "x$enable_modules" == "xyes"], [
+      AS_IF([test "x$enable_modules" = "xyes"], [
         AC_MSG_ERROR([Cannot enable modules without dlopen])
       ])
     ])
   ],[
     AC_MSG_RESULT(no)
-    AS_IF([test "x$enable_modules" == "xyes"], [
+    AS_IF([test "x$enable_modules" = "xyes"], [
       AC_MSG_ERROR([Cannot enable modules without dlfcn.h])
     ])
   ])
 ])
-AM_CONDITIONAL([ENABLE_MODULES], [test "x$modules_enabled" == "xyes"])
+AM_CONDITIONAL([ENABLE_MODULES], [test "x$modules_enabled" = "xyes"])
 
 # C++ libraries are harder (http://nerdland.net/2009/07/detecting-c-libraries-with-autotools/),
 # so use headers to check


antoninbas:
Hi,
Thanks for your patch. Could you confirm that you or your organization (company / university) are a member of p4.org? This is required in order for me to accept your patch (https://github.com/p4lang/behavioral-model/blob/master/CONTRIBUTING.md).

ayekat:
Thank you for your reply. I'm a bit irritated, though.

As far as *I* am concerend, this bug is fixed. I simply consider it good manners to propagate bugfixes for open source software upstream—in this case to you. So it feels a bit backwards that I am the one who has to jump through additional hoops; *you* are the ones profiting from external contributions (or not).

As someone who has merely built this software for trying it out, I'm not particularly concerned about whether you incorporate these changes, and even less motivated to go through all that paperwork. Feel free to reject this simple—and noteworthily trivial—bugfix.

jafingerhut:
I understand that such hoops can be annoying to jump through, but please also realize that when running an open source project, _without_ the proper legal copyright assignments in place for all contributions, the project's intellectual property cannot be changed to a different license later, and/or could be open to certain kinds of legal challenges in the future, that it would not if it requires these annoying hoops to be jumped through.

We do appreciate bug fixes and contributions, but if the result is a project that falls apart on legal grounds, the effort of previous contributors will all be for nought.

jafingerhut:
Note that _many_ major open source projects have very similar hoops to jump through, e.g. if you want to contribute fixes to Python, Linux, Clojure, any Apache foundation project, Mozilla, etc. they all have some kind of assignment or co-ownership of your contribution's copyright that must be agreed to first.

If you are a one-time small contributor, sometimes it is easier for an existing project maintainer to re-implement the fix, of course.

ayekat:
Thanks for the quick replies!

I have always assumed that for code contributions it is implicitly assumed that the involved parties agree with the terms of the license (in this particular case [Section 5 of the Apache License](https://github.com/p4lang/behavioral-model/blob/master/LICENSE#L130-L136))—or at least that was the case for all projects I have contributed to so far (which admittedly doesn't involve any of the ones you have mentioned).

But I'm not a lawyer; it does indeed appear reasonable to be extra safe about these kinds of things.

> If you are a one-time small contributor, sometimes it is easier for an existing project maintainer to re-implement the fix, of course.

That would be the most pragmatic solution, I think (to be explicit, I hereby renounce any copyright claims from my side on the attached change).

antoninbas:
Thanks @jafingerhut 
To be fair, I believe a lot of the projects you mention use a clock-through CLA that contributors can go through the first time they open a PR. This is definitely something I believe we should investigate for p4.org repositories.

jafingerhut:
@ayekat One reason for copyright assignment is that for some projects, if the project maintainers want to keep the option open to _change_ the license at some point in the future, they might not legally be able to do that if they do not have the agreement of all contributors, _or_ if they have at least joint copyright ownership in all of the code at the time they make such a license change.  There are major projects that have wanted to change their license, and had difficulty even contacting all of the people who had contributed code, let alone getting their agreement to the change of license.

PR #711, opened 2019-02-04
Store copy of user callback in PortMonitor

antoninbas:
Instead of storing a reference. This is much more convenient IMO because
of the lack of cookie in the CB registration, which makes it more likely
that the user will use something "more" than just a plain function
pointer. When using std::bind for example, it is convenient to let the
PortMonitor store the bind result. There is no drawback AFAIK. I also
modified the signature of the CB registration functions all the way to
DevMgr to take a value instead of a const reference (to make it more
obvious that we make a copy and that the function object must be
copyable).


antoninbas:
I changed my mind on this: I don't like storing the reference much but at least it gives us a simple way to deregister the callback if needed by providing the same reference...

PR #714, opened 2019-02-09
Fix #706 - unicast packets should begin egress with instance_type NORMAL

### LINKED ISSUE #706: What should instance_type be for a packet resubmitted on ingress, then sent to egress?
I have a test program I am trying on BMv2 simple_switch, where the packet is recirculated on ingress, so after that it starts ingress with instance_type=6 (PKT_INSTANCE_TYPE_RESUBMIT) for a resubmitted packet.  That is what I expected, so far.

On that next pass in ingress, the packet is sent to egress (i.e. no resubmit operation this time, no multicast, no clone).

If the packet had not been resubmitted, it would start egress processing with instance_type=0 (PKT_INSTANCE_TYPE_NORMAL).

But since it started ingress with instance_type=6, and didn't go through any special cases, the instance_type=6 was preserved through the packet buffer until the packet started egress processing.

I would have expected it to be 0 (NORMAL) at start of egress.

What would you expect?

jafingerhut:


jafingerhut:
I believe that test_packet_redirect is going into an infinite loop on the test named 'Recirculate' with these changes, because that test is written assuming that a recirculated packet reaches the egress with instance_type=6.  If we think this change is a correct one to make, that unit test needs to be updated in some way.

jafingerhut:
As an aside, at least on my Ubuntu 16.04 Linux VM, it seems like several of the tests run via 'make check' fail, or 'hang forever', fairly often, e.g. test_learning, test_meters, test_ageing

antoninbas:
Feel free to change the recirculation test in test_packet_redirect so that it works with your changes.
I haven't run into any issues with the test you mention in the past, and I don't recall ever seeing them hang or fail on Travis. The only problematic test that comes to mind is test_queueing, which tends to fail on slower machines. I guess the tests you list may also fail on a machine under *very* heavy load since there is a timing component to them. Maybe it would be worth compiling with O2 and without logging and see if it gets rid of the issue. Or maybe Cisco should buy you a better laptop :P

antoninbas:
Note: test_learning and test_ageing both rely on the same code to receive notifications (https://github.com/p4lang/behavioral-model/blob/master/tests/utils.h#L32) so my first impulse was to look there. However, test_meters doesn't rely on this code. So the only common factor between all these tests is that they are time-dependent (they use sleeps & timeouts).

jafingerhut:
Thanks for the feedback.  I will think a bit about how to update the test_packet_redirect named 'Recirculate' so that it works with these changes.  Many possibilities there, but it would be nice to find something that is a small a change to the existing test.

jafingerhut:
@antoninbas There is no hurry on your taking a look at these changes, but I have gotten all of the tests passing, only 2 of them needing small tweaks to continue working because they had been written to assume that on egress a recirculated packet had instance_type != 0, copied from the ingress value.

I have run 'make check' with p4c and these changes in my local copy of simple_switch, and they all passed, too, so it appears that none of the STF tests relied on the previous behavior.  Not surprising, given how few STF tests there are that do recirculate/resubmit/etc. operations.

antoninbas:
@jafingerhut possible idea for the tests that were failing: if you run bmv2 as sudo (e.g. to send / receive packets on veth interfaces) then run "make check" (without sudo), some tests can fail because of left over .ipc files under /tmp/ which have the wrong permissions.
The best way to get around this is to remove those *.ipc files before running make check...

PR #730, opened 2019-03-07
Fix ingress thread deadlock for resubmit packets

### LINKED ISSUE #729: Is there an input queue in ingress port to storage packets arriving at line rate?
Hi there.

I'm a begginer in P4 and BMv2. As far I know when we resubmit or recirculate a specific packet in bmv2, it will be reinjected at same ingress port that it arrived, right? My question is, what's happen when we resubmit or recirculate a packet in bmv2 but at the same time the same ingress port is receiving another packet? Which one would be parsed first, for example? Is there an input queue at ingress port to storage packets when bmv2 is receiving or parsing a resubmited/recirculated packet?

Thanks in advance.

antoninbas:
Resubmit packets were written to the input_buffer with a blocking call
by the ingress thread. Since the ingress thread is also in charge of
draining the input_buffer, this could have lead to a deadlock. Resubmit
packets can now be dropped if the buffer is full. To limit the number of
resubmit packets being lost, we place them is a higher priority queue
than "normal" packets.

Fixes #729

### DIFF
--- a/include/bm/bm_sim/queueing.h
+++ b/include/bm/bm_sim/queueing.h
@@ -397,9 +397,9 @@ class QueueingLogicRL {
 //! high priority can starve lower-priority queues. For example, if the queue
 //! with priority `0` always contains at least one element, the other queues
 //! will never be served.
-//! As for QueueingLogicRL, the write behavior (push_front()) is blocking: once
-//! a logical queue is full, subsequent incoming elements will be dropped until
-//! the queue starts draining again.
+//! As for QueueingLogicRL, the write behavior (push_front()) is not blocking:
+//! once a logical queue is full, subsequent incoming elements will be dropped
+//! until the queue starts draining again.
 //! Look at the documentation for QueueingLogic for more information about the
 //! template parameters (they are the same).
 template <typename T, typename FMap>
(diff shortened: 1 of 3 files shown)


include/bm/bm_sim/queueing.h
```diff
@@ -397,9 +397,9 @@ class QueueingLogicRL {
 //! high priority can starve lower-priority queues. For example, if the queue
 //! with priority `0` always contains at least one element, the other queues
 //! will never be served.
-//! As for QueueingLogicRL, the write behavior (push_front()) is blocking: once
-//! a logical queue is full, subsequent incoming elements will be dropped until
-//! the queue starts draining again.
+//! As for QueueingLogicRL, the write behavior (push_front()) is not blocking:
```
jafingerhut: Sorry, I do not know the BMv2 internals very well at all, so just throwing up a random red flag on comment changes -- After your changes is the phrase "As for QueueingLogicRL" still correct?

antoninbas: I didn't change anything in this file. I realized this comment was not correct so I fixed it, but it has nothing to do with the rest of the changes.

PR #736, opened 2019-03-17
add psa metadata requirements for psa_switch startup

peteli3:
changelog:
- remove v1model metadata header requirements from `psa_switch.cpp`
- add psa metadata header requirements to `psa_switch.cpp`


jafingerhut:
Looks reasonable to me.  I have commented to Peter in a separate channel that the force_arith_header calls on the new PSA headers will probably be necessary, but those can be added later when the code to access those fields is added.

PR #737, opened 2019-03-18
Added support for assert and assume primitives in bm_sim

enisah:
* In simulation, if condition of assert evaluates to false,
assert error message will be printed and packet will be dropped

* If condition of assume evaluates to true, assume error message
will be printed and packet will be dropped

Signed-off-by: Enisa Hadzic <enisa.hadzic@rt-rk.com>

### DIFF
--- a/include/bm/bm_sim/_assert.h
+++ b/include/bm/bm_sim/_assert.h
@@ -22,10 +22,13 @@
 #define BM_BM_SIM__ASSERT_H_
 
 // An assert that cannot be removed with NDEBUG
+#include <bm/bm_sim/source_info.h>
 
 namespace bm {
 
 [[ noreturn ]] void _assert(const char* expr, const char* file, int line);
+
+void error_message(const char* error_msg, const SourceInfo* srcInfo);
 
 }  // namespace bm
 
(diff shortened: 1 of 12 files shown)


jafingerhut:
You wrote "If condition of assume evaluates to true, assume error message
will be printed and packet will be dropped".  Did you mean "false" instead of "true"?

jafingerhut:
Looking at the code, it looks like it matches your statement in the comment above.

It seems better to fit the English meaning of the word "assume" if a simulator like behavioral-model simple_switch would cause a failure/error message if the assume condition evaluated to false.  If it evaluates to true, then your assumption was correct (in this case), so nothing has gone wrong.

If you are wondering why assert and assume are separate statements, it is not because of their behavior during dynamic execution, like in simple_switch.  It is in a formal proof system where they would behave differently, as I tried to explain in the issue here: https://github.com/p4lang/p4c/issues/1548

enisah:
Thank you for all comments.
Yes, I thought that the difference is in their behavior during execution, that is why I understood that assume should report error when condition evaluates to true. I will try to understand more about the differences from your explanation above.

PR #744, opened 2019-03-26
Add docs bmv2 direct counters meters

jafingerhut:


jafingerhut:
Only documentation changes here, so Travis being a little too sensitive here again.  I can try pushing a commit to force it to re-run if you want, but probably best done by making actual substantive changes, if any are suggested.

docs/simple_switch.md
```diff
@@ -656,3 +656,31 @@ if `max >= 1`, otherwise `base`.
...
+having a table property `counters = c` in its definition, the BMv2
+v1model implementation behaves as if every action for that table
+contains exactly one call to `c.count()`, whether it has none, one, or
+more than one.
+
+
+### BMv2 `direct_meter` implementation notes
+
+If a table `t` has a `direct_meter` object `m` associated with it by
+having a table property `meters = m` in its definition, then _at least
```
antoninbas: I was not aware of this, what does the compiler do if there is no call to `m.read(...)`? Does it complain because it cannot deduce the output field?

jafingerhut: I believe the latest p4c issues a CompilerBug and asserts.  I can create a separate issue on p4c for that, since it shouldn't be a CompilerBug, but ideally a slightly more informative message explaining why the program will not compile.

jafingerhut: Yes, CompilerBug today.  Issue here: https://github.com/p4lang/p4c/issues/1829

PR #745, opened 2019-03-26
Avoid crash on hash with max=0, and warn if lo > hi for random

jafingerhut:


targets/simple_switch/primitives.cpp
```diff
@@ -264,8 +274,19 @@ class modify_field_with_hash_based_offset
                            const NamedCalculation &, const Data &> {
   void operator ()(Data &dst, const Data &base,
                    const NamedCalculation &hash, const Data &size) {
-    uint64_t v =
-      (hash.output(get_packet()) % size.get<uint64_t>()) + base.get<uint64_t>();
+    uint64_t b = base.get<uint64_t>();
```
antoninbas: can you use `auto` for these variable declarations, the type is obvious from the `get<uint64_t>`

targets/simple_switch/primitives.cpp
```diff
@@ -59,8 +59,18 @@ class modify_field_rng_uniform
...
-    f.set(distribution(generator));
+    uint64_t lo = b.get_uint64();
+    uint64_t hi = e.get_uint64();
+    std::string warning = "";
+    if (lo > hi) {
+        warning = " warning: result is not specified when lo > hi";
+    }
+    distrib64 distribution(lo, hi);
+    uint64_t rand_val = distribution(generator);
+    BMLOG_TRACE_PKT(get_packet(),
```
antoninbas: I'm not sure it is necessary to have such a detailed level of logging (I think we've had this discussion before), but if you want to keep it, I think the message would look better like this:
```
random(lo=..., hi=...) = ...
```

targets/simple_switch/primitives.cpp
```diff
@@ -264,8 +274,19 @@ class modify_field_with_hash_based_offset
...
+    uint64_t sz = orig_sz;
+    std::string warning = "";
+    if (sz == (uint64_t) 0) {
+        sz = (uint64_t) 1;
+        warning = " warning: max given as 0, but treating it as 1";
+    }
+    uint64_t h = hash.output(get_packet());
+    uint64_t v = (h % sz) + b;
+    BMLOG_TRACE_PKT(get_packet(),
+                    "hash base {} max {} raw_hash {} result {}{}",
```
antoninbas: same as above, I think I would prefer something like:
```
hash(...) = ...
```

targets/simple_switch/primitives.cpp
```diff
@@ -264,8 +274,19 @@ class modify_field_with_hash_based_offset
                            const NamedCalculation &, const Data &> {
   void operator ()(Data &dst, const Data &base,
                    const NamedCalculation &hash, const Data &size) {
-    uint64_t v =
-      (hash.output(get_packet()) % size.get<uint64_t>()) + base.get<uint64_t>();
+    uint64_t b = base.get<uint64_t>();
+    uint64_t orig_sz = size.get<uint64_t>();
+    uint64_t sz = orig_sz;
+    std::string warning = "";
+    if (sz == (uint64_t) 0) {
```
antoninbas: These cast are not necessary, if you want to be explicit use 0ull and 1ull

targets/simple_switch/primitives.cpp
```diff
@@ -59,8 +59,18 @@ class modify_field_rng_uniform
     using hash = std::hash<std::thread::id>;
     static thread_local engine generator(hash()(std::this_thread::get_id()));
     using distrib64 = std::uniform_int_distribution<uint64_t>;
-    distrib64 distribution(b.get_uint64(), e.get_uint64());
-    f.set(distribution(generator));
+    uint64_t lo = b.get_uint64();
+    uint64_t hi = e.get_uint64();
+    std::string warning = "";
+    if (lo > hi) {
+        warning = " warning: result is not specified when lo > hi";
```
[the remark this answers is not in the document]
jafingerhut: Mainly because of my ignorance of the code.  I can make updates based on this and your other comments.  I wasn't actually expecting these changes to pass review, but didn't know a faster way than giving something undesirable, to find out what was better.

jafingerhut:
Updates committed based on your feedback.

I do not intend to be too pushy when it comes to adding such logging messages, but there are times when they can be quite helpful in debugging P4 program behavior.  I can remove them if you have a strong preference.

targets/simple_switch/primitives.cpp
```diff
@@ -59,17 +60,16 @@ class modify_field_rng_uniform
...
     static thread_local engine generator(hash()(std::this_thread::get_id()));
     using distrib64 = std::uniform_int_distribution<uint64_t>;
-    uint64_t lo = b.get_uint64();
-    uint64_t hi = e.get_uint64();
-    std::string warning = "";
+    auto lo = b.get_uint64();
+    auto hi = e.get_uint64();
     if (lo > hi) {
-        warning = " warning: result is not specified when lo > hi";
+        Logger::get()->warn("random result is not specified when lo > hi");
```
antoninbas: you should exit from the primitive as this point, not construct the `distribution` object, because the constructor call has undefined behavior (and that undefined behavior could be a crash). I suggest one for the following:
 1. return without writing to the field
 2. write 0 to the field and return
 3. write a sentinel value to the field and return

Any of them is fine by me, maybe 1. is my favorite

jafingerhut: Commit pushed with changes for your choice 1.

PR #748, opened 2019-03-29
Implement mark_to_drop primitive for simple_switch

antoninbas:
The primitive takes a header as a parameter, whose header type must have
a field called 'egress_spec'. This is part of an attempt to fix
mark_to_drop support in v1model / simple_switch. Because the
implementation of the primitive modifies a user-exposed field, it should
not be done in a hidden way. The primitive still modifies
intrinsic_metadata.mcast_grp instead of looking for a 'mcast_grp' field
in the provided header, but fixing this requires a bigger change (we
need to deprecate intrinsic_metadata.mcast_grp in favor of
standard_metadata.mcast_grp).

### DIFF
--- a/targets/simple_switch/primitives.cpp
+++ b/targets/simple_switch/primitives.cpp
@@ -165,6 +165,47 @@ class drop : public ActionPrimitive<> {
 
 REGISTER_PRIMITIVE(drop);
 
+class mark_to_drop : public ActionPrimitive<Header &> {
+  void operator ()(Header &std_hdr) {
+    if (egress_spec_offset == -1) {
+      const auto &header_type = std_hdr.get_header_type();
+      egress_spec_offset = header_type.get_field_offset("egress_spec");
+      if (egress_spec_offset == -1) {
+        Logger::get()->critical(
+            "Header {} must be of type standard_metadata but it does not have "
+            "an 'egress_spec' field",
+            std_hdr.get_name());
+        return;
+      }
+
+      mcast_grp_offset = header_type.get_field_offset("mcast_grp");
+    }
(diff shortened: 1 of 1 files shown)


antoninbas:
See https://github.com/p4lang/p4c/issues/1828

mbudiu-vmw:
Please note that I have an outstanding PR https://github.com/p4lang/p4c/pull/1704 against the compiler which moves all intrinsic_metadata fields automatically to standard_metadata. Then you may not need to do anything to handle programs that come from P4-14 through p4c.

antoninbas:
@mbudiu-vmw I think the issue is a bit different here. Historically, simple_switch assumes that the multicast group field is in the intrinsic_metadata header (https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md#intrinsic-metadata), which is why we have all these `@alias` annotation in v1model.p4. If all the "intrinsic_metadata" fields are moved to "standard_metadata" in v1model.p4, I can get rid of "intrinsic_metadata" in simple_switch as well. If I do that, maybe we can get rid of all the `@alias` annotations in v1model as well (so that's a 3 step process). However that means breaking backward-compatibility with all existing JSON files and with the legacy p4c-bm compiler (which is probably not a big deal at this stage).

mbudiu-vmw:
I think we should be able to support both the legacy programs and the new ones somehow.

antoninbas:
Actually, that was a stupid comment on my part, `mark_to_drop` is a new primitive as far as simple_switch is concerned, so there is no backward compatibility concern. I'll fix my PR.

mbudiu-vmw:
This does not show up as merged yet.

PR #753, opened 2019-04-11
Add units for meter rate to documentation of simple_switch

hesingh:
Also, a cosmetic change was made to relevant code to move mutiple lines of code on single line to another line.

### DIFF
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -2,7 +2,7 @@
 TODO: Write help for any CLI not preceded by "###".
 ###
 
-act_prof_add_member_to_group
+`act_prof_add_member_to_group`
 act_prof_create_group
 act_prof_create_member
 act_prof_delete_group
@@ -29,6 +29,10 @@ meter_array_set_rates
 meter_get_rates
 
 ### meter_set_rates
+The syntax for the CLI is provided below.
+`RuntimeCmd: help meter_set_rates
+Configure rates for a meter: meter_set_rates <name> <index> <rate_1>:<burst_1> <rate_2>:<burst_2> ...`
+
 Meter behavior is specified in RFC 2698.
 
 The user enters the meter rate in bytes/microsecond and burst_size in
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -2,84 +2,84 @@
 TODO: Write help for any CLI not preceded by "###".
 ###
 
-`act_prof_add_member_to_group`
-act_prof_create_group
-act_prof_create_member
-act_prof_delete_group
-act_prof_delete_member
-act_prof_dump
-act_prof_dump_group
-act_prof_dump_member
-act_prof_modify_member
-act_prof_remove_member_from_group
-counter_read
-counter_reset
-help
-load_new_config_file
-mc_dump
-mc_mgrp_create
-mc_mgrp_destroy
-mc_node_associate
-mc_node_create
-mc_node_destroy
-mc_node_dissociate
-mc_node_update
-mc_set_lag_membership
-meter_array_set_rates
-meter_get_rates
+act_prof_add_member_to_group  
+act_prof_create_group  
+act_prof_create_member  
+act_prof_delete_group  
+act_prof_delete_member  
+act_prof_dump  
+act_prof_dump_group  
+act_prof_dump_member  
+act_prof_modify_member  
+act_prof_remove_member_from_group  
+counter_read  
+counter_reset  
+help  
+load_new_config_file  
+mc_dump  
+mc_mgrp_create  
+mc_mgrp_destroy  
+mc_node_associate  
+mc_node_create  
+mc_node_destroy  
+mc_node_dissociate  
+mc_node_update  
+mc_set_lag_membership  
+meter_array_set_rates  
+meter_get_rates  
+  
+### meter_set_rates  
+The syntax for the CLI is provided below.  
+`RuntimeCmd: help meter_set_rates  
+Configure rates for a meter: meter_set_rates <name> <index> <rate_1>:<burst_1> <rate_2>:<burst_2> ...`  
+  
+Meter behavior is specified in RFC 2698.  
 
-### meter_set_rates
-The syntax for the CLI is provided below.
-`RuntimeCmd: help meter_set_rates
-Configure rates for a meter: meter_set_rates <name> <index> <rate_1>:<burst_1> <rate_2>:<burst_2> ...`
-
-Meter behavior is specified in RFC 2698.
-
-The user enters the meter rate in bytes/microsecond and burst_size in
-bytes.  If the meter type is packets, the rate is entered in packets/microsecond
-and burst_size is the number of packets.
-
-port_add
-port_remove
-register_read
-register_reset
-register_write
-reset_state
-serialize_state
-set_crc16_parameters
-set_crc32_parameters
-shell
-show_actions
-show_ports
-show_tables
-swap_configs
-switch_info
-table_add
-table_clear
-table_delete
-table_dump
-table_dump_entry
-table_dump_entry_from_key
-table_dump_group
-table_dump_member
-table_indirect_add
-table_indirect_add_member_to_group
-table_indirect_add_with_group
-table_indirect_create_group
-table_indirect_create_member
-table_indirect_delete
-table_indirect_delete_group
-table_indirect_delete_member
-table_indirect_modify_member
-table_indirect_remove_member_from_group
-table_indirect_reset_default
-table_indirect_set_default
-table_indirect_set_default_with_group
-table_info
-table_modify
-table_num_entries
-table_reset_default
-table_set_default
-table_set_timeout
-table_show_actions
-write_config_to_file
+The user enters the meter rate in bytes/microsecond and burst_size in  
+bytes.  If the meter type is packets, the rate is entered in packets/microsecond  
+and burst_size is the number of packets.  
+  
+port_add  
+port_remove  
+register_read  
+register_reset  
+register_write  
+reset_state  
+serialize_state  
+set_crc16_parameters  
+set_crc32_parameters  
+shell  
+show_actions  
+show_ports  
+show_tables  
+swap_configs  
+switch_info  
+table_add  
+table_clear  
+table_delete  
+table_dump  
+table_dump_entry  
+table_dump_entry_from_key  
+table_dump_group  
+table_dump_member  
+table_indirect_add  
+table_indirect_add_member_to_group  
+table_indirect_add_with_group  
+table_indirect_create_group  
+table_indirect_create_member  
+table_indirect_delete  
+table_indirect_delete_group  
+table_indirect_delete_member  
+table_indirect_modify_member  
+table_indirect_remove_member_from_group  
+table_indirect_reset_default  
+table_indirect_set_default  
+table_indirect_set_default_with_group  
+table_info  
+table_modify  
+table_num_entries  
+table_reset_default  
+table_set_default  
+table_set_timeout    
+table_show_actions  
+write_config_to_file  
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -1,7 +1,81 @@
-### BMv2 meter rate and burst_size units
+###
+TODO: Write help for any CLI not preceded by "###".
+###
+
+act_prof_add_member_to_group
+act_prof_create_group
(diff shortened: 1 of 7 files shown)


hesingh:
Please review this PR - thanks.

docs/simple_switch.md
```diff
@@ -685,3 +685,8 @@ you have two actions for table `t` where one has the action
 `m.read(result_field2)`, or if both of those calls are in the same
 action.  All calls to the `read()` method for `m` must have the same
 result parameter where the result is written.
+
+### BMv2 `meter rate units` implementation notes
+
+The user enters the meter rate and burtsize in
```
jafingerhut: It sounds reasonable that the meter _rate_ is in units of bytes per microsec, but the units of "burst size" should be some kind of unit like bytes.  It has nothing to do with time.  I do not know if simple_switch uses bytes or some other unit like kilobytes.

jafingerhut: From the names of variables and some of the math done to convert units, it appears like perhaps the burst sizes are specified in multiples of 1000 bits (Kbits would be confusing to many people, I think, as it would often be interpreted as multiples of 1024 bits).

jafingerhut: Also, these are likely the units used when a meter has meter type equal to MeterType.bytes in v1model.

Does BMv2 support different behavior for meters that have a type equal to MeterType.packets ?

If so, what are the units of the meter rates and burst sizes for such meters?

hesingh: > It sounds reasonable that the meter _rate_ is in units of bytes per microsec, but the units of "burst size" should be some kind of unit like bytes. It has nothing to do with time. I do not know if simple_switch uses bytes or some other unit like kilobytes.

Sorry, I missed the burst_size in bytes.  I will change the doc.

hesingh: > From the names of variables and some of the math done to convert units, it appears like perhaps the burst sizes are specified in multiples of 1000 bits (Kbits would be confusing to many people, I think, as it would often be interpreted as multiples of 1024 bits).

Makes sense to use number of kilo octets.

hesingh: > Also, these are likely the units used when a meter has meter type equal to MeterType.bytes in v1model.
> 
> Does BMv2 support different behavior for meters that have a type equal to MeterType.packets ?
> 
> If so, what are the units of the meter rates and burst sizes for such meters?

pps and packets.

hesingh:
The internal storage for meter rate by simple_switch is using kbps.  Thus, doesn't it make sense to change the documentation to let the user specify rate in kbps?  The code takes the rate in kbps and converts to bytes/microsecond.

pdfixed/include/bm/pdfixed/pd_common.h
```diff
@@ -73,7 +73,7 @@ typedef struct p4_pd_packets_meter_spec {
 
 typedef struct p4_pd_bytes_meter_spec {
   uint32_t cir_kbps;
-  uint32_t cburst_kbits;
```
antoninbas: we are not changing the interface, unless there is a bug that needs to be fixed

hesingh: There is no bug in the code.  However, one reviewer has already made a comment saying "burst sizes are specified in multiples of 1000 bits (Kbits would be confusing to many people, I think, as it would often be interpreted as multiples of 1024 bits)."  To address the comment, I decided to use kilo octets for burst_size.  On second thought, I think "kilo" is still confusing because folks will ask if it is 1024 or 1000?  

Why not use octets for burst_size? 

A size in several specifications uses octets.  See RFC 8200 for IPv6 Payload length which uses octets.

jafingerhut: If the code comments and conversions are correct, it is not kilo octets.  It is 10^3 _bits_.  Or you could say it is in multiples of 125 bytes/octets.

hesingh: i know the code conversion is correct.  But I changed the code to use kilo octets.  But kilo is ambiguous.  So, we could use `10^3 octets`.

hesingh: Or `10^3 bytes` would work too.  Then the original code conversion remains unchanged.

hesingh:
Please see if latest changes work for reviewers - thanks.

docs/simple_switch.md
```diff
@@ -685,3 +685,12 @@ you have two actions for table `t` where one has the action
 `m.read(result_field2)`, or if both of those calls are in the same
 action.  All calls to the `read()` method for `m` must have the same
 result parameter where the result is written.
+
+### BMv2 `meter rate units` implementation notes
```
antoninbas: what is this documentation for? The runtime CLI? The Thrift service? The PD API?

pdfixed/src/pd_helpers.cpp
```diff
@@ -33,13 +33,16 @@ pd_bytes_meter_spec_to_rates(p4_pd_bytes_meter_spec_t *meter_spec) {
 
   // bytes per microsecond
   info_rate = static_cast<double>(meter_spec->cir_kbps) / 8000.;
+  // 10^6 bytes
```
antoninbas: that's incorrect
it's converting it to bytes, not Mbytes

docs/simple_switch.md
hesingh: simple_switch_CLI

pdfixed/src/pd_helpers.cpp
hesingh: Oops, you are correct.

hesingh:
> I don't see the value of this PR. The simple_switch.md documentation is supposed to help people write P4 programs for v1model.p4 and potentially write control-plane applications for them. But there are many tools to set meter rates & burst sizes, and the change does not provide any context.

In the P4 Slack general channel today, two users asked the units for meter rate and burst_size because they weren't sure.  So, it was asked to document, at least, what the simple_switch_CLI uses.   The code changes are cosmetic.  I updated the PR.

docs/simple_switch.md
antoninbas: that's not accurate then
IIRC, the burst size is in "units" (packets or bytes) and the rate is in "units per microsecond" (https://github.com/p4lang/behavioral-model/blob/master/thrift_src/standard.thrift#L88)

hesingh: Wow, even I quoted the same line of code in the Slack channel today.  But, seeing the code in pd_helpers.cpp, I saw the conversion from kbps and kbits.  This cpp code also happens to be the only code in the behavioral-model repo converting rate and burst_size and thus I assumed the code is used with simpe_switch_CLI as well.

hesingh: > that's not accurate then
> IIRC, the burst size is in "units" (packets or bytes) and the rate is in "units per microsecond" (https://github.com/p4lang/behavioral-model/blob/master/thrift_src/standard.thrift#L88)

There is some confusion.  Based on your reply for meter type as packet, its "units in microsecond" and this is the relevant code in pd_helpers.cpp. I see the division by 10^6 for microseconds.

`info_rate = static_cast<double>(meter_spec->cir_pps) / 1000000.;`

Isn't the user input captured by simple_switch_CLI in meter_spec->cir_pps and thus the user inputs in pps units?  Or is the code in pd_helpers.cpp not used at all by simple_switch_CLI?

antoninbas: The PD layer is not used at all by the CLI. If you look at the CLI code, you will see that it calls the Thrift service directly.
It is fine to have this information in simple_switch.md BTW, but it would be far more useful to improve the runtime_CLI help message for the `meter_set_rates` CLI command, so that users can get this information directly from the CLI a they are using it.

hesingh: Ok, I will edit the simple_switch.md doc for CLI rates and burst_size.  Even, I was thinking of adding the help message for the meter_set_rates.  I will add that.  Thanks!

hesingh:
Unit test

```
RuntimeCmd: help meter_set_rates
Configure rates for a meter: meter_set_rates <name> <index> <rate_1>:<burst_1> <rate_2>:<burst_2> ...
Rate uses units/microsecond and burst uses units where units is bytes or packets
RuntimeCmd: 
```

hesingh:
All checks have passed.  Please review final diffs.

hesingh:
@antoninbas I restored the pdfixed file to its original form.  Now the diffs have only two files changed.  Please review - thanks.

hesingh:
Any other reviewers to help review and check in this PR? @antoninbas seems busy.  The PR has only changed one .md file for documentation and added help for one CLI in Python code.

docs/simple_switch.md
```diff
@@ -685,3 +685,12 @@ you have two actions for table `t` where one has the action
 `m.read(result_field2)`, or if both of those calls are in the same
 action.  All calls to the `read()` method for `m` must have the same
 result parameter where the result is written.
+
+### BMv2 `meter rate and burst_size units` implementation notes.
```
antoninbas: Why is this between backticks?
why do you think this qualifies as "implementation notes"?

docs/simple_switch.md
```diff
@@ -685,3 +685,12 @@ you have two actions for table `t` where one has the action
...
+### Units help configuring meter with simple_switch_CLI
```
antoninbas: This is not specific to simple_switch so I don't understand why we are putting this in this document.
Why not start a separate runtime_CLI documentation if this is your intent?

hesingh: On 04/10, in the P4 Slack channel, a note said to add the info to simple_switch.md.  The note is copied below.

"If you learn the answer, it would be good to document the answer in this file: https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md"

After all, p4runtime_CLI does not run unless simple_switch is running first.  

Anyway, your point makes sense.  Let me add a new file and move the documentation there.  thanks.

antoninbas: > After all, p4runtime_CLI does not run unless simple_switch is running first.

That is not accurate. p4runtime_CLI works with any bmv2 target. It does not have to be simple_switch.

hesingh: I see - thanks.  Sorry, I haven't used other bmv2 targets and have used p4runtime_CLI only with simple_switch.  If I do not launch the simple_switch, I see this error.

```
hemant@ubuntu:~/behavioral-model/tools$ ./runtime_CLI.py
Could not connect to thrift client on port 9090
Make sure the switch is running and that you have the right port
hemant@ubuntu:~/behavioral-model/tools$ 
```

docs/simple_switch.md
hesingh: I removed the backticks since the exact names of the CLI args are not used.  Also removed "implemented notes"

hesingh:
I don't understand the `codecov/project` failure.  I think, it doesn't look like a blocking issue to merge the changes.

docs/runtime_CLI.md
```diff
@@ -0,0 +1,7 @@
+### BMv2 meter rate and burst_size units
```
antoninbas: I think if we are going to start a CLI documentation we should do it right and organize it by command. For the `meter_set_rates` command, more information may be useful, and in particular a pointer to RFC 2698.

docs/runtime_CLI.md
```diff
@@ -0,0 +1,7 @@
...
+
+The user enters the meter rate in bytes/microsecond and burst_size in
+bytes.
+
+If the meter type is packets, the rate is enterted in packets/microsecond and
+burst_size is number of packets.
```
antoninbas: is *the* number of packets

hesingh:
@antoninbas Please review the last changes and if they look good, please merge this code - thanks.

docs/runtime_CLI.md
```diff
@@ -0,0 +1,81 @@
...
+mc_node_dissociate
+mc_node_update
+mc_set_lag_membership
+meter_array_set_rates
+meter_get_rates
+
+### meter_set_rates
+Meter behavior is specified in RFC 2698.
+
+The user enters the meter rate in bytes/microsecond and burst_size in
```
antoninbas: why not give the command syntax here as well?

hesingh:
> Have you taken a look at the rendered file on Github?

Wow, did not.  Good catch!  I checked in code to fix rendering.  Syntax of CLI was added by a prior checkin.  Code is ready for review.

hesingh:
@antoninbas Travis passed with latest changes(add syntax and fix rendering) and only codecov has failed.

PR #755, opened 2019-04-12
Psa ucast mcast demo

derekso1:
Made minimal changes to psa_switch to support.

changelog:

- psa_switch's main control-flow functions (receive_, ingress_thread, egress_thread, transmit_thread), v1model standard metadata references were removed, replaced with psa arch-specific metadata needed for unicast/multicast
    
- ingress deparser and egress parser added and applied in appropriate parts of main control-flow
- minimal logging added for debugging purposes

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -345,16 +345,12 @@ PsaSwitch::ingress_thread() {
     BMLOG_DEBUG_PKT(*packet, "Processing packet received on port {}",
                     ingress_port);
 
-    std::cout << "Ready to parse and pipe packet\n";
     parser->parse(packet.get());
-    std::cout << "parsed correctly\n";
     ingress_mau->apply(packet.get());
-    std::cout << "piped correctly\n";
 
     packet->reset_exit();
     Field &f_egress_spec = phv->get_field("psa_ingress_output_metadata.egress_port");
     port_t egress_spec = f_egress_spec.get_uint();
-    std::cout << egress_spec << std::endl;
     port_t egress_port = egress_spec;
 
     egress_port = egress_spec;
@@ -364,7 +360,8 @@ PsaSwitch::ingress_thread() {
       BMLOG_DEBUG_PKT(*packet, "Dropping packet at the end of ingress");
       continue;
     }
-
+    Deparser *deparser = this->get_deparser("ingress_deparser");
+    deparser->deparse(packet.get());
     enqueue(egress_port, std::move(packet));
   }
 }
@@ -383,15 +380,14 @@ PsaSwitch::egress_thread(size_t worker_id) {
     egress_buffers.pop_back(worker_id, &port, &packet);
 #endif
     if (packet == nullptr) break;
-    std::cout << "packet in egress" << std::endl;
-    std::cout << port << std::endl;
 
+    Parser *parser = this->get_parser("egress_parser");
+    parser->parse(packet.get());
     Deparser *deparser = this->get_deparser("egress_deparser");
     Pipeline *egress_mau = this->get_pipeline("egress");
     phv = packet->get_phv();
     egress_mau->apply(packet.get());
     deparser->deparse(packet.get());
-    std::cout << "ready to send out" << std::endl;
     output_buffer.push_front(std::move(packet));
   }
 }
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -154,7 +154,7 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
   phv->get_field("psa_ingress_parser_input_metadata.packet_path").set(PKT_INSTANCE_TYPE_NORMAL);
   // using packet register 0 to store length, this register will be updated for
   // each add_header / remove_header primitive call
-//   packet->set_register(PACKET_LENGTH_REG_IDX, len);
+   packet->set_register(PACKET_LENGTH_REG_IDX, len);
 //   phv->get_field("standard_metadata.packet_length").set(len);
 //   Field &f_instance_type = phv->get_field("standard_metadata.instance_type");
 //   f_instance_type.set(PKT_INSTANCE_TYPE_NORMAL);
@@ -346,6 +346,25 @@ PsaSwitch::ingress_thread() {
 
     parser->parse(packet.get());
     ingress_mau->apply(packet.get());
+
+    // Handleing multicast
+    unsigned int mgid = 0u;
+    if (phv->has_field("psa_ingress_output_metadata.multicast_group")) {
+      Field &f_mgid = phv->get_field("psa_ingress_output_metadata.multicast_group");
+      mgid = f_mgid.get_uint();
+      
+      const auto pre_out = pre->replicate({mgid});
+      auto packet_size = packet->get_register(PACKET_LENGTH_REG_IDX);
+      for(const auto &out : pre_out){
+        auto egress_port = out.egress_port;
+        std::unique_ptr<Packet> packet_copy = packet->clone_with_phv_ptr();
+        packet_copy->set_register(PACKET_LENGTH_REG_IDX, packet_size);
+        enqueue(egress_port, std::move(packet_copy));
+      }
+      continue;
+    }
+
+
 
     packet->reset_exit();
     Field &f_egress_spec = phv->get_field("psa_ingress_output_metadata.egress_port");
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -367,7 +367,6 @@ PsaSwitch::ingress_thread() {
 
 void
 PsaSwitch::egress_thread(size_t worker_id) {
-  PHV *phv;
 
   while (1) {
     std::unique_ptr<Packet> packet;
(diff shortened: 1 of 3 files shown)


targets/psa_switch/psa_switch.cpp
```diff
@@ -206,7 +208,9 @@ PsaSwitch::set_egress_queue_depth(size_t port, const size_t depth_pkts) {
 
 int
 PsaSwitch::set_all_egress_queue_depths(const size_t depth_pkts) {
-  egress_buffers.set_capacity_for_all(depth_pkts);
+  for (uint32_t i = 0; i < max_port; i++) {
+    set_egress_queue_depth(i, depth_pkts);
+  }
```
jafingerhut: Is this present in simple_switch code as well?  I would guess that when Antonin recently made the changes to allow a configurable max_port value, he might also have allowed the set of valid ports to not always be consecutive from [0, max_port-1], i.e. there might be "holes" in that number range that do not represent valid ports.  Not a big deal, but worth checking.

targets/psa_switch/psa_switch.cpp
```diff
@@ -333,115 +344,34 @@ PsaSwitch::ingress_thread() {
...
-    packet->reset_exit();
-
-    Field &f_egress_spec = phv->get_field("standard_metadata.egress_spec");
-    port_t egress_spec = f_egress_spec.get_uint();
-
-    Field &f_clone_spec = phv->get_field("standard_metadata.clone_spec");
-    unsigned int clone_spec = f_clone_spec.get_uint();
-
-    int learn_id = 0;
+    // Handleing multicast
```
jafingerhut: typo: "Handleing" -> "Handling"

jafingerhut:
It can be part of a later commit if you prefer, but I believe added lines like shown in the diff below are necessary in order for the P4 program to be able to access these fields (or perhaps these calls enable psa_switch to access the fields -- I am not sure).  In any case, my testing with the drop-all example program crashes without these changes when psa_switch.cpp code tries to get the integer value of the multicast_group field.
```
diff --git a/targets/psa_switch/psa_switch.cpp b/targets/psa_switch/psa_switch.cpp
index 955ad75..928a13d 100644
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -122,6 +122,14 @@ PsaSwitch::PsaSwitch(port_t max_port, bool enable_swap)
 
   add_required_field("psa_egress_deparser_input_metadata", "egress_port");
 
+  force_arith_header("psa_ingress_parser_input_metadata");
+  force_arith_header("psa_ingress_input_metadata");
+  force_arith_header("psa_ingress_output_metadata");
+  force_arith_header("psa_egress_parser_input_metadata");
+  force_arith_header("psa_egress_input_metadata");
+  force_arith_header("psa_egress_output_metadata");
+  force_arith_header("psa_egress_deparser_input_metadata");
+
   import_primitives();
 }
```

Another thing that can be part of a future PR: Do not use the condition `egress_port == 511` to decide to drop the packet, but instead use the condition `drop == 1`, which is what the PSA spec says should be the deciding factor for dropping a packet.

targets/psa_switch/psa_switch.cpp
```diff
@@ -66,8 +66,9 @@ extern int import_primitives();
 
 packet_id_t PsaSwitch::packet_id = 0;
 
-PsaSwitch::PsaSwitch(bool enable_swap)
+PsaSwitch::PsaSwitch(port_t max_port, bool enable_swap)
```
antoninbas: This PR removes some of the recent changes I did to support arbitrary port numbers (https://github.com/p4lang/behavioral-model/pull/747), it would be great not to undo them

targets/psa_switch/psa_switch.cpp
```diff
@@ -149,13 +150,14 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
...
+  phv->get_field("psa_ingress_parser_input_metadata.ingress_port").set(port_num);
+  phv->get_field("psa_ingress_parser_input_metadata.packet_path").set(PKT_INSTANCE_TYPE_NORMAL);
   // using packet register 0 to store length, this register will be updated for
   // each add_header / remove_header primitive call
-  packet->set_register(PACKET_LENGTH_REG_IDX, len);
-  phv->get_field("standard_metadata.packet_length").set(len);
-  Field &f_instance_type = phv->get_field("standard_metadata.instance_type");
-  f_instance_type.set(PKT_INSTANCE_TYPE_NORMAL);
+   packet->set_register(PACKET_LENGTH_REG_IDX, len);
+//   phv->get_field("standard_metadata.packet_length").set(len);
```
antoninbas: do we need a TODO for this?

targets/psa_switch/psa_switch.cpp
```diff
@@ -149,13 +150,14 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
...
+//   Field &f_instance_type = phv->get_field("standard_metadata.instance_type");
```
antoninbas: seems that this can be removed since we use `"psa_ingress_parser_input_metadata.packet_path"`

targets/psa_switch/psa_switch.cpp
```diff
@@ -149,13 +150,14 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
 
   // setting standard metadata
 
-  phv->get_field("standard_metadata.ingress_port").set(port_num);
+  phv->get_field("psa_ingress_parser_input_metadata.ingress_port").set(port_num);
+  phv->get_field("psa_ingress_parser_input_metadata.packet_path").set(PKT_INSTANCE_TYPE_NORMAL);
```
antoninbas: I thought psa.p4 introduced an enum for this, why not use the appropriate enum member value that should be included in the JSON?

targets/psa_switch/psa_switch.cpp
```diff
@@ -333,115 +344,34 @@ PsaSwitch::ingress_thread() {
...
-    if (phv->has_field("intrinsic_metadata.mcast_grp")) {
-      Field &f_mgid = phv->get_field("intrinsic_metadata.mcast_grp");
+    if (phv->has_field("psa_ingress_output_metadata.multicast_group")) {
+      Field &f_mgid = phv->get_field("psa_ingress_output_metadata.multicast_group");
       mgid = f_mgid.get_uint();
-    }
-
-    port_t egress_port;
-
-    // INGRESS CLONING
```
antoninbas: are you removing a lot of cloning / resubmit / ... support with the intention of adding it back later in the correct form for PSA?

derekso1: Yes that is the plan.

targets/psa_switch/psa_switch.cpp
derekso1: Yes, seems like there was some accidental overwrites that we did, I have gone through and made sure they have all been reverted.

targets/psa_switch/psa_switch.cpp
derekso1: Accidental overwrites occurred, reverted.

targets/psa_switch/psa_switch.cpp
derekso1: Actually that's a great point. There has been an ongoing issue with enum translations in p4c so they currently aren't emitted in the JSON. However PSA_PacketPath_t is correctly translated just not emitted.

I can make changes to p4c, have those committed, and come back to this PR and have logic changed to use the enum member values in the JSON.

antoninbas: Since "normal" packet type will always be 0, you can just put a TODO as a remainder, then come back to it later. There is no need to delay merging this PR for this.

derekso1:
> It can be part of a later commit if you prefer, but I believe added lines like shown in the diff below are necessary in order for the P4 program to be able to access these fields (or perhaps these calls enable psa_switch to access the fields -- I am not sure). In any case, my testing with the drop-all example program crashes without these changes when psa_switch.cpp code tries to get the integer value of the multicast_group field.
> 
> ```
> diff --git a/targets/psa_switch/psa_switch.cpp b/targets/psa_switch/psa_switch.cpp
> index 955ad75..928a13d 100644
> --- a/targets/psa_switch/psa_switch.cpp
> +++ b/targets/psa_switch/psa_switch.cpp
> @@ -122,6 +122,14 @@ PsaSwitch::PsaSwitch(port_t max_port, bool enable_swap)
>  
>    add_required_field("psa_egress_deparser_input_metadata", "egress_port");
>  
> +  force_arith_header("psa_ingress_parser_input_metadata");
> +  force_arith_header("psa_ingress_input_metadata");
> +  force_arith_header("psa_ingress_output_metadata");
> +  force_arith_header("psa_egress_parser_input_metadata");
> +  force_arith_header("psa_egress_input_metadata");
> +  force_arith_header("psa_egress_output_metadata");
> +  force_arith_header("psa_egress_deparser_input_metadata");
> +
>    import_primitives();
>  }
> ```
> 
> Another thing that can be part of a future PR: Do not use the condition `egress_port == 511` to decide to drop the packet, but instead use the condition `drop == 1`, which is what the PSA spec says should be the deciding factor for dropping a packet.

I think this would go well with a later PR to fix the PacketPath Enum Antonin brought up.

derekso1:
All checks have passed with new changes.

Issue #727, opened 2019-03-04
packet.advance


mbudiu-vmw:
The p4-16 spec has a method on packet_in called advance which takes a 32-bit number. If the value is compile-time constant it's easy to implement this, but if it is variable we have to implement a parser loop. Should we consider adding it as a primitive to bmv2 instead?

antoninbas:
I can improve the "shift" parser primitive operation to support arbitrary expressions.

mbudiu-vmw:
What is the constraint today on the argument of shift?

mbudiu-vmw:
It looks like it must be a constant.

antoninbas:
Correct. It will be easy to support an expression. I'll see if I can alias it to "advance" as well, to make the bmv2 logs easier to decipher.

mbudiu-vmw:
The answer should probably depend on whether this is something that hardware will support easily.
If not, maybe it has to be done by the compiler in software.

mbudiu-vmw:
But it does look like the easiest way to implement it is allowing expressions in bmv2. Thank you.

antoninbas:
I opened a PR for this. In your case, you are more likely to use the "field" parameter type I think (or an "expression" parameter which just loads a field value).
For example, `pkt.advance(pkt.lookahead<bit<32>>() + 32w16)` will need to be transformed to:
```
bit<32> tmp;
tmp = pkt.lookahead<bit<32>>();  // "set" operation from a lookahead value
tmp = tmp + 32w16;  // "set" operation from an expression
pkt.advance(tmp);  // "advance" operation from field
```
or
```
bit<32> tmp;
tmp = pkt.lookahead<bit<32>>();  // "set" operation from a lookahead value
pkt.advance(tmp + 32w16);  // "advance" operation from expression
```
Creating a temporary field is only required when the input expression includes a lookahead operation, because the bmv2 expression evaluation engine doesn't support lookahead. This is the same as for variable-length extraction BTW.

mbudiu-vmw:
Yes, the compiler can do these rewrites easily.

mbudiu-vmw:
I have created a sample program that uses advance, but it does not seem to have any effect on the packet. I am attaching the P4 and JSON.

[issue1755-bmv2.txt](https://github.com/p4lang/behavioral-model/files/2942643/issue1755-bmv2.txt)

[...]

PR #757, opened 2019-04-15
Add ParserInvalidArgument parser error

jafingerhut:


jafingerhut:
Intended as a follow-up to completing the work started with these issues/PRs:
https://github.com/p4lang/behavioral-model/issues/727
https://github.com/p4lang/p4c/pull/1855

src/bm_sim/parser.cpp
```diff
@@ -259,8 +259,8 @@ void extract_VL(Header *hdr,
 
   VL_expr.eval(*phv, &computed_nbits);
   auto nbits = computed_nbits.get<int>();
-  // TODO(antonin): temporary limitation?
-  _BM_ASSERT(nbits % 8 == 0 && "VL field bitwidth needs to be a multiple of 8");
+  if ((nbits % 8) != 0)
+    throw parser_exception_core(ErrorCodeMap::Core::ParserInvalidArgument);
```
antoninbas: Could we log an error as well?

src/bm_sim/parser.cpp
```diff
@@ -465,9 +465,8 @@ ParserOpAdvance<Data>::operator()(Packet *pkt, const char *data,
   // The JSON should never specify a negative constant as the shift amount.
   assert(shift_bits.sign() >= 0);
   const auto shift_bits_uint = shift_bits.get<size_t>();
-  // TODO(antonin): temporary limitation?
-  _BM_ASSERT(shift_bits_uint % 8 == 0 &&
-             "advance bits amount must be a multiple of 8");
+  if ((shift_bits_uint % 8) != 0)
```
antoninbas: Could we log an error here too?
I also counted 2 other instances of that check in the file.

jafingerhut:
@antoninbas Regarding this comment of yours: "Could we add unit tests like this one (https://github.com/p4lang/behavioral-model/blob/master/tests/test_parser.cpp#L1570) for the new error?"

I have written a P4_16 program with STF test that causes the new error code to be exercised, with an output packet containing an integer code that distinguishes between all of the possible parser errors, including the new one.  Would that be sufficient?

I ask primarily because that style of unit test would probably take me a lot longer to figure out how to write for a 2-argument extract method call.

src/bm_sim/parser.cpp
jafingerhut: done

src/bm_sim/parser.cpp
jafingerhut: done

antoninbas:
I prefer to always have some unit tests which are part of CI, but we can merge this now and I'll see if I have time to add them myself later.

PR #758, opened 2019-04-15
psa_switch support recirculate

derekso1:
Changes:

- Added support for packet recirculation in psa_switch
- accompanying STF test is currently in [this PR](https://github.com/p4lang/p4c/pull/1857)

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -155,14 +155,12 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
   phv->get_field("psa_ingress_parser_input_metadata.packet_path").set(PKT_INSTANCE_TYPE_NORMAL);
   // using packet register 0 to store length, this register will be updated for
   // each add_header / remove_header primitive call
-   packet->set_register(PACKET_LENGTH_REG_IDX, len);
+  packet->set_register(PACKET_LENGTH_REG_IDX, len);
    // TODO
 //   phv->get_field("standard_metadata.packet_length").set(len);
 
-  if (phv->has_field("intrinsic_metadata.ingress_global_timestamp")) {
-    phv->get_field("intrinsic_metadata.ingress_global_timestamp")
-        .set(get_ts().count());
-  }
+  phv->get_field("psa_ingress_input_metadata.ingress_timestamp")
+    .set(get_ts().count());
 
   input_buffer.push_front(std::move(packet));
   return 0;
@@ -378,6 +376,7 @@ PsaSwitch::ingress_thread() {
 
 void
 PsaSwitch::egress_thread(size_t worker_id) {
+  PHV *phv;
 
   while (1) {
     std::unique_ptr<Packet> packet;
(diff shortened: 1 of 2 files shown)


targets/psa_switch/psa_switch.cpp
```diff
@@ -396,6 +396,20 @@ PsaSwitch::egress_thread(size_t worker_id) {
     Pipeline *egress_mau = this->get_pipeline("egress");
     egress_mau->apply(packet.get());
     deparser->deparse(packet.get());
+    
+    if (port == PSA_PORT_RECIRCULATE) {
```
derekso1: Currently this does a packet copy whenever we recirculate a packet. This is very similar to how simple_switch handles recirculation. Seems like its a possibility to remove the packet copying? Thoughts?

targets/psa_switch/psa_switch.cpp
```diff
@@ -396,6 +396,20 @@ PsaSwitch::egress_thread(size_t worker_id) {
...
+      BMLOG_DEBUG_PKT(*packet, "Recirculating packet");
+      std::unique_ptr<Packet> packet_copy = packet->clone_no_phv_ptr();
+      PHV *phv_copy = packet_copy->get_phv();
+      phv_copy->reset_metadata();
+      phv_copy->get_field("psa_ingress_input_metadata.packet_path")
```
derekso1: As mentioned in a previous PR, we should be using the PACKET_PATH enum from JSON. Compiler support is currently pending approval on p4c PR.

jafingerhut: I think there are 2 different packet_path metadata fields that must be assigned here, the one you have already, plus the one in `psa_ingress_parser_input_metadata`

jafingerhut: The PSA spec also says that recirculated packets should have both of these fields assigned the value of PSA_RECIRCULATE_PORT for recirculated packets.  That should be done somewhere near this line of code, too:

+ psa_ingress_parser_input_metadata.ingress_port
+ psa_ingress_input_metadata..ingress_port

jafingerhut: Another field: psa_ingress_input_metadata.ingress_timestamp should have the current time stored into it when a packet is recirculated, according to the PSA spec.

targets/psa_switch/psa_switch.cpp
jafingerhut: This comment isn't helpful, other than to say I do not know the answer here.  I am hoping @antoninbas might have time to look at this question and say what he knows of it?

antoninbas: By the way you a `reset_metadata` but aren't you supposed to preserve the recirculation metadata?

jafingerhut: I think the idea is to get recirculate working first, without preserving metadata, and to save for later whatever changes are required to get recirculate while preserving metadata working.

derekso1: Yes that is the plan.

derekso1:
I removed packet cloning in recirculation and updated the psa metadata (apologies slipped my mind). Also updated the ingress timestamp in receive_ since that was not done previously.

PR #762, opened 2019-04-19
Added support for assert and assume primitives in bm_sim

enisah:
* Assert and assume primitives are added as part of the implementation
assert and assume support in p4c with id - #1859

* In simulation, if condition of assert/assume evaluates to false,
assert error message will be printed and simulation will be stopped

Signed-off-by: Enisa Hadzic <enisa.hadzic@rt-rk.com>


targets/psa_switch/primitives.cpp
```diff
@@ -398,6 +398,22 @@ class truncate_ : public ActionPrimitive<const Data &> {
 
 REGISTER_PRIMITIVE_W_NAME("truncate", truncate_);
 
+class assert_ : public ActionPrimitive<const Data &> {
```
antoninbas: I don't understand why these are not added to the core primitives: https://github.com/p4lang/behavioral-model/blob/master/src/bm_sim/core/primitives.cpp

tests/test_assert_assume.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+#include <bm/bm_sim/actions.h>
+#include <bm/bm_sim/core/primitives.h>
+#include <bm/bm_sim/phv.h>
+#include <bm/bm_sim/parser_error.h>
+#include <bm/bm_sim/packet.h>
+
+using namespace bm;
+
+class assert_ : public ActionPrimitive<const Data &> {
```
antoninbas: if you define them in the core primitives, you won't need to duplicate the code

tests/test_assert_assume.cpp
```diff
@@ -0,0 +1,127 @@
...
+    std::unique_ptr<ArithExpression> condition(expr);
+    testActionFn.parameter_push_back_expression(std::move(condition));
+
+    testActionFnEntry(pkt.get());
+
+    check_result(withError);
+  }
+};
+
+TEST_F(AssertAssumeTest, AssumeBoolConstError) {
```
antoninbas: Thanks for adding a test. This seems like a perfect candidate for a value-parametrized google test
It will be less verbose and give better error logs in case of test failure

src/bm_sim/_assert.cpp
```diff
@@ -31,4 +32,16 @@ void _assert(const char* expr, const char* file, int line) {
   std::abort();
 }
 
+void abort_msg(const char* error_msg, const SourceInfo* srcInfo) {
```
antoninbas: it would be better IMO to move this function to https://github.com/p4lang/behavioral-model/blob/master/src/bm_sim/core/primitives.cpp along with the definition of `assert` and `assume`. This file was never intended to have a dependency on the logger.

src/bm_sim/actions.cpp
```diff
@@ -283,6 +283,9 @@ void
 ActionPrimitiveCall::execute(ActionEngineState *state,
                              const ActionParam *args) const {
   // TODO(unknown): log source info?
+  if (source_info != nullptr) {
```
antoninbas: is there really a need for a conditional here? it seems to be you can just call `set_source_info` without checking for null.
it's even worse than that actually. If it's null and you don't set it, isn't there a chance of a "stale" / incorrect source info (from a different primitive call) being displayed?

include/bm/bm_sim/actions.h
```diff
@@ -543,7 +547,7 @@ class ActionPrimitive_ {
   // (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60056).
   static thread_local Packet *pkt;
   static thread_local PHV *phv;
-
+  SourceInfo *srcInfo = nullptr;
```
antoninbas: this should be renamed as it is misleading in its current form. This is not the source info for the action primitive, it is the source info for the action primitive *call*. The variable naming scheme follows the wrong convention as well. So maybe something like `call_source_info`?

tests/test_assert_assume.cpp
```diff
@@ -0,0 +1,127 @@
+/* Copyright 2019-Barefoot Networks, Inc.
```
antoninbas: one more thing: please fix this copyright, unless I'm mistaken and you actually work at Barefoot

targets/psa_switch/primitives.cpp
enisah: I have read that l2_switch and simple_router don't work with P416 since last year and that is why I put it separated previously. Now, I move them in core/primitives.cpp

tests/test_assert_assume.cpp
enisah: Primitives in core/primitives.cpp do abort() when condition is false, so it made my test to fail. That is why I needed these primitives to mark that assert/assume was triggered.

antoninbas: Did you try this: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-to-write-a-death-test

enisah: Thank you for the suggestion, I was not familiar with this. I will definitely  try this one.

src/bm_sim/actions.cpp
enisah: You are right, it seems that can just call set_source_info. I have made a change according to your comment.

jafingerhut:
Antonin should of course be happy with any changes made for this PR as well, but I wanted to mention that I ran some tests with this modified version of behavioral-model, plus the corresponding changes to p4c here: https://github.com/p4lang/p4c/pull/1859

Things seem to be working as I would expect with this combination of changes, so I hope @enisah you and Antonin can find the appropriate changes to the proposed PR to make this acceptable.  I am happy to redo my little bit of testing again if further changes are made to either or both of the PRs before they are ready for merging.

jafingerhut:
@antoninbas Just a quick note to let you know that @enisah says he force pushed a commit to this PR, so even though only 1 commit shows, it is a new commit, not additional commit on top of the previous one you reviewed.

I will try to find some time to test these newer changes fairly soon.

jafingerhut:
I built the latest versions of this PR and the corresponding one for p4c here: https://github.com/p4lang/p4c/pull/1859 as of 2019-May-28, and things appear to be working as I would expect.

tests/test_assert_assume.cpp
```diff
@@ -0,0 +1,95 @@
...
+    EXPECT_DEATH(testActionFnEntry(pkt.get()), "");
+  }
+};
+
+TEST_F(AssertAssume_DeathTest, AssumeBoolConstError) {
+  verify_test(false, true);
+}
+
+TEST_F(AssertAssume_DeathTest, AssertBoolConstError) {
+  verify_test(true, true);
```
antoninbas: I think a value-parameterized test (https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#value-parameterized-tests) with 4 combinations would have been neat: ("assert", true), ("assert", false), ("assume", true), ("assume", false)
But that's more of a "nice-to-have", so I won't block this PR because of this.

The instantiation could look like this:
```
INSTANTIATE_TEST_CASE_P(
  AssertAssumeDeathTest,
  AssertAssume_DeathTest,
  Combine(Values("assert", "assume"), Bool()));
```

enisah: I tried to follow your suggestions, with this kind of tests and earlier suggestions not to duplicate the code with new primitives in tests/primitives.cpp. If the value is true in primitives it would not be death test anymore and it will fail. If I try to catch it in any other way and stop the failure of test, I had to change the code or even duplicate it again. Due to neat of the code, I supposed it is better to leave with these two combinations of values, if you agree.

antoninbas: No problem

PR #763, opened 2019-04-20
rm remaining v1model std metadata references from psa_switch

peteli3:
Changelog:
- removes all references to `standard_metadata`, `intrinsic_metadata`, `queueing_metadata` and their containing functions (if ONLY used because of v1model std metadata)
- forces all new `psa_*_metadata` headers to be arith (they are all either uint or p4 enums), this appeases some stf tests that were previously failing to extract uint from Field object associated with some psa standard metadata.
- ingress_thread should drop packets if `psa_ingress_output_metadata.drop` is true
- we no longer use `packet_copy` (since we didn't see a need for it at the moment, to get the bare bones psa_switch passing some tests)

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -152,10 +152,10 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
   PHV *phv = packet->get_phv();
   phv->reset_metadata();
 
-  // TODO
-  phv->get_field("psa_ingress_parser_input_metadata.ingress_port").set(port_num);
+  // TODO use appropriate enum member from JSON
   phv->get_field("psa_ingress_parser_input_metadata.packet_path").set(PKT_INSTANCE_TYPE_NORMAL);
-
+  phv->get_field("psa_ingress_parser_input_metadata.ingress_port").set(port_num);
+  
   // using packet register 0 to store length, this register will be updated for
   // each add_header / remove_header primitive call
   packet->set_register(PACKET_LENGTH_REG_IDX, len);
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -256,14 +256,7 @@ PsaSwitch::get_ts() const {
 void
 PsaSwitch::enqueue(port_t egress_port, std::unique_ptr<Packet> &&packet) {
     packet->set_egress_port(egress_port);
-
     PHV *phv = packet->get_phv();
-
-    if (with_queueing_metadata) {
-      phv->get_field("queueing_metadata.enq_timestamp").set(get_ts().count());
-      phv->get_field("queueing_metadata.enq_qdepth")
-          .set(egress_buffers.size(egress_port));
-    }
 
 #ifdef SSWITCH_PRIORITY_QUEUEING_ON
     size_t priority = phv->has_field(SSWITCH_PRIORITY_QUEUEING_SRC) ?
@@ -290,7 +283,6 @@ PsaSwitch::copy_field_list_and_set_type(
   phv_copy->reset_metadata();
   FieldList *field_list = this->get_field_list(field_list_id);
   field_list->copy_fields_between_phvs(phv_copy, packet->get_phv());
-  phv_copy->get_field("standard_metadata.instance_type").set(copy_type);
 }
 
 void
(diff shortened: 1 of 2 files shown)


targets/psa_switch/psa_switch.cpp
```diff
@@ -121,18 +121,21 @@ PsaSwitch::PsaSwitch(bool enable_swap)
...
   import_primitives();
 }
 
 #define PACKET_LENGTH_REG_IDX 0
 
 int
 PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
-  // this is a good place to call this, because blocking this thread will not
-  // block the processing of existing packet instances, which is a requirement
-  if (do_swap() == 0) {
```
antoninbas: you shouldn't remove the `do_swap` call, or program swap (e.g. with P4Runtime) won't be supported

targets/psa_switch/psa_switch.cpp
```diff
@@ -141,23 +144,16 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
                                bm::PacketBuffer(len + 512, buffer, len));
 
   BMELOG(packet_in, *packet);
-
   PHV *phv = packet->get_phv();
-  // many current P4 programs assume this
-  // it is also part of the original P4 spec
   phv->reset_metadata();
```
antoninbas: I think if you remove the comment, you should also remove the call to `reset_metadata`. Plus, according to the PSA spec:

> The PSA architecture does not mandate initialization of user-defined metadata to known values as given as input to the ingress parser.

antoninbas: Maybe @jafingerhut can comment on this

targets/psa_switch/psa_switch.cpp
```diff
@@ -141,23 +144,16 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
...
 
-  // setting standard metadata
-
-  phv->get_field("psa_ingress_parser_input_metadata.ingress_port").set(port_num);
   // TODO
-  // Use appropriate enum member value in JSON
```
antoninbas: why are we removing this comment, the TODO is still valid

peteli3: my bad, there's work for this in a separate branch that's not quite done yet, perhaps this should not have been removed just yet.

targets/psa_switch/psa_switch.cpp
peteli3: I updated this comment to include this excerpt, originally (and eagerly) removed the comment thinking that it's implied

targets/psa_switch/psa_switch.cpp
peteli3: fixed

targets/psa_switch/psa_switch.cpp
jafingerhut: So the PSA architecture does not require initialization of user-defined metadata.

A particular PSA _implementation_ still can do so, if the implementer wants to do that.

Some tradeoffs:

If all user-defined metadata is initialized to a well known value, it makes the P4 programs more predictable in their behavior.  However, if they want to catch portability bugs in their programs where they are using uninitialized values, such an implementation will not help them much.  However, even assigning some kind of "random" values intentionally as initial values for user-defined metadata is not guaranteed to catch such bugs -- you really want a static analysis tool for that purpose, which do exist (e.g. p4v, p4pktgen).

Conclusion:

I think it is reasonable for psa_switch to initialize all user-defined metadata to fixed values, e.g. 0 for `bit<W>` and `int<W>` types, and whatever the bit pattern 0 in psa_switch memory means for `enum` types (maybe no legal value, which is OK for an implementation to do).  It would be good to have a comment that while this is not required behavior for a PSA implementation, it is allowed behavior, and someone may want to consider changing that behavior in psa_switch later.

antoninbas:
It seems that Peter has added a comment, so we should be good to merge

jafingerhut:
Cool.  I can look into creating a small PR for p4c to update the XFAIL tests that are probably going to start succeeding, causing the p4c Travis builds to fail, later today.

In general, if it is easy, it saves a little bit of churn if the order of enhancements goes in this order, to avoid breaking the p4c build:

+ enhancements to behavioral-model
+ add test case to p4c that exercises the new behavioral-model capabilities

I probably suggested violating that order earlier by adding tests to p4c that needed XFAIL, so my bad there if I did.  It isn't a huge deal -- mainly an annoyance to temporarily break the p4c builds of PRs created by others for a day or so

PR #764, opened 2019-04-22
PSA Counters (Work In Progress)

derekso1:
Changelog:

**runtime_CLI**
- I added support for writing to a psa counter/ss counter array in the runtime_CLI

**psa_switch**
- p4 programs can now define and use counters in psa_switch
- added unit tests for the newly created psa counters

**testing**
It seems like STF does not support counter reads / writes so I do not have an stf test provided to show that the code is working. I made some quick changes to the runtime_CLI for this reason. The runtime_CLI can now read and write from p4 defined counters. The command to write to a counter is
`counter_write <counter_name> <index> <packets> <bytes>`  I have manually verified that counters are read, written, and incremented correctly.

Mainly looking for feedback as to whether I made the right design choices. I tried my best to not interfere with anything simple switch related yet at the same time try to keep psa and simple switch counters separate from each other. I have left a review of the code myself detailing the choices I have made.


include/bm/bm_sim/extern.h
```diff
@@ -65,6 +65,16 @@ class ExternFactoryMap {
            #extern_name,                                                \
            [](){ return std::unique_ptr<ExternType>(new extern__()); });
 
+#define BM_REGISTER_EXTERN_TYPE(extern_name)                               \
```
derekso1: Seems redundant since BM_REGISTER_EXTERN already exists, but I created this to suppress a compiler warning where BM_REGISTER_EXTERN has an unused variable.

include/bm/bm_sim/psa_counter.h
```diff
@@ -0,0 +1,87 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
```
derekso1: This is the psa equivalent of a counter_array, extends the extern type. P_Counter is literally identical to the Counter defined in counter.h. I believe @jnfoster said at the end of the day, we would like to not intertwine the implementations of simple switch counters and psa counters (I may have also misunderstood what he said).

src/bm_sim/P4Objects.cpp
```diff
@@ -890,6 +890,7 @@ P4Objects::init_header_union_stacks(const Json::Value &cfg_root,
 
 void
 P4Objects::init_extern_instances(const Json::Value &cfg_root) {
+  BM_REGISTER_EXTERN_TYPE(PSA_Counter);
```
derekso1: It doesn't feel right putting this here, is there a better location?

src/bm_sim/context.cpp
```diff
@@ -527,6 +527,20 @@ Context::read_counters(const std::string &counter_name, size_t idx,
   return (*counter_array)[idx].query_counter(bytes, packets);
 }
```
derekso1: Here, I decided to implement two new methods in context that allow the control plane to read and write to psa_counters since changing anything here would also affect simple switch.

targets/psa_switch/psa_switch.h
```diff
@@ -79,6 +83,28 @@ class PsaSwitch : public Switch {
   // by default, swapping is off
   explicit PsaSwitch(bool enable_swap = false);
```
derekso1: What I did here might seem hacky, but I did not want to interfere with the runtime interface directly since simple switch will also be affected. Instead I did an override of the read_counters and write_counters functions which would then call the new methods I wrote in context.cpp. This way, the interface for both psa and simple switch stays the same but each switch reads from their own implementation of counter.

antoninbas:
An alternative, which is probably cleaner for the compiler is this:
```
class PSACounter {
  public:
    BM_EXTERN_ATTRIBUTES {
        BM_EXTERN_ATTRIBUTE_ADD(init_count);
        BM_EXTERN_ATTRIBUTE_ADD(type);
    }
    void init() override {
        _counter = std::unique_ptr<bm::CounterArray>(new bm::CounterArray(...));
    }

    count(const Data &index) {  // the PSA count "dataplane" method, the JSON will include calls to this method (see bmv2 JSON documentation)
        _counter->get_counter(index.get<size_t>()).increment(get_packet());
    }
  private:
    std::unique_ptr<bm::CounterArray> _counter;  // does not exist outside of this object
};
BM_REGISTER_EXTERN(PSACounter);
BM_REGISTER_EXTERN_METHOD(PSACounter, count, const Data &);
```

I like this one, it doesn't require a `bm::CounterArray` to be initialized from the JSON, which means extern instances can be treated more uniformly by the compiler backend. You also avoid all the code duplication becvause you encapsulate a `bm::CounterArray` instance. However, it requires overriding `read_counters` in `PSASwitch`, which is fine IMO and which you have already done in this PR.

derekso1:
Ah I understand now, I've always misunderstood that counter arrays were unique to simple switch. Let me reiterate just to make sure I'm on the right track this time. Really what should be happening is that we are creating both counter arrays and PSA counters. The counter array will be wrapped by PSA counters. The CLI runtime doesn't change since it allows runtime to change counter arrays which is our PSA counter. Then the last difference between simple switch and PSA switch would be the primitive count(). Instead of directly counting the counter array like in simple switch, PSA switch will increment the counter array through the new PSA counter wrapper.

antoninbas:
So basically - and I was probably too negative in my very first comment, especially since I think the overriding of `read_counters` in `PSASwitch` is a good idea in the end - the big issue with the current patch is that is assumes the bmv2 core logic knows about PSA objects, when the whole point of the extern indirection is to avoid that and keep architecture-specific constructs separated from the core library (just like for the P4 language).
We can discuss on Slack tomorrow if you want or even have a hangout meeting.

derekso1:
Yes that would be great thanks Antonin, I will close this for now.

antoninbas:
I sent an invite. Feel free to pick a different time (you can edit the invite).
I liked the second approach a lot I described until I realized some tools which consume the JSON would stop working or not work as well (e.g. no more autocomplete for the CLI). I still think it is a good option. We'll discuss the options tomorrow and see what's the best approach of the 2.

jnfoster:
I think I am slightly guilty for causing this confusion. @derekso1 and I discussed on Friday and I misunderstood the intention for these Bmv2 counters. Sorry about that.

PR #765, opened 2019-04-22
Add simple_switch docs for specifying const entries

jafingerhut:


docs/simple_switch.md
```diff
@@ -484,6 +484,124 @@ relative order that entries appear in the P4 program is unimportant in
...
+A restriction on all allowed cases is that `lo`, `hi`, `val`, and `mask` must be
+a legal possible value for the field, i.e. not outside of that field's range of
+values. All of them can be arithmetic expressions containing compile time
+constant values.
+
+|                  | `range`      | `ternary`    | `lpm`        | `exact` |
+| ---------------- | ------------ | ------------ | ------------ | ------- |
+| `lo .. hi`       | yes (Note 1) | no           | no           | no      |
+| `val &&& mask`   | no           | yes (Note 2) | yes (Note 3) | no      |
+| `val`            | yes (Note 4) | yes (Note 5) | yes (Note 5) | yes     |
```
antoninbas: I don't think this case is supported for range.
The code doesn't seem to support this case (https://github.com/p4lang/p4c/blob/master/backends/bmv2/common/control.cpp#L87) and I confirmed it by editing the table-entries-range-bmv2 test case.
```
../testdata/p4_16_samples/table-entries-range-bmv2.p4(65): [--Werror=unsupported] error: 9: Unsupported range key expression
            9: a_with_control_params(22);
            ^
```
I have a fairly recent compiler version but maybe not the latest. Did you run your `t1` example below through the compiler? 

It is a bit confusing because the P4Runtime code that generates a `WriteRequest` message for static const entries does handle that case gracefully (https://github.com/p4lang/p4c/blob/master/control-plane/p4RuntimeSerializer.cpp#L1511), but bmv2 doesn't currently use that message. Instead the entries are embedded in the JSON file, and the code that does that is slightly different from the P4Runtime code. Shouldn't be hard to update the bmv2 backend code to support this as well.

jafingerhut: You are correct.

Before I submitted this issue, I verified with a command like the following that the P4Info 'entries' file can be generated correctly by p4test, but I didn't check with the BMv2 back end, because I assumed that it shouldn't care about the const entries.

Do you know why the p4c BMv2 back end examines the const entries at all?  I didn't think it put anything for those const entries data into the BMv2 JSON file at all, does it?  If it doesn't, what is it doing with those enries?

jafingerhut: Sorry, forgot to give the command that I tested successfully with:
```
$ p4c --target bmv2 --arch v1model --p4runtime-entries-files range.entries.txt range.p4
```
It appears to generate a range.entries.txt file with a correct lo=hi="the one value given in the source code" when you have a `range` field with a single value like `15` given.

antoninbas: Currently bmv2 ignores whatever is generated by `--p4runtime-entries-files`, and as I mentioned above all the entries are embedded in the JSON and bmv2 inserts the entries in their respective tables when the JSON is loaded. This is very similar to how default actions are handled: the information is included in the JSON and bmv2 sets the default action when loading the JSON, as soon as the bmv2 table is constructed.
This system predates `--p4runtime-entries-files` and has some advantages, mainly that it installs static entries even when you are not using P4Runtime (for example simple_switch doesn't support P4Runtime).

jafingerhut: So one way forward here: I can change this PR to indicate that the combination of `range` and `val` is not supported today, and create an Issue for p4c with a test program that uses that combination, in hopes that it will be enhanced later?

antoninbas: I'm fine with this solution but it may be just as little work to fix it directly in p4c (the code can be "copied" from the p4RuntimeSerializer).

jafingerhut: OK, enhancing p4c to support this case, then.  Created PR https://github.com/p4lang/p4c/pull/1883 that should make the enhanced version of the docs true soon.

jafingerhut: That PR has been merged, so this documentation should now be correct.

PR #768, opened 2019-04-23
Change "private" v1model standard_metadata fields to BMv2 packet regi…

jafingerhut:
…sters

Remove the use of the following 4 fields from BMv2 simple_switch and
simple_switch_grpc.  They no longer need to be in the BMv2 JSON file
for such programs, and clone, recirculate, resubmit, and digest
operations are still able to work without them there, because they are
now stored in a BMv2 "packet register", similar to how the packet
length field currently is.

+ clone_spec
+ resubmit_flag
+ recirculate_flag
+ lf_field_list

### DIFF
--- a/include/bm/bm_sim/packet.h
+++ b/include/bm/bm_sim/packet.h
@@ -105,7 +105,7 @@ class Packet final {
   using buffer_state_t = PacketBuffer::state_t;
 
   //! Number of general purpose registers per packet
-  static constexpr size_t nb_registers = 2u;
+  static constexpr size_t nb_registers = 3u;
 
   static constexpr size_t INVALID_ENTRY_INDEX =
       std::numeric_limits<size_t>::max();
--- a/include/bm/bm_sim/packet.h
+++ b/include/bm/bm_sim/packet.h
@@ -105,7 +105,7 @@ class Packet final {
   using buffer_state_t = PacketBuffer::state_t;
 
   //! Number of general purpose registers per packet
-  static constexpr size_t nb_registers = 3u;
+  static constexpr size_t nb_registers = 4u;
 
   static constexpr size_t INVALID_ENTRY_INDEX =
       std::numeric_limits<size_t>::max();
--- a/targets/simple_switch/primitives.cpp
+++ b/targets/simple_switch/primitives.cpp
@@ -226,7 +226,7 @@ class generate_digest : public ActionPrimitive<const Data &, const Data &> {
     // discared receiver for now
     (void) receiver;
     auto &packet = get_packet();
-    RegisterAccess::set_lf_field_list(packet, learn_id.get<uint16_t>());
+    RegisterAccess::set_lf_field_list(&packet, learn_id.get<uint16_t>());
   }
 };
 
@@ -288,9 +288,9 @@ class clone_ingress_pkt_to_egress
     // significant bit to always make the value non-0.  This enables
     // cleanly supporting mirror_session_id == 0, in case that is ever
     // helpful.
-    RegisterAccess::set_clone_mirror_session_id(packet,
+    RegisterAccess::set_clone_mirror_session_id(&packet,
         mirror_session_id.get<uint16_t>() | (1u << 15));
-    RegisterAccess::set_clone_field_list(packet,
+    RegisterAccess::set_clone_field_list(&packet,
         field_list_id.get<uint16_t>());
   }
 };
@@ -302,9 +302,9 @@ class clone_egress_pkt_to_egress
   void operator ()(const Data &mirror_session_id, const Data &field_list_id) {
     auto &packet = get_packet();
     // See clone_ingress_pkt_to_egress for why the arithmetic.
-    RegisterAccess::set_clone_mirror_session_id(packet,
+    RegisterAccess::set_clone_mirror_session_id(&packet,
         mirror_session_id.get<uint16_t>() | (1u << 15));
-    RegisterAccess::set_clone_field_list(packet,
+    RegisterAccess::set_clone_field_list(&packet,
         field_list_id.get<uint16_t>());
   }
 };
@@ -314,7 +314,7 @@ REGISTER_PRIMITIVE(clone_egress_pkt_to_egress);
 class resubmit : public ActionPrimitive<const Data &> {
   void operator ()(const Data &field_list_id) {
     auto &packet = get_packet();
-    RegisterAccess::set_resubmit_flag(packet, field_list_id.get<uint16_t>());
+    RegisterAccess::set_resubmit_flag(&packet, field_list_id.get<uint16_t>());
   }
 };
 
@@ -323,7 +323,8 @@ REGISTER_PRIMITIVE(resubmit);
 class recirculate : public ActionPrimitive<const Data &> {
   void operator ()(const Data &field_list_id) {
     auto &packet = get_packet();
-    RegisterAccess::set_recirculate_flag(packet, field_list_id.get<uint16_t>());
+    RegisterAccess::set_recirculate_flag(&packet,
+                                         field_list_id.get<uint16_t>());
   }
 };
 
--- a/targets/simple_switch/primitives.cpp
+++ b/targets/simple_switch/primitives.cpp
@@ -284,12 +284,14 @@ class clone_ingress_pkt_to_egress
   : public ActionPrimitive<const Data &, const Data &> {
   void operator ()(const Data &mirror_session_id, const Data &field_list_id) {
     auto &packet = get_packet();
-    // Assume mirror_session_id < (1u << 15) so that we can use most
-    // significant bit to always make the value non-0.  This enables
-    // cleanly supporting mirror_session_id == 0, in case that is ever
-    // helpful.
+    // We limit mirror_session_id values to small enough values that
+    // we can use one of the bit positions as a "clone was performed"
+    // indicator, making mirror_seesion_id stored here always non-0 if
+    // a clone was done.  This enables cleanly supporting
+    // mirror_session_id == 0, in case that is ever helpful.
     RegisterAccess::set_clone_mirror_session_id(&packet,
-        mirror_session_id.get<uint16_t>() | (1u << 15));
+        mirror_session_id.get<uint16_t>() |
+        RegisterAccess::MIRROR_SESSION_ID_VALID_MASK);
     RegisterAccess::set_clone_field_list(&packet,
         field_list_id.get<uint16_t>());
   }
@@ -303,7 +305,8 @@ class clone_egress_pkt_to_egress
     auto &packet = get_packet();
     // See clone_ingress_pkt_to_egress for why the arithmetic.
     RegisterAccess::set_clone_mirror_session_id(&packet,
-        mirror_session_id.get<uint16_t>() | (1u << 15));
+        mirror_session_id.get<uint16_t>() |
+        RegisterAccess::MIRROR_SESSION_ID_VALID_MASK);
     RegisterAccess::set_clone_field_list(&packet,
         field_list_id.get<uint16_t>());
   }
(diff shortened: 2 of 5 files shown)


jafingerhut:
A first draft of modifying simple_switch and simple_switch_grpc so that they no longer access clone_spec, lf_field_list, resubmit_flag, and recirculate_flag in the BMv2 JSON file, and those fields thus need not be in the JSON file at all in order for the corresponding features to work.

If any of this looks like bad C++, it probably is.  Let me know what I've done badly, and hopefully I will learn.

jafingerhut:
Forgot to mention: I have run all of the p4c automated tests on my local dev machine with this modified version of simple_switch, and except for clone-bmv2.p4 they all pass.  That test appears fairly old, and has no STF test, so could probably be removed.

targets/simple_switch/register_access.h
```diff
@@ -0,0 +1,108 @@
...
+
+/*
+ * Andy Fingerhut (jafinger@cisco.com)
+ *
+ */
+
+#ifndef SIMPLE_SWITCH_REGISTER_ACCESS_H_
+#define SIMPLE_SWITCH_REGISTER_ACCESS_H_
+
+#define PACKET_LENGTH_REG_IDX             0
```
antoninbas: could we use `static constexpr` variables inside the `RegisterAccess` class instead of all these macros, they will be strongly typed

targets/simple_switch/register_access.h
```diff
@@ -0,0 +1,108 @@
...
+#define RESUBMIT_FLAG_SHIFT               48
+
+#define RECIRCULATE_FLAG_REG_IDX          2
+#define RECIRCULATE_FLAG_MASK             0x000000000000ffff
+#define RECIRCULATE_FLAG_SHIFT            0
+
+
+class RegisterAccess {
+ public:
+    static void clear_all(Packet &pkt) {
```
antoninbas: let's use a `Packet` pointer for all the functions which modify the packet.

targets/simple_switch/register_access.h
```diff
@@ -0,0 +1,108 @@
...
+        // except do not clear packet length
+        pkt.set_register(1, 0);
+        pkt.set_register(2, 0);
+    }
+    static uint16_t get_clone_mirror_session_id(Packet &pkt) {
+        uint64_t rv = pkt.get_register(CLONE_MIRROR_SESSION_ID_REG_IDX);
+        return (uint16_t) ((rv & CLONE_MIRROR_SESSION_ID_MASK) >>
```
antoninbas: use `static_cast<T>` for all numerical casts

targets/simple_switch/simple_switch.cpp
```diff
@@ -501,17 +501,14 @@ SimpleSwitch::ingress_thread() {
     Field &f_egress_spec = phv->get_field("standard_metadata.egress_spec");
     port_t egress_spec = f_egress_spec.get_uint();
 
-    Field &f_clone_spec = phv->get_field("standard_metadata.clone_spec");
-    unsigned int clone_spec = f_clone_spec.get_uint();
+    unsigned int clone_mirror_session_id =
```
antoninbas: do those need to be `unsigned int` or can we just use `auto` and use whichever type is returned by the register access proxy

include/bm/bm_sim/packet.h
```diff
@@ -105,7 +105,7 @@ class Packet final {
   using buffer_state_t = PacketBuffer::state_t;
 
   //! Number of general purpose registers per packet
-  static constexpr size_t nb_registers = 2u;
+  static constexpr size_t nb_registers = 3u;
```
antoninbas: could we make it 4?
a very good reason: I don't like 3
a not so good reason: reserve one register for future use

targets/simple_switch/register_access.h
jafingerhut: yep

targets/simple_switch/register_access.h
jafingerhut: Here is where my C++ ignorance shows again.  When I use packet pointers, there are places where I attempt to call one of these functions that give an error message I'm not sure of the best way to address. Example below.  What would you recommend for this?

I change the definition of  `get_clone_mirror_session_id` to this in register_access.h, to use a pointer to Packet instead of a reference:
```
    static uint16_t get_clone_mirror_session_id(Packet *pkt) {
        uint64_t rv = pkt->get_register(CLONE_MIRROR_SESSION_ID_REG_IDX);
        return (uint16_t) ((rv & CLONE_MIRROR_SESSION_ID_MASK) >>
                           CLONE_MIRROR_SESSION_ID_SHIFT);
    }
```

In simple_switch.cpp:
```
void
SimpleSwitch::ingress_thread() {

// ... later ...

    unsigned int clone_mirror_session_id =
        RegisterAccess::get_clone_mirror_session_id(packet);
```
when compiling I get this error message that I would love recomendations for fixing:
```
simple_switch.cpp: In member function ‘void SimpleSwitch::ingress_thread()’:
simple_switch.cpp:505:59: error: no matching function for call to ‘RegisterAccess::get_clone_mirror_session_id(std::unique_ptr<bm::Packet>&)’
         RegisterAccess::get_clone_mirror_session_id(packet);
                                                           ^
In file included from simple_switch.cpp:36:0:
register_access.h:51:21: note: candidate: static uint16_t RegisterAccess::get_clone_mirror_session_id(bm::Packet*)
     static uint16_t get_clone_mirror_session_id(Packet *pkt) {
                     ^
```

include/bm/bm_sim/packet.h
jafingerhut: no problem

targets/simple_switch/register_access.h
jafingerhut: done

targets/simple_switch/simple_switch.cpp
jafingerhut: Will try replacing all of those types with `auto` and see how it goes.

targets/simple_switch/simple_switch.cpp
```diff
@@ -520,18 +517,20 @@ SimpleSwitch::ingress_thread() {
     }
 
     // INGRESS CLONING
-    if (clone_spec) {
+    if (clone_mirror_session_id) {
```
jafingerhut: Antonin, one nitpick question for you I wasn't sure about.  The `if` condition here is actually slightly different than the original.  To make it completely equivalent to the original condition, I could replace it with: `if (clone_mirror_session_id || clone_field_list)`, but I was guessing that perhaps clone mirror session ids are restricted to be non-0.  Is that true?

jafingerhut: And if they can be 0, then it was probably slightly buggy to begin with, if P4_16 clone() was implemented by using a field_list_id of 0.

antoninbas: Actually that's an important question :)
I believe the assumption is that the mirror session id is non-0. Even though the P4Runtime spec states that 0 is a valid value, the reference P4Runtime implementation will currently not let you use 0 (https://github.com/p4lang/PI/blob/master/proto/frontend/src/pre_clone_mgr.h#L69). So maybe documenting this (in the code / the simple_switch documentation) is important.

jafingerhut: If you want, I can with these changes add another "architecture private" field that is a 1-bit boolean flag remembering whether a clone operation was ever performed, and use that in the `if` conditions.  Then it will cleanly support a clone/mirror session id of 0.

antoninbas: Up to you, you can also use the msb of the 16-bit register
personally I don't think it's a big deal since there is no regression

jafingerhut: I will go for using msb of 16-bit value in next commit.  Seems worth cleanly supporting clone/mirror session id of 0 as a small enhancement.  Then no need to document that 0 is not supported :)

targets/simple_switch/register_access.h
jafingerhut: With latest commit, this comment is the only one that I have not addressed yet.  Open to suggestions on this one on best way to address it.

antoninbas: When packet is a `std::unique_ptr`, and you call a function that doesn't take ownership of the pointer, but mutates the pointee, just pass a raw pointer to the function. So in your case, the simple_switch.cpp code should be:
```
auto clone_mirror_session_id =
        RegisterAccess::get_clone_mirror_session_id(packet.get());
```
https://en.cppreference.com/w/cpp/memory/unique_ptr/get

jafingerhut: Done.  Thanks for the coding advice!

targets/simple_switch/primitives.cpp
```diff
@@ -274,47 +280,51 @@ class copy_header : public ActionPrimitive<Header &, const Header &> {
...
-   field list id to copy (16 MSB) */
 class clone_ingress_pkt_to_egress
   : public ActionPrimitive<const Data &, const Data &> {
-  void operator ()(const Data &clone_spec, const Data &field_list_id) {
-    Field &f_clone_spec = get_field("standard_metadata.clone_spec");
-    f_clone_spec.shift_left(field_list_id, 16);
-    f_clone_spec.add(f_clone_spec, clone_spec);
+  void operator ()(const Data &mirror_session_id, const Data &field_list_id) {
+    auto &packet = get_packet();
+    // Assume mirror_session_id < (1u << 15) so that we can use most
```
antoninbas: Could we enforce this when the control-plane creates a session here: https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L78, i.e. log an error and return `false` is the session id is `>= (1 << 15)`?

targets/simple_switch/register_access.h
```diff
@@ -0,0 +1,110 @@
...
+    static uint16_t get_clone_mirror_session_id(Packet *pkt) {
+        uint64_t rv = pkt->get_register(CLONE_MIRROR_SESSION_ID_REG_IDX);
+        return static_cast<uint16_t>((rv & CLONE_MIRROR_SESSION_ID_MASK) >>
+                                     CLONE_MIRROR_SESSION_ID_SHIFT);
+    }
+    static void set_clone_mirror_session_id(Packet *pkt,
+                                            uint16_t mirror_session_id) {
+        uint64_t rv = pkt->get_register(CLONE_MIRROR_SESSION_ID_REG_IDX);
+        rv = ((rv & ~CLONE_MIRROR_SESSION_ID_MASK) |
+              (((uint64_t) mirror_session_id) <<
```
antoninbas: There are still some C-style casts, I'm actually surprised the style checker is not complaining.

targets/simple_switch/primitives.cpp
jafingerhut: updated this

targets/simple_switch/register_access.h
jafingerhut: Fixed the ones I found in this file.  Don't know anything about the style checker behavior.

targets/simple_switch/simple_switch.cpp
```diff
@@ -75,13 +76,23 @@ class SimpleSwitch::MirroringSessions {
   bool add_session(mirror_id_t mirror_id,
                    const MirroringSessionConfig &config) {
     Lock lock(mutex);
-    sessions_map[mirror_id] = config;
-    return true;
+    if (0 <= mirror_id && mirror_id <= RegisterAccess::MAX_MIRROR_SESSION_ID) {
```
antoninbas: you must be the kind of person that uses `100[array]` for array access :)

jafingerhut: Only in an obfuscated code contest :)  But I'm missing what about that code change was odd enough to make you think that?

antoninbas: I think it's just a mater of habit, I never write `0 <= x` instead of `x >= 0`, so my brain was frozen for a second.

PR #769, opened 2019-04-25
psa switch should invalidate all headers for pkts entering egress

peteli3:
@jafingerhut 
- As we discussed, pkt headers should be invalidated before egress pipeline starts.
- Also confirmed locally that this change causes the `psa-unicast-or-drop` test to pass (so we should remove that from xfail)

Appending to changelog:
- moved multicast logic to be after ingress deparsing, as it should be
- added or moved `packet->reset_exit()` calls to happen right after `[ingress | egress]_mau->apply()`
- removed some redundant "field exists" checks

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -308,16 +308,22 @@ PsaSwitch::ingress_thread() {
     input_buffer.pop_back(&packet);
     if (packet == nullptr) break;
 
-    Parser *parser = this->get_parser("ingress_parser");
-    Pipeline *ingress_mau = this->get_pipeline("ingress");
-    phv = packet->get_phv();
-
     port_t ingress_port = packet->get_ingress_port();
     (void) ingress_port;
     BMLOG_DEBUG_PKT(*packet, "Processing packet received on port {}",
                     ingress_port);
 
+    phv = packet->get_phv();
+
+    Parser *parser = this->get_parser("ingress_parser");
     parser->parse(packet.get());
+
+    if (phv->has_field("psa_ingress_output_metadata.drop")) {
+      phv->get_field("psa_ingress_output_metadata.drop")
+        .set(true);
+    }
+
+    Pipeline *ingress_mau = this->get_pipeline("ingress");
     ingress_mau->apply(packet.get());
 
     // prioritize dropping if marked as such - do not move below other checks
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -332,12 +332,6 @@ PsaSwitch::ingress_thread() {
       continue;
     }
 
-    Field &f_egress_spec = phv->get_field("psa_ingress_output_metadata.egress_port");
-    port_t egress_spec = f_egress_spec.get_uint();
-    port_t egress_port = egress_spec;
-    egress_port = egress_spec;
-    BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
-
     Deparser *deparser = this->get_deparser("ingress_deparser");
     deparser->deparse(packet.get());
 
@@ -357,6 +351,12 @@ PsaSwitch::ingress_thread() {
       }
       continue;
     }
+
+    Field &f_egress_spec = phv->get_field("psa_ingress_output_metadata.egress_port");
+    port_t egress_spec = f_egress_spec.get_uint();
+    port_t egress_port = egress_spec;
+    egress_port = egress_spec;
+    BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
 
     enqueue(egress_port, std::move(packet));
   }
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -318,51 +318,45 @@ PsaSwitch::ingress_thread() {
     Parser *parser = this->get_parser("ingress_parser");
     parser->parse(packet.get());
 
-    if (phv->has_field("psa_ingress_output_metadata.drop")) {
-      phv->get_field("psa_ingress_output_metadata.drop")
-        .set(true);
-    }
+    // set default std metadata values
+    phv->get_field("psa_ingress_output_metadata.drop").set(true);
 
     Pipeline *ingress_mau = this->get_pipeline("ingress");
     ingress_mau->apply(packet.get());
+    packet->reset_exit();
 
     // prioritize dropping if marked as such - do not move below other checks
-    if (phv->has_field("psa_ingress_output_metadata.drop")) {
-      Field &f_drop = phv->get_field("psa_ingress_output_metadata.drop");
-      if (f_drop.get_int()) {
-        continue;
-      }
+    Field &f_drop = phv->get_field("psa_ingress_output_metadata.drop");
+    if (f_drop.get_int()) {
+      continue;
     }
 
-    // handling multicast
-    unsigned int mgid = 0u;
-    if (phv->has_field("psa_ingress_output_metadata.multicast_group")) {
-      Field &f_mgid = phv->get_field("psa_ingress_output_metadata.multicast_group");
-      mgid = f_mgid.get_uint();
-
-      if(mgid != 0){
-        const auto pre_out = pre->replicate({mgid});
-        auto packet_size = packet->get_register(PACKET_LENGTH_REG_IDX);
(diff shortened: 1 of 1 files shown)


jafingerhut:
Looks reasonable to me.  If you want to kill two birds with one stone (or two XFAILs with one PR, you could also add a change to this PR to initialize "psa_ingress_output_metadata_t.drop" to true (probably integer value 1 in psa_switch.cpp), and that should enable the psa-drop-all-bmv2.p4 test to pass, too.

peteli3:
> Looks reasonable to me. If you want to kill two birds with one stone (or two XFAILs with one PR, you could also add a change to this PR to initialize "psa_ingress_output_metadata_t.drop" to true (probably integer value 1 in psa_switch.cpp), and that should enable the psa-drop-all-bmv2.p4 test to pass, too.

I will add that to this PR. Originally wanted to hold off since it's in a branch of one of our teammates and I wanted to avoid merge conflicts for when she submits the PR. However we probably need to do some version control cleaning on her branch anyway - so I'll add it here. Thanks for bringing it up again!

targets/psa_switch/psa_switch.cpp
```diff
@@ -308,16 +308,22 @@ PsaSwitch::ingress_thread() {
...
     (void) ingress_port;
     BMLOG_DEBUG_PKT(*packet, "Processing packet received on port {}",
                     ingress_port);
 
+    phv = packet->get_phv();
+
+    Parser *parser = this->get_parser("ingress_parser");
     parser->parse(packet.get());
+
+    if (phv->has_field("psa_ingress_output_metadata.drop")) {
```
jafingerhut: You should be able to skip all of the has_field() checks for fields that were verified that they existed by an add_required_field() call in the PsaSwitch constructor.  If the field did not exist in the BMv2 JSON file, then an error would have occurred then, so you know it is guaranteed to exist when processing each packet.

targets/psa_switch/psa_switch.cpp
```diff
@@ -308,16 +308,22 @@ PsaSwitch::ingress_thread() {
...
+      phv->get_field("psa_ingress_output_metadata.drop")
+        .set(true);
+    }
+
+    Pipeline *ingress_mau = this->get_pipeline("ingress");
     ingress_mau->apply(packet.get());
```
jafingerhut: I believe the reset_exit() call below should be done just after ingress_mau->apply().  As the code is now, that is never being done for multicast packets.

Also the deparser should be done before the multicast packet handling.  Right now that is never being done for multicast packets, either.

peteli3: Yes, thanks for the catch... didn't even realize we had allowed mcast packets to skip deparsing. Will fix this.

peteli3: @jafingerhut does this mean `packet->reset_exit()` also needs to be called after the egress pipeline in the event it is recirculated? I saw a comment in packet.cpp that says

> Do not forget to call reset_exit() at the end of the pipeline, if you intend the packet to go through another pipeline
 afterwards; otherwise this other pipeline will just be skipped as well.

We are somehow passing recirc tests even without having done this in egress. I'll investigate into what it does - but am also wondering if you have insight on whether it's needed after `egress_mau->apply()`?

jafingerhut: I was wondering the same thing last night myself while reviewing what reset_exit did and looking at the corresponding code for simple_switch, which also does not have a reset_exit call after egress_mau processing.

@antoninbas Do you know whether that could cause a bug for recirculated packets that make an exit call in egress control processing before recirculating?  Or is the fact that recirculated packets are implemented by copying the packet data structure somehow avoiding a bug there?

It certainly seems cleaner to do reset_exit() just after egress control processing, just in case.

jafingerhut: Not that I think that forgetting to do reset_exit() will only cause a problem if the user's program invoked an 'exit' statement in their program, and then the packet was recirculated, and there is probably no existing test case that covers this.  Perhaps there should be....

antoninbas: **recirculate is now implemented without a packet copy** (for psa switch), as it should be, so you definitely need to call `reset_exit`
if you write the test you are describing, I expect it to fail, the ingress pipe will be skipped by the recirculated packet

peteli3: addressed all comments/test suggestions in my most recent push.

targets/psa_switch/psa_switch.cpp
peteli3: Agreed. Fixed all the instances of this that I saw while making the other changes.

targets/psa_switch/psa_switch.cpp
```diff
@@ -308,55 +308,55 @@ PsaSwitch::ingress_thread() {
...
     BMLOG_DEBUG_PKT(*packet, "Processing packet received on port {}",
                     ingress_port);
 
+    phv = packet->get_phv();
+
+    Parser *parser = this->get_parser("ingress_parser");
     parser->parse(packet.get());
+
+    // set default std metadata values
+    phv->get_field("psa_ingress_output_metadata.drop").set(true);
```
antoninbas: you use a boolean here but below you do a `get_int` which is inconsistent

targets/psa_switch/psa_switch.cpp
```diff
@@ -308,55 +308,55 @@ PsaSwitch::ingress_thread() {
...
-          auto egress_port = out.egress_port;
-          std::unique_ptr<Packet> packet_copy = packet->clone_with_phv_ptr();
-          packet_copy->set_register(PACKET_LENGTH_REG_IDX, packet_size);
-          enqueue(egress_port, std::move(packet_copy));
-        }
-        continue;
-      }
+    Field &f_drop = phv->get_field("psa_ingress_output_metadata.drop");
+    if (f_drop.get_int()) {
+      continue;
```
antoninbas: a debug log message would be nice here no?

targets/psa_switch/psa_switch.cpp
peteli3: fixed

targets/psa_switch/psa_switch.cpp
peteli3: yepp, agreed, fixed in latest push

targets/psa_switch/psa_switch.cpp
```diff
@@ -308,55 +308,56 @@ PsaSwitch::ingress_thread() {
...
     }
 
-    packet->reset_exit();
     Field &f_egress_spec = phv->get_field("psa_ingress_output_metadata.egress_port");
     port_t egress_spec = f_egress_spec.get_uint();
     port_t egress_port = egress_spec;
-
     egress_port = egress_spec;
     BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
```
jafingerhut: The 5 lines of code above starting with get_field on egress_port, down to loggin the egress_port, are all specific to unicast packets, and should probably be done later after the mgid != 0 check is false, just before the enqueue call.

peteli3: Fixed! Thanks for pointing this out.

targets/psa_switch/psa_switch.cpp
```diff
@@ -358,6 +352,12 @@ PsaSwitch::ingress_thread() {
       continue;
     }
 
+    Field &f_egress_spec = phv->get_field("psa_ingress_output_metadata.egress_port");
+    port_t egress_spec = f_egress_spec.get_uint();
```
antoninbas: what's the point of the `egress_spec` variable here?

jafingerhut: I suspect that this code is partially a remnant of simple_switch code for handling unicast.

There eventually _should_ be something here for PSA where the unicast egress_port is copied from psa_ingress_output_metadata.egress_port into these three other places, before egress processing begins:

+ psa_egress_parser_input_metadata.egress_port
+ psa_egress_input_metadata.egress_port
+ psa_egress_deparser_input_metadata.egress_port

That does not have to be in this PR, necessarily.

peteli3: It's a remnant of simple_switch. Seems either way we can probably do without the unused "temp var"; so I'll push that as well. Sorry!! Forgetting to remove this was an oversight on my part. It's a very quick fix.

The passing that Andy mentions here will be in a separate PR.

PR #772, opened 2019-04-27
initialize meta data and meta data passing

Tian-Ren:


targets/psa_switch/psa_switch.cpp
```diff
@@ -316,8 +316,32 @@ PsaSwitch::ingress_thread() {
     Parser *parser = this->get_parser("ingress_parser");
     parser->parse(packet.get());
 
-    // set default std metadata values
+    // passing metadata
+    if (phv->has_field("psa_ingress_input_metadata.ingress_port")) {
```
jafingerhut: All of these has_field calls are guaranteed to return true, and can thus be removed.

antoninbas: I think she removed them already

targets/psa_switch/psa_switch.cpp
```diff
@@ -316,8 +316,24 @@ PsaSwitch::ingress_thread() {
     Parser *parser = this->get_parser("ingress_parser");
     parser->parse(packet.get());
 
-    // set default std metadata values
+    // passing metadata
```
antoninbas: that's a bit of a succinct comment...

targets/psa_switch/psa_switch.cpp
```diff
@@ -316,8 +316,24 @@ PsaSwitch::ingress_thread() {
...
+    phv->get_field("psa_ingress_input_metadata.ingress_port").set(ingress_port);
+
+    phv->get_field("psa_ingress_input_metadata.packet_path").set(
+        phv->get_field("psa_ingress_parser_input_metadata.packet_path"));
+
+    phv->get_field("psa_ingress_input_metadata.parser_error").set(
+        packet->get_error_code().get());
+
+    phv->get_field("psa_ingress_input_metadata.ingress_timestamp")
+        .set(get_ts().count());
```
jafingerhut: The PSA spec actually recommends setting this timestamp to the time that the packet began the ingress parser, so perhaps a little bit better to save the value of `get_ts().count` in a local variable just before parsing, then assign this value here.

jafingerhut:
@Tian-Ren It looks like the changes for timestamp you made are not compiling correctly.  Do you see this when attempting to compile these changes on your local machine, too?  If not, look at the Travis build logs for errors near the end: https://travis-ci.org/p4lang/behavioral-model/jobs/525442315

targets/psa_switch/psa_switch.cpp
```diff
@@ -385,11 +395,22 @@ PsaSwitch::egress_thread(size_t worker_id) {
     Parser *parser = this->get_parser("egress_parser");
     parser->parse(packet.get());
 
+    phv->get_field("psa_egress_input_metadata.egress_port").set(
+        phv->get_field("psa_egress_parser_input_metadata.egress_port"));
```
jafingerhut: There is no need to change this PR for this purpose (unless you want to add a "TODO" comment to check this later), but while this is correct for unicast packets sent from ingress to egress, there might be other code needed for handling multicast packets, or cloned packets.

jafingerhut: I think I misplaced this comment.  Copying this value from EgressParser to Egress is always correct, as long as it was correct at the beginning of the EgressParser, which is where it might be useful to have different code for different packet paths.  Not sure yet.

jafingerhut:
@antoninbas I think the only thing "wrong" with the automated test results is that the code coverage is a teensy bit lower.  Good to merge?

targets/psa_switch/psa_switch.cpp
antoninbas: I think Andy is right, it seems that "psa_egress_parser_input_metadata.egress_port" is never set. Shouldn't there be a `phv->get_field("psa_egress_parser_input_metadata.egress_port").set(port)` before the egress parser?

antoninbas: Right now it seems that we are copying from a "uninitialized" metadata field, which means that its value is likely to be always 0

jafingerhut: Would it maybe be OK to merge this in, and create an issue to create some STF tests that pass only if all PSA standard metadata fields are being filled in correctly?  Such a P4 PSA program and STF tests wouldn't be quick to write, and should probably be spread over a few different programs rather than only one, but it would be good to have such tests, in any case.  I can write one that only passes if egress_port is filled in correctly, and it should catch any missing cases of initializing this.

antoninbas: I'm not sure I understand the rationale for not fixing this PR before merging it. It is a one line fix, unless I am missing something. I'm not asking for a test case here, but since we did bring an issue to light I don't know what would justify not fixing it. After all, the name of the PR is "initialize metadata".

Tian-Ren: The egress_port is enqueued at the end of ingress thread. I thought that would populate the port in egress parser. Did I misunderstand that?

jafingerhut:
@Tian-Ren Addressing Antonin's comment seems like it might be just one or two additional lines of code, in the ingress_thread function, near the end just before the call to `enqueue` and just after these lines of code:
```
Field &f_egress_port = phv->get_field("psa_ingress_output_metadata.egress_port");
    port_t egress_port = f_egress_port.get_uint();
```
We just need to add a line something like this:
```
phv->get_field("psa_egress_parser_input_metadata.egress_port").set(egress_port);
```
That will handle the unicast packet case, at least.  We can worry about multicast and other packet paths in a later commit.

targets/psa_switch/psa_switch.cpp
jafingerhut: I think the _only_ thing that causes a PSA standard metadata field to change is an explicit call to the `set(value)` method on `phv->get_field("psa_egress_parser_input_metadata.egress_port")`.

I have a small STF test I have been hacking on this morning, and can confirm that with the current changes as I write this comment, the egress_port value is 0 while the egress code executes.

antoninbas:
The "set" should be at the beginning of the "egress_thread" function, so that it is correct for all packet paths

jafingerhut:
Yeah, sorry, Antonin.  Tian-Ren, Antonin's suggestion is much better than mine.  Somehow I didn't realize that the egress_thread function receives the output port of a packet with the packet.  See simple_switch.cpp code for how it sets standard_metadata.egress_port for a reference.

antoninbas:
Can anyone with access to the branch make that one line change so we can merge this?

jafingerhut:
I do not know how busy Cornell team might be with finals and other coursework, but FYI I am happy to make a quick follow-up PR with the additional 1 line of code desired if you want to go ahead and merge this.

antoninbas:
@jafingerhut Sounds good

Tian-Ren:
@jafingerhut Hey Andy, I am sorry with the late respond. I have been working on all other projects because tomorrow is the last day of this semester. I will add the line and make a follow-up request.

jafingerhut:
No worries that you are busy on other projects.  I have already added the line in a separate PR and Antonin has merged this one plus that other one, here: https://github.com/p4lang/behavioral-model/pull/774

PR #773, opened 2019-04-27
Psa resubmit

kevinye2:
- Copy packet right before ingress processing and if marked as resubmit, resubmit the copy
- Pass some metadata from parser to ingress
- On line 351, we hard-code the `PSA_PacketPath_t.RESUBMIT` value in cpp since we don't have access to the analogous enum in bmv2 just yet


targets/psa_switch/psa_switch.cpp
```diff
@@ -333,6 +344,16 @@ PsaSwitch::ingress_thread() {
     Deparser *deparser = this->get_deparser("ingress_deparser");
     deparser->deparse(packet.get());
 
+    // resubmit - do not move below multicast
+    Field &f_resubmit = phv->get_field("psa_ingress_output_metadata.resubmit");
+    if (f_resubmit.get_int()) {
+      // since we resubmit the copy & toss original, we update the copy's metadata
+      phv_copy->get_field("psa_ingress_parser_input_metadata.packet_path").set(5);
```
jafingerhut: I would recommend defining something like `static constexpr int PACKET_PATH_RESUBMIT = 5;` and using the name, in case the number ends up being used more than once it will be easier to find them all later and update them to the new enum encodings

targets/psa_switch/psa_switch.cpp
```diff
@@ -313,9 +314,19 @@ PsaSwitch::ingress_thread() {
 
     phv = packet->get_phv();
 
+    // PSA wants us to resubmit the instance of the packet BEFORE ingress
+    // processing, so make a copy before ingress touches the packet
+    std::unique_ptr<Packet> packet_copy = packet->clone_no_phv_ptr();
+    PHV *phv_copy = packet_copy->get_phv();
+    phv_copy->reset_metadata();
```
jafingerhut: I do not know the memory alloc/dealloc or performance implications of doing resubmit this way, vs. the way it is done in simple_switch, which _also_ resubmits the packet as it was before ingress processing began.  @antoninbas can comment  more intelligently than I can there.

targets/psa_switch/psa_switch.cpp
jafingerhut: It would also be nice to assign the resubmitted packet's "ingress_timestamp" field the value of the current time here, too.  PSA specifies that it should be updated, not merely copied from the original packet that caused the resubmit.

targets/psa_switch/psa_switch.cpp
peteli3: @jafingerhut ours is the "wasteful/naive" method.

I think @antoninbas used a buffer mechanism for the packet to store a copy of itself (the entire unprased packet) before v1model parsing? But maybe I'm wrong haha since there's some syntax in packet_buffer that I haven't seen before, his comments would be more valuable here. 

Basing my understanding off of the following:

https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L497

https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/packet.h#L181

https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/packet_buffer.h#L103

antoninbas: `buffer_state_t` doesn't copy anything, it just stores an offset. You currently make a copy for all ingress packets, even the ones that won't be resubmitted. It's not more complicated to use `buffer_state_t`, since that code already exists in simple_switch. Ideally, you may be able to avoid any kind of copying for resubmit, like @derekso1 did for recirculate. But if there is no way to do that, making a copy *only* for packets which are resubmitted is fine.

targets/psa_switch/psa_switch.cpp
kevinye2: We addressed both of these in a later commit

targets/psa_switch/psa_switch.cpp
kevinye2: We have to make copies for packets ahead of time unless we know which packets are resubmit ahead of ingress MAU. Currently this is the only solution we could come up with after discussing with Andy. Please let us know if any alternative ideas come up :)

antoninbas: I already pointed you to the simple_switch code: https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L497, which doesn't do a copy.

jafingerhut: Kevin, I believe that the idea of the simple_switch implementation is that even though it does not know before executing ingress code whether the packet will be resubmitted or not, simple_switch does not modify the original packet until after ingress is complete (and perhaps not in any case does it ever modify it?  Not sure about that).  Thus resubmit with the original packet works.  (Do not trust me blindly on that explanation, as I may be slightly off somehow).

peteli3: Hi so I've been working with Kevin on this PR. We have a test that passes only if the original packet has been modified directly after ingress MAU and infinite loops otherwise.

With the simple switch code in place of what we have, we are passing that test. I'll go back and double check but we reiterated over this twice and observed the same behavior which lead us to the comment from above. Would it perhaps be best to get on a call briefly to discuss?

antoninbas: A Packet object is essentially the packet buffer + PHV (set of extracted fields and metadata). The important thing is that for the resubmitted packet, the packet buffer be the same as for the original packet. The PHV will be "initialized" by the parser + by the reset metadata operation (which is really optional for PSA). The parser implementation modifies the packet buffer (the payload pointer is moved every time an header is extracted), which is why the packet buffer state needs to be saved (https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L497) before parsing and restored (https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L582) after the MAU pipeline. This basically requires 2 pointer copies instead of a much more expensive packet copy.

Note that I'm assuming the original packet does not go through the ingress deparser when we resubmit it. I think the PSA packet paths diagram (https://p4.org/p4-spec/docs/PSA-v1.1.0.html#packet-paths) is misleading. Since the resubmitted packet uses the original headers, the deparser is not involved.

targets/psa_switch/psa_switch.cpp
```diff
@@ -333,6 +331,17 @@ PsaSwitch::ingress_thread() {
...
+    // resubmit - do not move below multicast
+    Field &f_resubmit = phv->get_field("psa_ingress_output_metadata.resubmit");
+    if (f_resubmit.get_int()) {
+      // since we resubmit the copy & toss original, we update the copy's metadata
+      phv_copy->get_field("psa_ingress_parser_input_metadata.packet_path").set(PACKET_PATH_RESUBMIT);
+      phv_copy->get_field("psa_ingress_input_metadata.ingress_timestamp").set(get_ts().count());
+      BMLOG_DEBUG_PKT(*packet, "Resubmitting packet");
+      input_buffer.push_front(std::move(packet_copy));
+      continue;
+    }
```
jafingerhut: This resubmit check and code should, I think, be moved above the invocation of the deparser.  As Antonin mentioned, there is no functional need to invoke the deparser for resubmitted packets.  At the very least, it is inefficient and a waste of psa_switch CPU cycles.  It may also be preventing the cheaper simple_switch way of resubmitting the original packet contents from working.

targets/psa_switch/psa_switch.cpp
peteli3: resolved now

peteli3: https://github.com/p4lang/behavioral-model/pull/773/files#diff-5ae8a8504eb56bc8167d4b69469ed7e8R339

targets/psa_switch/psa_switch.cpp
peteli3: also resolved in the latest commit @jafingerhut

targets/psa_switch/psa_switch.cpp
```diff
@@ -313,9 +301,21 @@ PsaSwitch::ingress_thread() {
 
     phv = packet->get_phv();
 
+    /* PSA wants us to resubmit the instance of the packet BEFORE ingress
+       processing, so store the "parsed headers" prior to parsing. Feels a bit
+       hacky though, same as in simple_switch.cpp
```
jafingerhut: The comment 'so store the "parsed headers" prior to parsing' leads me to wonder if there is some confusion here.

The contents of the "parsed headers" prior to parsing should _always_ be the same: all headers invalid.

I do not know the precise effect of packet->save_buffer_state(), and whether what it is doing is only saving the state of the parsed headers, or perhaps is also doing other things.

jafingerhut: Except for that one comment of mine, which is only about the wording of a comment and so not necessary to change for code-functionality reasons, this all looks good to me.  I haven't tested it yet, though.

peteli3: I probably have this wrong. Will look into it to correct the comment

jafingerhut: From an earlier comment by Antonin, who knows this code much better than I: https://github.com/p4lang/behavioral-model/pull/773#discussion_r281438818

it seems that perhaps one of the things stored in the packet object is the number of bytes "consumed" by the parser, i.e. bytes that are extract'ed, or advance'd past.  Restoring that number of bytes back to 0 must be done for resubmitted packets (and also ingress-to-egress cloned packets -- not handled yet in PSA implementation), so that when the parser code is executed again, it starts at the beginning of the packet again.  That might be what the save_buffer_state() is accomplishing here (or at least the part about saving the offset of 0 -- the restoring it back to the saved offset of 0 is done later with a different call, I believe).

peteli3: thanks - comment has been updated!

targets/psa_switch/psa_switch.cpp
```diff
@@ -342,6 +341,21 @@ PsaSwitch::ingress_thread() {
       continue;
     }
 
+    // resubmit - these packets get immediately resub'd to ingress, and skip
+    //            deparsing for now, do not move below multicast or deparse
```
antoninbas: what does "for now" mean here? resubmitted packets should not be deparsed so I don't think this should change in the future.

targets/psa_switch/psa_switch.cpp
```diff
@@ -313,6 +301,16 @@ PsaSwitch::ingress_thread() {
...
     phv = packet->get_phv();
+
+    /* Ingress cloning and resubmitting work on the packet before parsing.
+       `buffer_state` contains the `data_size` field which tracks how many
+       bytes are parsed by the parser ("lifted" into p4 headers). Here, we
+       track the buffer_state prior to parsing so that we can put it back
+       for packets that are cloned or resubmitted, same as in simple_switch.cpp
+
+       TODO update this along with simple_switch.cpp's analogous section */
+    const Packet::buffer_state_t packet_in_state = packet->save_buffer_state();
```
antoninbas: you can use `auto packet_in_state = ...`

targets/psa_switch/psa_switch.cpp
```diff
@@ -342,6 +341,21 @@ PsaSwitch::ingress_thread() {
       continue;
     }
 
+    // resubmit - these packets get immediately resub'd to ingress, and skip
+    //            deparsing for now, do not move below multicast or deparse
+    Field &f_resubmit = phv->get_field("psa_ingress_output_metadata.resubmit");
```
antoninbas: you can use `auto &` here, and should probably be `const` as well
`const auto &f_resubmit = ...`

targets/psa_switch/psa_switch.cpp
peteli3: sure! will make the change - but just for transparency (and since I'm quite new to c++) whats the purpose of favoring `auto` over the explicit type?

targets/psa_switch/psa_switch.cpp
peteli3: I meant it as "if I'm marked as resubmit, skip deparsing on this pass into ingress, but I'll get deparsed in a later pass" - but perhaps my wording is off.

peteli3: also I guess what I said isn't true for dropped packets in our implementation of bmv2... I'll just remove the "for now" from this comment

jafingerhut:
@antoninbas Does it look to you like the reasons the Travis tests are failing are because of some other recent commits, not because of the changes in this PR?

targets/psa_switch/psa_switch.cpp
antoninbas: I'll refer you to "Effective Modern C++" by Scott Meyers
In this specific case, it is mostly a matter of personal preference and uniformity. I do prefer not repeating the type when it can be avoided. If anyone ever needs to rename / change the return type, it will mean less changes.

antoninbas:
@jafingerhut yes, it looks like I need to merge my PR first: https://github.com/p4lang/behavioral-model/pull/787

antoninbas:
@peteli3 it seems that I cannot merge this because of conflicts

peteli3:
@antoninbas just rebased this branch against master. Can we try merging again?

PR #782, opened 2019-05-20
Support for group member activate / deactivate in PI implementation

antoninbas:
This in turn enables support for P4Runtime watch ports.
We leverage the ability to register a custom "selector" object with bmv2
ActionProfile instances, which takes care of dataplane member selection based on
a hash value computed by bmv2. This "selector" is notified by bmv2 when a member
is added / removed and by PI when a member is activated / deactivated.


antoninbas:
Depends on https://github.com/p4lang/PI/pull/471

PR #787, opened 2019-05-31
Implement new PI functions

antoninbas:
Ability to read single match table entries with:
 * pi_table_entries_fetch_one
 * pi_table_entries_fetch_wkey

Ability to read single action profile members / groups with:
 * pi_act_prof_mbr_fetch
 * pi_act_prof_grp_fetch

Also added a dummy implementation for
pi_table_default_action_get_handle. Note that at the moment bmv2 does
not support direct resources (meter / counter) for default entries, so
the handle which is returned by this function is very much useless.

### DIFF
--- a/PI/src/common.h
+++ b/PI/src/common.h
@@ -85,6 +85,14 @@ class Buffer {
   std::vector<char> data_{};
 };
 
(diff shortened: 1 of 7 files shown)


antoninbas:
Depends on https://github.com/p4lang/PI/pull/483

PR #789, opened 2019-06-04
Add "assert" and "assume" to JSON format documentation

antoninbas:
Also did some minor refactoring of assert / assume gtest.

Even though this is a pretty minor change, the JSON version number was
bumped up to 2.20.

### DIFF
--- a/docs/JSON_format.md
+++ b/docs/JSON_format.md
@@ -10,7 +10,7 @@ on each attribute.
 
 ## Current bmv2 JSON format version
 
-The version described in this document is *2.19*.
+The version described in this document is *2.20*.
 
 The major version number will be increased by the compiler only when
 backward-compatibility of the JSON format is broken. After a major version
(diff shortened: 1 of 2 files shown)


docs/JSON_format.md
```diff
@@ -482,6 +482,14 @@ in the enclosing action.
 must both resolve to integral values, which we call respectively `cond` and
 `offset`. If `cond` is 0, the primitive behaves exactly like `_jump` and we jump
 to the primitive call at index `offset`; otherwise the primitive is a no-op.
+- `assume` and `assert`: these 2 statements are used in P4 for formal program
+verification. They both take a single parameter (a boolean expression) and they
```
jafingerhut: The documentation changes look good to me.

I just wanted to mention that given the run-time behavior, they are not only for formal program verification, but also test-based verification, too.

PR #791, opened 2019-06-10
Make exit a core primitive

### LINKED ISSUE #790: Document "exit" primitive in docs/JSON_format.md
I just ran across my first BMv2 JSON file created by p4c with an "exit" primitive operation call in it.  I didn't see "exit" mentioned in the docs/JSON_format.md file, though.

antoninbas:
"exit" is a P4_16 built-in statement and should therefore be implemented
in a bmv2 target-agnostic way as much as possible. We make exit a bmv2
core primitive, and remove the exit definition from simple_switch +
psa_switch.

exit is therefore added to the JSON_format document, and the JSON
version number is bumped up to 2.21.

Note that at the moment targets are still required to call
`Packet::reset_exit` when the pipeline exits to reset the flag. I feel
like this could be done just before returning from `Pipeline::apply`,
but there may be some historical reason why I did not do it that way (so
to be safe I did not change anything...).

Fixes #790


jafingerhut:
For reference, this 2019-May-06 commit to p4c https://github.com/p4lang/p4c/commit/6a2546218904051950bfe38a3d7fcc80f03b8279 causes some P4_16+v1model programs to use the "exit" primitive in some compiled BMv2 JSON files.  I haven't tested, but it probably does the same for some PSA programs, too.

antoninbas:
Well, the "exit" primitive is meant to be used, and it should work properly for both simple_switch and psa_switch. This PR is just a small refactoring + documentation update.

jafingerhut:
Oh, this PR looks all good to me.  I only commented as some background info on why I even asked about the "exit" primitive in BMv2 at all.

I was running some tests with a tool that consume BMv2 JSON (p4pktgen), and found that some programs compiled with recent versions of p4c started using the "exit" primitive, whereas before that change I believe p4c was always avoiding using the "exit" primitive, instead using BMv2 table control flow to behave in the correct way.

PR #794, opened 2019-06-24
Runtime cli docs for multicast commands

jafingerhut:


jafingerhut:
It seemed reasonable to try documenting the multicast-related commands together, with an example.  Hopefully that will be straightforward for readers to understand what is going on.

PR #802, opened 2019-07-19
Update README to reference p4c compiler

antoninbas:
Instead of legacy p4c-bm compiler

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,21 +92,34 @@ To run the unit tests, simply do:
 
 ## Running your P4 program
 
-To run your own P4 programs in bmv2, you first need to transform the P4 code
+To run your own P4 programs in bmv2, you first need to compile the P4 code
 into a json representation which can be consumed by the software switch. This
 representation will tell bmv2 which tables to initialize, how to configure the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,21 +92,34 @@ To run the unit tests, simply do:
...
+ * [p4c-bm](https://github.com/p4lang/p4c-bm) is the legacy compiler for bmv2
+   (no longer actively maintained) and only supports P4_14 programs.
 
-The json file can now be 'fed' to bmv2. Assuming you are using the
-*simple_switch* target:
+Assuming you have installed the p4c compiler, you can obtain the json file for a
+P4_16 v1model program as follows:
 
-    sudo ./simple_switch -i 0@<iface0> -i 1@<iface1> <path to JSON file>
+    p4c --target bmv2 --arch v1model --std p4-16 --json <prog>.p4
```
jafingerhut: `--std p4-16` is unnecessary, but also not a problem.

I normally do not use the `--json` option mentioned here.  Would you recommend using it?

antoninbas: Sorry, it's a typo, I'll remove it :)

Issue #687, opened 2018-11-21
Logging/Printing user-defined messages


endrigoshi:
Hi all!
Is it possible with p4_16 to log user-defined messages that are later printed in terminal using `--log-console` option?
Normally I am assuming that this is not possible but just in case I am asking to be completely sure about it.

jafingerhut:
I think such a thing would be a very useful addition to the language, even if it is only for debugging programs on software switches like simple_switch.

At the moment, a slightly more verbose alternative is to use a 'debug table', which is simply a table with `actions = { NoAction; }` and `const default_action = NoAction;`.  Put in the key all header and/or metadata fields that you want to see the value of at a particular point in the program's execution, and then `apply` that table at that point in the program.  The values of all of those fields will show up in the console log when execution reaches that point.

antoninbas:
If an extern is added to v1model to log user messages, it would be pretty straightforward to implement it for bmv2 simple _switch and redirect messages to the console logs ... assuming that this extern only supports logging plain string messages or single values.

```
extern void log_msg(in string msg);
extern void log_value<T>(in string msg, in T v);
```

Of course the P4_16 language would need to expose a string literal type for the above to work.

Having more general support like in the fmt library (https://github.com/fmtlib/fmt) would be awesome, but that couldn't be done with existing bmv2 capabilities because bmv2 doesn't support variadic externs functions (P4_16 doesn't support them either):
```
log("I'm in action 'a', parameter 'x' is equal to {} and parameter 'y' is equal to {}", x, y);
```

endrigoshi:
@jafingerhut thank you for your reply. that seems a really smart trick! :)

antoninbas:
The solution described in https://github.com/p4lang/behavioral-model/issues/687#issuecomment-440824548 was implemented a while back.

PR #805, opened 2019-07-26
Support for logging user defined messages (#687)

lbanduka:
Added support for logging user defined messages in P4 that can be later
printed in terminal using --log-console option.

Added two variations of function log_msg:
- void log_msg(string s);
	used to log plain string s.
- void log_msg(string format, {value v1, value v2, ...});
	Format is a string containing one or more curly braces '{}' that is replaced
	by values v1, v2 and so on, in the same order.

Example:
bit<4> x = 5;
bit<4> y = 7;
log_msg("x = {}, y = {}", {x, y});
Output: "x = 5, y = 7"

Or:
log_msg("Entering action XY");
Output: "Entering action XY"

NOTES:
If you want to log character '{' or '}' you must add escape character '\'.
(log_msg("Logging open bracket \{") will log message "Logging open bracket {")


jafingerhut:
Nice!  I tried out this modified version of behavioral-model and your corresponding modifications for p4c on a little test program, and it all seems to work as advertised.  I noticed that you even added support for enabling log_msg calls inside of deparser code, which is a little bit special in the BMv2 JSON data created, and how it is implemented in simple_switch.  Nice.  And it supports calling log_msg from parser code, too.

I can easily see people wanting later to have support for controlling whether the numeric values are printed in decimal, hexadecimal, binary, etc., but this is a good step towards that, and it seems like it should be extensible in that direction later, after this is merged in, if these changes otherwise look acceptable in terms of code quality, style, correctness, etc.

src/bm_sim/P4Objects.cpp
```diff
@@ -422,6 +422,57 @@ P4Objects::add_primitive_to_action(const Json::Value &cfg_primitive,
     if (type == "hexstr") {
       const auto value_hexstr = cfg_parameter["value"].asString();
       action_fn->parameter_push_back_const(Data(value_hexstr));
+    } else if (type == "parameters_vector") {
```
antoninbas: is there any chance that this code and the "regular" parameter code can be unified in some way?

src/bm_sim/P4Objects.cpp
```diff
@@ -422,6 +422,57 @@ P4Objects::add_primitive_to_action(const Json::Value &cfg_primitive,
...
+      std::vector<ActionParam> params_vector;
```
antoninbas: Is it possible to move the logic to build the parameter vector to actions.h / actions.cpp, so we don't have to expose the "internals" of the actions module (such as the `ActionParam` class)?

Something like:
```
struct ActionParamVectorBuilder {
    ActionParamVectorBuilder(ActionFn *action_fn)
        : action_fn(action_fn) { }

    void push_back_action_data(int action_data_offset);
    void push_back_const(...);
    void push_back_field(...);

    std::vector<ActionParam> get() const { return param_vector; }

    ActionFn *action_fn;  // needed for push_back_const
    std::vector<ActionParam> param_vector;
};
```

Please let me know if you see an issue with that approach.

targets/simple_switch/primitives.cpp
```diff
@@ -473,6 +475,66 @@ class truncate_ : public ActionPrimitive<const Data &> {
 
 REGISTER_PRIMITIVE_W_NAME("truncate", truncate_);
 
+class log_msg : public ActionPrimitive<const std::string &, std::vector<Data>> {
```
antoninbas: please use a const reference for the vector, there is no need to pay for a copy here I think

targets/simple_switch/tests/test_logging.cpp
```diff
@@ -0,0 +1,139 @@
+/* Copyright 2019-RT-RK Computer Based System
```
antoninbas: nit: shouldn't it be `2019 RT-RK Computer Based System`

targets/simple_switch/primitives.cpp
```diff
@@ -473,6 +475,66 @@ class truncate_ : public ActionPrimitive<const Data &> {
...
+  void operator ()(const std::string &format, std::vector<Data> data_vector) {
+    std::string str = format;
+    std::string::size_type open = 0;  // position of the opened bracket
+    std::string::size_type end = 0;  // position of the matching closed bracket
+    unsigned int counter = 0;  // number occurrences of '{}' in string format
+    std::string::size_type tmp;
+
+    while ((open = str.find('{', open+1)) != std::string::npos) {
```
antoninbas: I wonder if we can make that code less verbose by using C++11 regex, but not a strict requirement IMO

lbanduka:
Hello everybody,
Thank you for your time and reviews. I will try to follow your suggestions and make these changes.
As soon as I finish, I will update PR.

include/bm/bm_sim/deparser.h
```diff
@@ -23,18 +23,42 @@
...
 #include "phv_forward.h"
+#include "stateful.h"
 
 namespace bm {
 
 class Packet;
 
 class Checksum;
 
+struct DeparserOp {
```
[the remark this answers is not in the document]
lbanduka: DeparserOp cant be moved to deparser.cpp because it is type argument to unique_ptr and the type argument to unique_ptr must be complete type (forward declared struct is incomplete). With DeparserOpMethodCall there is no problem so I moved it.

src/bm_sim/P4Objects.cpp
lbanduka: I think there is no chance that this code can be unified with regular one because when we have parameter vector we go one step further, and i think we must go through for loop separately for parameter vector types.

targets/simple_switch/primitives.cpp
lbanduka: I think we need to pay for a copy here. We make new local variable of vector<Data> in "ActionParam::to" function, and we cant use reference to a local variable. If I misunderstood something here please let me know.

src/bm_sim/P4Objects.cpp
lbanduka: I need to think about this one more time, maybe I can solve this

include/bm/bm_sim/deparser.h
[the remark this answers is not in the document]
antoninbas: This is not true. The destructor definition needs the complete type declaration, so the destructor needs to be defined in the cpp file, but it is perfectly fine to forward-declare the type argument to `unique_ptr`. This is actually used very often in modern C++ when defining PIMPL, with a smart pointer for the implementation member. You can Google "PIMPL with unique_ptr". See for example: https://stackoverflow.com/questions/9020372/how-do-i-use-unique-ptr-for-pimpl

lbanduka:
Hi everyone,
I made changes according to your suggestions and updated PR. If you have more requests feel free to comment.

include/bm/bm_sim/actions.h
```diff
@@ -749,6 +794,24 @@ class ActionFnEntry {
...
+  void parameter_push_back_last_header_stack_field(
+      header_stack_id_t header_stack, int field_offset);
+  void parameter_push_back_field(header_id_t h_id, int offset);
+  void parameter_push_back_param_vector();
+
+  size_t get_num_params() const { return af->get_num_params(); }
+
+ private:
+  std::vector<ActionParam> param_vector;
+  ActionFn *af;
```
antoninbas: the design doesn't look great to me here: you inherit from a class that you also encapsulate.
The previous version of the code had some duplication, but it made sense to me.

PR #806, opened 2019-08-06
Update _pi_init implementation

antoninbas:
_pi_init is now required to return the "ABI version" so that the core PI
library can check for compatibility at runtime.


antoninbas:
See https://github.com/p4lang/PI/pull/491

PR #817, opened 2019-10-26
V1model notes on register extern

jafingerhut:


docs/simple_switch.md
```diff
@@ -856,10 +925,10 @@ accesses some of the same register objects.  You need not use the
 `@atomic` annotation in your P4_16 program in order for this level of
 atomicity to be guaranteed for you.
 
-It is still recommended to use the `@atomic` annotation in your P4_16
-code to mark blocks of code containing multiple register accesses that
-you wish to be atomic relative to other packets being processed, but
-as of March 2019, the BMv2 implementation ignores such annotations.
+BMv2 v1model as of October 2019 _ignores_ `@atomic` annotations in
+your P4_16 programs.  Thus even if you use such annotations, this does
+not cause BMv2 to treat any block of code larger than one action call
```
antoninbas: to be exact it's one action call or one parse state

jafingerhut: You are saying that simple_switch supports accessing registers inside of a parser state?  I never tried that before, and would have guessed it wasn't supported, but I may try it now to see what happens.

antoninbas: Yes bmv2 support extern method calls in the parser. The real question is: is it supported by the p4c bmv2 backend :) ?

https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md#parsers

jafingerhut: I tried a program that read and wrote to a register inside of a parser state, and p4c gave an error while compiling it, so it appears that it does not support it even if bmv2 does:
```
$ p4c --target bmv2 --arch v1model access-register-inside-parser.p4
access-register-inside-parser.p4(74): [--Werror=unsupported] error: MethodCallStatement: Unsupported in parser on this target
        input_port_pkt_count.read(pkt_count, (bit<32>) stdmeta.ingress_port);
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
access-register-inside-parser.p4(77): [--Werror=unsupported] error: MethodCallStatement: Unsupported in parser on this target
        input_port_pkt_count.write((bit<32>) stdmeta.ingress_port, pkt_count);
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
The program I tried is attached, in case anyone is curious.

It doesn't seem to let me attach it to this comment, so will try it on a 'top level' comment rather than on a comment replying to this one.

jafingerhut:
Sample program I tried that p4c gives an error message while compiling, mentioned in another comment on this PR.
[reg-in-parser.p4.txt](https://github.com/p4lang/behavioral-model/files/3775249/reg-in-parser.p4.txt)

docs/simple_switch.md
antoninbas: I suspected as much

jafingerhut: I made another commit briefly mentioning parser states being atomic as well as actions.

antoninbas: Ooops, I merged too quickly then.
Feel free to open a new PR / save it for future edits / forget I ever mentioned anything about parser states.

jafingerhut: No worries, it wasn't a big addition, and it seems somewhat low priority to add a note on parser states being atomic given that p4c seems to support a fairly short list of method calls there.

PR #833, opened 2019-12-11
Document command-line options for simple_switch_grpc

### LINKED ISSUE #831: Is there a simple_switch_grpc switch port for P4Runtime PacketIn/PacketOut ?
Is there some either hard-coded or configurable port number that one can send packets to in a P4 program, e.g. using the P4_16 v1model architecture, such that packets sent to that port become PacketIn messages to a P4Runtime client that is connected to simple_switch_grpc (when it is acting as a P4Runtime server) ?

Similarly for PacketOut messages, which I presume would be the same port number according to the P4 code.

antoninbas:
Fixes #831

### DIFF
--- a/targets/simple_switch_grpc/README.md
+++ b/targets/simple_switch_grpc/README.md
@@ -36,6 +36,45 @@ Github or the p4-dev mailing list.
     [sudo] make install  # if desired
     ```
 
+## Running simple_switch_grpc
+
+For information on how to run simple_switch_grpc, use `simple_switch_grpc
+--help`. In particular, take a look at the target-specific command-line options
+at the end of the help message. These command-line options are specific to
+simple_switch_grpc (to be precise, some of them are shared with the
+simple_switch binary). When invoking simple_switch_grpc, these options will need
+to be separated from standard bmv2 options using `--`.
+
+Most users will invoke simple_switch_grpc as follows:
+```
+simple_switch_grpc --skip-p4 \
+    -i <PORT1>@<IFACE1> -i <PORT2>@<IFACE2> <more ports> \
+    -- --grpc-server-addr <IP>:<TCP PORT> --cpu-port <CPU_PORT>
+```
+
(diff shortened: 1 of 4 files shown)


targets/simple_switch_grpc/README.md
```diff
@@ -36,6 +36,46 @@ Github or the p4-dev mailing list.
...
+  provide a P4 "pipeline" (including a JSON file and a P4Info message) with the
+  P4Runtime `SetForwardingPipelineConfig` RPC. **This is the recommended way to
+  use simple_switch_grpc.**
+* `--grpc-server-addr` is used to provide a socket address on which the
+  P4Runtime server will be run (e.g. localhost:1234, 192.168.1.1:31416,
+  [::1]:27182, ...). It will default to 0.0.0.0:50051.
+* `--cpu-port` is used to provide a CPU port number and enable P4Runtime
+  packet-in / packet-out support. **We recommend that you use this option** as
+  otherwise you will not be able to receive / send packets using the P4Runtime
+  `StreamChannel` bi-directional stream. Packets sent to this port in ingress
```
jafingerhut: Maybe simply "packets sent to this port" without the "in ingress" part?  The packet will still go through egress processing before going out of the port, yes?  And egress might drop it, or recirculate it, and then it will not go to the controller (or at least not yet)?

antoninbas: I wanted to make it clear that you have to set the egress spec in ingress, just in case... I'll clarify.

jafingerhut: Can the cpu_port be one of several ports in a multicast group, and that works?  Can it be assigned as the port to go to in a mirror/clone session, and that works?  If so, assigning the value of the cpu port to egress_spec during egress is just one of the ways a packet gets to that port, and cpu_port is no different than any other port in that regard.  Maybe I am too close to the subject material, though :-)

antoninbas: Thanks for thinking of everything Andy :P I'll remove the "clarification" and let people deal with it.

Issue #844, opened 2020-01-18
Implementing match_kind optional in simple_switch?


jafingerhut:
Related to this PR for p4c: https://github.com/p4lang/p4c/pull/2141

At least couple of possible approaches are possible here:

+ add a match_kind "optional" to the BMv2 JSON file, including defining how such fields should be described in the "entries" section of the BMv2 JSON file when the P4 source code has "const entries" for a table with a match_kind optional field as part of its key

+ turn all match_kind "optional" fields in P4_16 source code into "ternary" in the BMv2 JSON.  Possible down side: The information about the field being "optional" in the source code is then represented nowhere in the BMv2 JSON file, so simple_switch would have no information on which it could know that it should reject table entry insert/modify operations that used a mask that should not be allowed for "optional" fields.

Any preferences for one of those two approaches?  Or some third option I'm not thinking of?

jafingerhut:
Addendum to previous comment: The second option could still have the knowledge of which table search key fields were match_kind "optional" in the P4Info file, so the P4Runtime API should still be able to enforce the mask restrictions on them, even if the BMv2 JSON file called the ternary.  I think it would only be the Thrift API, or others that did not have some other source of information outside of the BMv2 JSON file, that would lose that information and "believe" that those fields were "ternary" in the source code.  Maybe that is an acceptable loss of information for Thrift API clients?

antoninbas:
I think the decision based on the discussion we had in https://github.com/p4lang/behavioral-model/pull/845 is to go with your second option.

jafingerhut:
Antonin, where do you believe code changes would need to be made in order for simple_switch_grpc to support the new OPTIONAL match_kind in a P4Info file?

I am assuming that would require first merging in a change like in this p4lang/p4runtime PR: https://github.com/p4lang/p4runtime/pull/259

but I suspect that more is needed?

antoninbas:
Most of the changes will be in https://github.com/p4lang/PI.

You can look for occurrences of `TERNARY` in https://github.com/p4lang/PI/blob/master/proto/frontend/src/device_mgr.cpp. Unfortunately, all the changes will not be contained in this file. The P4Info message is translated into a C struct when `SetForwadingPipelineConfig` pushes a new config. We then use that in-memory struct when processing P4Runtime Write RPCs (to validate the table entry, etc...). So that struct, which is part of the PI C library will need to store information about which field is an optional match.

[...]

PR #845, opened 2020-01-18
Add support for match_kind optional to BMv2 JSON file

jafingerhut:
Only add a match_type of optional for individual fields of a table
search key.  The overall table match_type is still ternary if any of
the fields are match_type ternary or optional, but none are range.

const entries table fields with match_kind optional are described by
JSON keys "key" and "mask" exactly like table fields with match_kind
ternary are.  There is no difference in syntax, but the value
associated with the "mask" key will always be 0 for a completely
wildcarded field value, or all 1s in every bit position of the entire
field width for an exact match field value.  No other masks will ever
appear.

### DIFF
--- a/include/bm/bm_sim/match_units.h
+++ b/include/bm/bm_sim/match_units.h
@@ -50,11 +50,14 @@ struct MatchKeyParam {
   // VALID used to be before RANGE, but when RANGE support was added, it was
   // easier (implementation-wise) to put RANGE first. Note that this order only
   // matters for the implementation.
+  // TODO(jafingerhut): Where should OPTIONAL be added in this
+  // sequence, if order is important?
   enum class Type {
     RANGE,
     VALID,
     EXACT,
     LPM,
+    OPTIONAL,
     TERNARY
   };
 
--- a/src/bm_sim/match_tables.cpp
+++ b/src/bm_sim/match_tables.cpp
@@ -51,7 +51,7 @@ create_match_unit(const std::string match_type, const size_t size,
   else if (match_type == "lpm")
     match_unit = std::unique_ptr<MULPM>(
         new MULPM(size, match_key_builder, lookup_factory));
-  else if (match_type == "ternary")
+  else if ((match_type == "ternary") || (match_type == "optional"))
     match_unit = std::unique_ptr<MUTernary>(
         new MUTernary(size, match_key_builder, lookup_factory));
   else if (match_type == "range")
--- a/src/bm_sim/match_units.cpp
+++ b/src/bm_sim/match_units.cpp
@@ -66,6 +66,8 @@ MatchKeyParam::type_to_string(Type t) {
       return "LPM";
     case Type::TERNARY:
       return "TERNARY";
+    case Type::OPTIONAL:
+      return "OPTIONAL";
   }
   return "";
 }
@@ -86,6 +88,7 @@ std::ostream& operator<<(std::ostream &out, const MatchKeyParam &p) {
       out << "/" << p.prefix_length;
       break;
     case MatchKeyParam::Type::TERNARY:
+    case MatchKeyParam::Type::OPTIONAL:
       out << " &&& ";
       dump_hexstring(out, p.mask);
       break;
@@ -383,6 +386,7 @@ class MatchKeyBuilderHelper {
           // should only happen for RangeMatchKey
           // range treated the same as ternary
         case MatchKeyParam::Type::TERNARY:
+        case MatchKeyParam::Type::OPTIONAL:
           {
             auto mask_start = key.mask.begin() + byte_offset;
             auto mask_end = mask_start + nbytes;
@@ -468,6 +472,7 @@ class MatchKeyBuilderHelper {
           break;
         case MatchKeyParam::Type::RANGE:
         case MatchKeyParam::Type::TERNARY:
+        case MatchKeyParam::Type::OPTIONAL:
           assert(0);
       }
       first_byte += param.key.size();
@@ -537,6 +542,7 @@ class MatchKeyBuilderHelper {
           break;
         case MatchKeyParam::Type::RANGE:
         case MatchKeyParam::Type::TERNARY:
+        case MatchKeyParam::Type::OPTIONAL:
           key->mask.append(param.mask);
           break;
       }
@@ -584,6 +590,7 @@ MatchKeyBuilder::match_params_sanity_check(
           return false;
         break;
       case MatchKeyParam::Type::TERNARY:
+      case MatchKeyParam::Type::OPTIONAL:
         if (param.mask.size() != nbytes) return false;
         break;
       case MatchKeyParam::Type::RANGE:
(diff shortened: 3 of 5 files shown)


jafingerhut:
If the answer to the question in issue https://github.com/p4lang/behavioral-model/issues/844 is 'we should add a match_kind "optional" to the BMv2 JSON file', then the changes in this PR _might_ be a start at implementing that.  In looking over the code, I realize I have knowledge of only a fairly small amount of the behavioral-model code base, so even these few changes should be treated as suspect until reviewed by someone more knowledgeable of the code base, or perhaps even better rewritten by them.

These changes do not yet make any attempt at all to modify any control plane API implementations such as Thrift or P4Runtime, only the parsing and use of the BMv2 JSON file.

include/bm/bm_sim/match_units.h
```diff
@@ -50,11 +50,14 @@ struct MatchKeyParam {
   // VALID used to be before RANGE, but when RANGE support was added, it was
   // easier (implementation-wise) to put RANGE first. Note that this order only
   // matters for the implementation.
+  // TODO(jafingerhut): Where should OPTIONAL be added in this
```
antoninbas: You just need to put OPTIONAL with TERNARY here since they are implemented the same way. I suggest putting OPTIONAL *after* TERNARY and replacing this TODO comment with something like: "OPTIONAL is just a TERNARY match where the mask is constrained to be one of 2 values (all 0s or all 1s), so it is implemented just like TERNARY. The 2 match types should be listed one after the other in the enum definition."

src/bm_sim/match_units.cpp
```diff
@@ -584,6 +590,7 @@ MatchKeyBuilder::match_params_sanity_check(
           return false;
         break;
       case MatchKeyParam::Type::TERNARY:
+      case MatchKeyParam::Type::OPTIONAL:
```
antoninbas: let's treat them differently here. This function (sanity check) is probably a great place to validate the mask for the optional case.

src/bm_sim/match_tables.cpp
```diff
@@ -51,7 +51,7 @@ create_match_unit(const std::string match_type, const size_t size,
   else if (match_type == "lpm")
     match_unit = std::unique_ptr<MULPM>(
         new MULPM(size, match_key_builder, lookup_factory));
-  else if (match_type == "ternary")
+  else if ((match_type == "ternary") || (match_type == "optional"))
```
antoninbas: IIRC this is not a correct change. The "match_type" here is the match type of the table, not the match type of an individual field. And a table is never going to have the optional match type, it will just be ternary.

antoninbas: By the way, this match type is determined by the compiler backend.

jafingerhut:
@antoninbas You said in an earlier comment: "I'm leaning towards hiding "optional" completely from the bmv2 core code, and instead handling it entirely at the runtime interface level (both Thrift - i.e. src/bm_runtime - and P4Runtime). The runtime can validate the mask and then just pass it on to bmv2 as a ternary match."

Are you saying that you think:

(a) "optional" is best left completely out of the BMv2 JSON file format, using "ternary" everywhere "optional" appears in the P4_16 source code?  In this case, I think the only changes needed to simple_switch_grpc would be for handling "optional" in the P4Info file.

(b) "optional" should be used in the BMv2 JSON file, only as the "match_type" of table search key fields that are defined as "optional" in the P4_16 source code, but it should be "converted to ternary" almost immediately as soon as simple_switch reads it, so that most of the behavioral-model source code is not aware of it?

(c) something else?

antoninbas:
@jafingerhut when I wrote the review I was thinking about option (a). However I also thought that we would be able to expose "optional" in the Thrift API and in the CLI, which as you noted in issue #844 would not be possible if "optional" is left out of the JSON file.

I still think (a) is a solid option. **It won't break existing tools that consume the JSON by introducing a new match type**. I don't know how many people still use the Thrift API directly or the runtime CLI, but they can probably live with that small discrepancy (I imagine there are other discrepancies already). We would still be able to support "optional" properly in P4Runtime. However, if you are willing to spend some additional time on this PR to make sure that nothing breaks by adding native "optional" support in the JSON, I think this is potentially also a good way to go, and it gives us more opportunity in the future to improve logging for example. I left a few comments already in my review that you would need to address. In addition to that you would need to audit the tools that consume the JSON. The runtime CLI and the bmv2 debugger come to mind. Unfortunately we don't have unit tests for these, so the breakage probability is high. I'm almost 100% sure the CLI will crash if there is an unknown match type in the JSON. I doubt anyone is still generating the PD for bmv2, but that tool (https://github.com/p4lang/p4c-bm) also consumed bmv2. So does some code in https://github.com/p4lang/PI.

Let's not go with (b).

So to conclude, I'm really on the fence, and I think I'll leave the decision to you. If you are willing to  update & test a lot of things, I won't stand in your way :) If we go with (a), the only work required is for this repo is updating the documentation, and maybe adding a comment in the Thrift IDL file and improving help messages for the runtime CLI. The only significant drawback is that we do not enforce constraints on the mask, but I'm not sure how many people will be impacted negatively by that. Let me know what you think.

antoninbas:
Regarding what needs to be updated if "optional" is added to the bmv2 JSON:
* p4c-bm is officially deprecated as per the README, so no need to worry about it
* maybe it's a good opportunity to deprecate the alternate PI implementation for bmv2: https://github.com/p4lang/PI/tree/master/targets/bmv2, so that would be one less piece of code code that needs to be updated

The p4lang software should definitely be manageable. And if anyone else is using the JSON output by p4c-bmv2, it seems reasonable to ask them to update their tools (after all it is called the *bmv2* JSON). It is also reasonable to ask users to update bmv2 to a newer version if they start using "optional" in their P4 programs.

src/bm_sim/match_units.cpp
antoninbas: one more thing that comes to mind:
the check `if (param.type != f_info.mtype) return false` above will fail at the moment for optional match fields, that will require changes in src/bm_runtime

jafingerhut:
I would prefer to keep the list of things to update and test shorter than option (b), personally.  Option (a) sounds much better to me.

If it means that someone can use the Thrift API and/or simple_switch_CLI, and they see "ternary" instead of "optional" for those table key fields, and can insert disallowed masks, that seems reasonable to document as a limitation of the current implementation.

antoninbas:
Just to be clear, I was comparing option (a) with adding "optional" to the bmv2 JSON. I didn't consider (b) a real contender. But I think we are on the same page. We can start with the least painful option - (a) - and improve from there if needed.

jafingerhut:
I am closing this PR, as if we go with the plan of using the existing match_kind ternary in the BMv2 JSON file for all table search key fields with match_kind optional, the only changes required to p4lang/behavioral-model code will be for simple_switch_grpc code to handle the new OPTIONAL match_kind in P4Info files that it reads.

Antonin, feel free to reopen this PR if you think there is something here that should be kept, based upon that plan of no changes to the BMv2 JSON file format.

PR #852, opened 2020-01-26
Add more commands to runtime cli

jafingerhut:
Also make several corrections and clarifications to the documentation
describing how cloned packets are handled by simple_switch, and add
several simple_switch_CLI commands that were not added earlier.

Also add several simple_switch_CLI commands that were not added earlier.

### DIFF
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -348,7 +348,7 @@ processing.  Each of the clones is processed independently from each
 other, and independently from the original packet.
 
 Search for occurrences of the word "clone" in [this
-article](https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md#pseudocode-for-what-happens-at-the-end-of-ingress-and-egress-processing)
+article](simple_switch.md#pseudocode-for-what-happens-at-the-end-of-ingress-and-egress-processing)
 for more details on how `simple_switch` deals with cloned packets.
 
 `simple_switch` supports 32,768 independent clone sessions, numbered
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -11,6 +11,9 @@ TODO: act_prof_modify_member
 TODO: act_prof_remove_member_from_group
 TODO: counter_read
 TODO: counter_reset
+TODO: counter_write
+TODO: get_time_elapsed
+TODO: get_time_since_epoch
 TODO: help
 TODO: load_new_config_file
 ```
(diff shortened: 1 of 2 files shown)


docs/runtime_CLI.md
```diff
@@ -317,6 +320,99 @@ the meter type is packets, the rate is entered in packets/microsecond and
...
+processing.  Each of the clones is processed independently from each
+other, and independently from the original packet.
+
+Search for occurrences of the word "clone" in [this
+article](https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md#pseudocode-for-what-happens-at-the-end-of-ingress-and-egress-processing)
+for more details on how `simple_switch` deals with cloned packets.
+
+TBD: How many clone sessions are supported by `simple_switch`, without
+using any special default compilation options nor modifications to the
+source code?
```
jafingerhut: @antoninbas Do you happen to know the answer to this?  I can search through the code to find out, if. you prefer.  Also, I believe it is correct that clone session id 0 is not supported in the simple_switch data plane, but good if you can double-check me on that.

antoninbas: You're the one who wrote that code Andy: https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/register_access.h#L46 :)
* 0 is a valid  clone session id for bmv2 (so it can be used with the CLI) but not for P4Runtime. This is because we use an extra bit to encode whether the packet is being cloned.
* because the validity bit + the session id use 16 bits, that leaves us with 15 bits for the session id so valid values are in the range `[0, 32768]`.

jafingerhut: I have updated that text to give the range of clone session ids supported as 0 up to 32,767.  Thanks for reminding me :-)

jafingerhut:
Good comment on being explicit about simple_switch_CLI-specific things vs. those that are not.  I have marked the new commands as [simple_switch_CLI only].  If some other way of marking these would be better, let me know.

docs/runtime_CLI.md
```diff
@@ -317,6 +320,97 @@ the meter type is packets, the rate is entered in packets/microsecond and
...
+whatever it normally does, as if your program had not called a clone
+operation, but the PRE (Packet Replication Engine) part of
+`simple_switch` will also create zero or more copies, or clones, of
+the current packet, and cause them to be enqueued in the packet buffer
+such that later each of the cloned packet(s) will perform egress
+processing.  Each of the clones is processed independently from each
+other, and independently from the original packet.
+
+Search for occurrences of the word "clone" in [this
+article](https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md#pseudocode-for-what-happens-at-the-end-of-ingress-and-egress-processing)
```
antoninbas: please just use `(/docs/simple_switch.md#pseudocode-for-what-happens-at-the-end-of-ingress-and-egress-processing)` as the link

jafingerhut: I tried without the `/docs/` and it works when clicking the links, and I think is the shortest path that works.  Let me know if you really want the `/docs/` in the two links that I changed in the last two commits.

antoninbas: Without docs/ is good too

PR #858, opened 2020-02-02
mininet: Close thrift socket after test

rst0git:
Make sure the socket file descriptor is closed after testing connection to Thrift server.


rst0git:
I just realised that this works only with Python 3 and in Python 2 it fails with `AttributeError: __exit__`:
https://stackoverflow.com/questions/48055880/attributeerror-exit-with-socketserver-on-python-3-4-3

https://github.com/p4lang/behavioral-model/pull/860 should fix this.

Issue #857, opened 2020-02-01
Support non-exact matches for value_sets


antoninbas:
As per the P4_16 specification, the `@match` annotation can be used when declaring a struct to be used as the type for a `value_set`:
> If the value_set takes a struct as type parameter, the runtime API can use the struct field names to name the objects in the value set. The match type of the struct field is specified with the `@match` annotation. If the `@match` annotation is not specified on a struct field, by default it is assumed to be `@match(exact)`. A single non-exact field must be placed into a struct by itself, with the desired `@match` annotation.

We should try to support this in the bmv2 implementation. At the moment we only support exact matches. This is probably a significant change however, which will require a JSON format update.

jafingerhut:
I have no near-term plans to tackle this, but wanted to point out that there is significant similarity between parser value sets and P4 tables, when the `@match` annotation is supported.

+ Both can have any number of match fields, where each match field has an independent name, width in bits, and match_kind.
+ Both support the ability to add and remove entries at run time, via a control plane API.  I do not know whether anyone cares to support modify of an existing entry in a parser value set, but I do not see any fundamental reason why it could not be supported, if someone wanted to implement it.

The primary difference is that unlike a P4 table, the entries in a parser value set have no action, nor action parameters.

So it seems there is an opportunity to share at least some code that implements these capabilities in P4 tables, when implementing parser value sets.

PR #859, opened 2020-02-02
Add support for parser value sets (PVS) to runtime_CLI

### LINKED ISSUE #855: Support parser value sets in runtime_CLI
By integrating with the Thrift API: https://github.com/p4lang/behavioral-model/blob/master/thrift_src/standard.thrift#L659

The user will need to input "compressed" binary values which is not great for the multi-field case, but that's better than nothing. In particular, it will enable writing STF tests in p4c.

antoninbas:
We now provide the following commands: pvs_add, pvs_remove, pvs_get,
pvs_clear. Implementing pvs_get and pvs_clear required adding the
equivalent RPCs to the Thrift API.

This commit also adds a bmv2 JSON sanity check for the "mask" attribute,
when it is used in a PVS parser transition.

The included JSON file, pvs_struct_2.json, was used to test the
implementation of the CLI commands. It was generated from a P4 program
written by @jafingerhut for the compiler testsuite.

Fixes #855

### DIFF
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -418,6 +418,50 @@ configuration of a clone session.
 ```
 TODO: port_add
 TODO: port_remove
+```
+
+### pvs_add, pvs_clear, pvs_get, pvs_remove
+
+These commands are used to control an instance of a P4 parser value
+set. `pvs_add` is used to add an entry to a value set, `pvs_remove` is used to
+remove an entry, `pvs_clear` is used to remove *all* entries, and `pvs_get` is
+used to list all entries.
+
+P4 allows a value_set to take a struct as the type parameter, such as in the
+following example:
+```
+struct vsk_t {
+    bit<16> f1;
+    bit<7> f2;
+}
+value_set<vsk_t>(4) pvs;
+select (<X>) {
+    pvs: accept;
+    _: reject;
+}
+```
+When adding or removing an entry, you must provide a single integer value, which
+must fit within the value set's "compressed bitwidth", or the CLI will display
+an error. When a value set's entries consist of multiple individual fields, as
+is the case in the example above, the "compressed bitwidth" is defined as the
+sum of the bitwidths of all these individual fields, without padding. When the
+type parameter for the value set is an integer (signed or unsigned), or a struct
+with a single field, the "compressed bitwidth" is simply the bitwidth of the
+integer / field.
+
+In the future, the CLI may provide a more intuitive interface and enable you to
+provide individual integer values for the different fields which constitute the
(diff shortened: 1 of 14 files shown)


hesingh:
The JSON generated by bmv2 has a unique id for each parse_vset.  Why don't you use the id to search before add or delete?

Thanks for working  on these changes so quick.

docs/runtime_CLI.md
```diff
@@ -320,6 +320,50 @@ burst_size is the number of packets.
...
+}
+value_set<vsk_t>(4) pvs;
+select (<X>) {
+    pvs: accept;
+    _: reject;
+}
+```
+When adding or removing an entry, you must provide a single integer value, which
+must fit within the value set's "compressed bitwidth", or the CLI will display
+an error. When a value set's entries consitute of multiple individual fields, as
```
jafingerhut: maybe "consist of" instead of "consitute of" ?  "consitute" is not a word, in any case.

jafingerhut:
I can look into this in more detail to see if I can figure it out, but it is currently confusing me how this implementation can work, when the "compile time constant" entries of parser transitions of a select must have each field padded out to a whole number of bytes, but the parser value set entries must not be padded out to a whole number of bytes.

jafingerhut:
OK, looking more closely, I guess the value of the BMv2 JSON key "transition_key" is a list of fields that each have an exact bit width that need not be a multiple of 8, and matches what is in the P4 source code.  That answers my question of how simple_switch can have enough information to interpret a compressed bit string correctly, and also the other constant select expressions in the BMv2 JSON file.

docs/runtime_CLI.md
antoninbas: Fixed, thanks for pointing this out

antoninbas:
@jafingerhut yes that's correct. And I believe the rationale for the current implementation is that it was introduced for P4_14 parser value sets, which were initialized with a single bitwidth, and while you could combine multiple fields when performing a lookup on the set, the declaration of the pvs itself didn't impose any structure.

I think the PVS implementation in bmv2 could benefit from a major overhaul (maybe https://github.com/p4lang/behavioral-model/issues/857 should be a more general issue). The implementation really caters to P4_14 at the moment. It even accommodates for the case where a single PVS is used in 2 different parser states with different lookup keys.

antoninbas:
To illustrate what I meant at the end of my previous comment. I believe the following program is a valid P4_14 program but cannot be translated easily in a P4_16 program and is rejected by p4c:
```

mbudiu-vmw:
You should submit the p4-14 program as an issue to p4c so we can fix it.

PR #860, opened 2020-02-02
mininet: Fix closing thrift test socket

rst0git:
Support for the context manager protocol for socket was added in Python 3.6

https://docs.python.org/3/library/socketserver.html#socketserver.BaseServer.verify_request

### DIFF
--- a/mininet/p4_mininet.py
+++ b/mininet/p4_mininet.py
@@ -99,9 +99,12 @@ class P4Switch(Switch):
         while True:
             if not os.path.exists(os.path.join("/proc", str(pid))):
                 return False
-            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            try:
                 sock.settimeout(0.5)
                 result = sock.connect_ex(("localhost", self.thrift_port))
+            finally:
+                sock.close()
             if result == 0:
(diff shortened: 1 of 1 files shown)


mininet/p4_mininet.py
```diff
@@ -99,9 +99,10 @@ def check_switch_started(self, pid):
         while True:
             if not os.path.exists(os.path.join("/proc", str(pid))):
                 return False
-            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
-                sock.settimeout(0.5)
-                result = sock.connect_ex(("localhost", self.thrift_port))
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            sock.settimeout(0.5)
+            result = sock.connect_ex(("localhost", self.thrift_port))
+            sock.close()
```
jafingerhut: I am probably ignorant of the details of the Python `with` statement here, but isn't its purpose to automatically close `sock` when the scope of the `with` block is left, whether by normal execution, or also by an exception being raised that is not caught inside the `with`?  If so, it seems this change causes the new code to close the socket on normal execution, but not if an exception is raised in the 3 lines of code before sock.close()

rst0git: @jafingerhut thank you for pointing this out. I updated the PR (using `try`/`finally`) to close the socket on both normal execution and if an exception is raised.

PR #861, opened 2020-02-02
Add support for optional match type to simple_switch_grpc

antoninbas:
The change required in the PI implementation is very minor. The diff is
large because we are adding a test to simple_switch_grpc.


antoninbas:
This PR is tagged as a draft as it depends on https://github.com/p4lang/PI/pull/504.

PR #864, opened 2020-02-07
psa: Fix passing packet_path value to ingress

rst0git:
When a packet is resubmitted the value of `psa_ingress_input_metadata.packet_path` is  set (incorrectly) to `RECIRCULATE`, instead of `RESUBMIT`.

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -319,7 +319,7 @@ PsaSwitch::ingress_thread() {
     // ingress_timestamp is already set at receive time
     phv->get_field("psa_ingress_input_metadata.ingress_port").set(ingress_port);
     phv->get_field("psa_ingress_input_metadata.packet_path").set(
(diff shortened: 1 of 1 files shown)


targets/psa_switch/psa_switch.cpp
```diff
@@ -319,7 +319,7 @@ PsaSwitch::ingress_thread() {
     // ingress_timestamp is already set at receive time
     phv->get_field("psa_ingress_input_metadata.ingress_port").set(ingress_port);
     phv->get_field("psa_ingress_input_metadata.packet_path").set(
-        phv->get_field("psa_ingress_parser_input_metadata.packet_path"));
+        phv->get_field("psa_ingress_parser_input_metadata.packet_path").get_int());
```
antoninbas: unless I'm missing something, these 2 lines should be equivalent, except that the first one doesn't require a conversion to an integer type

so what is the issue that this is trying to fix?

rst0git: Sorry for the noise. The issue was with other changes I've made to bmv2.

Issue #856, opened 2020-02-01
Enforce parser value set size limit in bmv2


antoninbas:
When declaring a `value_set` in P4_16, one must specify a maximum number of entries. We should include this number in the bmv2 JSON and enforce it, like we do for the match table implementation. If the JSON attribute is missing, we will keep implementing the default behavior.

hesingh:
I didn't want p4c to be behind.  At this public repo, I added "size" to parse_vset JSON.

https://gitlab.com/hemant_mnkcg/p4c-valueset

This is a snippet to generated JSON.

```
"parse_vsets" : [
    {
      "name" : "MyParser.pvs",
      "id" : 0,
      "compressed_bitwidth" : 32,
      "size" : 4
    }
  ],
```

The repo also includes the fix to compute bitwidth for struct used by parse_vset.  Also included is the fix to have parse_vset mask set to null.

PR #865, opened 2020-02-08
Add size to parse_vset doc

hesingh:
Recently, p4c added size to parser value set.  This PR changes the JSON doc to include size.

### DIFF
--- a/docs/JSON_format.md
+++ b/docs/JSON_format.md
@@ -387,6 +387,7 @@ array item has the following attributes:
 - `id`: a unique integer (unique with respect to other parse value-sets)
 - `compressed_bitwidth`: the bitwidth of the values which can be added to the
 set. Note that this bitwidth does not include any padding.
+- `size` - number of elements in the set.
 
 ### `deparsers`
 
--- a/docs/JSON_format.md
+++ b/docs/JSON_format.md
@@ -10,7 +10,7 @@ on each attribute.
 
 ## Current bmv2 JSON format version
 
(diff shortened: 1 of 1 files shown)


docs/JSON_format.md
```diff
@@ -387,6 +387,7 @@ array item has the following attributes:
 - `id`: a unique integer (unique with respect to other parse value-sets)
 - `compressed_bitwidth`: the bitwidth of the values which can be added to the
 set. Note that this bitwidth does not include any padding.
+- `size` - number of elements in the set.
```
antoninbas: comparing it with the rest of the JSON attributes (e.g. for tables) I now wish we had called it `max_size` for consistency; I believe we should still make that change

hesingh: The p4c IR definition for P4ValueSet uses `size`.  So does the P4-16 spec.  However, I see that JSON format uses max_size.  I'd be happy to change p4c to use max_size. 

Regarding "I would rather not merge a PR that only documents the new attribute without implementing support for it"

Are you saying the behavioral model does not support it yet?  If that is correct, I can put this PR on hold.

hesingh: See this p4c PR for maz_size JSON output.

https://github.com/p4lang/p4c/pull/2195

antoninbas: If it supported it already, I would have updated this document as well. There is an open issue for this https://github.com/p4lang/behavioral-model/issues/856

PR #866, opened 2020-02-09
[RFC] psa_switch: Add recirculation/resubmission limit

rst0git:
The PSA specification does not provide a mechanism to prevent a single received packet to continue to recirculate or resubmit indefinitely and implies that targets may impose limits on the number of resubmissions/recirculations. 
 
This PR adds max packet recirculation and resubmission limit with default value set to 255 (unlimited).

Four options are added to `psa_switch_CLI`: `set_max_recirculations`, `set_max_resubmissions`, `get_max_recirculations` and `get_max_resubmissions`. The recirculation/resubmission limit can be changed to a value between 0 and 254, and the special value of 255 (or -1) is used as unlimited.

In addition, two `bit<8>` metadata fields are added to the types `psa_ingress_input_metadata_t` and `psa_egress_input_metadata_t` - `max_recirculations` and `max_resubmissions` - to allow P4 programs to read the current recirculation/resubmission limit set on a target. (https://github.com/rst0git/p4c/tree/psa-max-recirc-resub)

The values of `max_recirculations` and `max_resubmissions` are updated after each recirculation/resubmission to indicate how many more recirculations/resubmissions a packet is allowed to perform. A packet is dropped when the recirculation/resubmission limit has been reached.

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -101,6 +101,8 @@ PsaSwitch::PsaSwitch(bool enable_swap)
   add_required_field("psa_ingress_input_metadata", "packet_path");
   add_required_field("psa_ingress_input_metadata", "ingress_timestamp");
   add_required_field("psa_ingress_input_metadata", "parser_error");
+  add_required_field("psa_ingress_input_metadata", "max_recirculations");
+  add_required_field("psa_ingress_input_metadata", "max_resubmissions");
 
   add_required_field("psa_ingress_output_metadata", "class_of_service");
   add_required_field("psa_ingress_output_metadata", "clone");
@@ -119,6 +121,8 @@ PsaSwitch::PsaSwitch(bool enable_swap)
   add_required_field("psa_egress_input_metadata", "instance");
   add_required_field("psa_egress_input_metadata", "egress_timestamp");
   add_required_field("psa_egress_input_metadata", "parser_error");
+  add_required_field("psa_egress_input_metadata", "max_recirculations");
+  add_required_field("psa_egress_input_metadata", "max_resubmissions");
 
   add_required_field("psa_egress_output_metadata", "clone");
   add_required_field("psa_egress_output_metadata", "clone_session_id");
@@ -172,6 +176,9 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
 
   phv->get_field("psa_ingress_input_metadata.ingress_timestamp").set(
       get_ts().count());
+
+  phv->reset_recirculation_counter();
+  phv->reset_resubmission_counter();
 
   input_buffer.push_front(std::move(packet));
   return 0;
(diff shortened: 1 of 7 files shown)


jafingerhut:
@rst0git Interesting submission.  Out of curiosity, does the p4c PSA back end have enough working functionality that you have found a PSA source program you can compile that exercises the change in behavior you have written here?  That can go a long way to help find bugs in such features.  If not, did you find some other way to test this?

rst0git:
Hi @jafingerhut, I used a simple P4 program to test the resubmission and recirculation changes.

https://github.com/rst0git/psa-examples

There are three examples in this repository, each with one switch and two hosts (h1 and h2).
1. Run `make` in the recirculate (or resubmit) subfolder. This should start a mininet CLI.
2. `mininet> xterm h1 h2`
3. In h2's terminal: `./receive.py`
4. In h1's terminal: `./send.py 10.0.2.2 foo`

For both recirculate and resubmit examples the sent packet will be recirculated/resubmitted indefinitely. When a limit is set using `set_max_resubmissions` or `set_max_recirculations` in `psa_switch_CLI`, the packet will arrive at h2.

jafingerhut:
I have no objections to such a feature in psa_switch, personally.

I view other basic functionality that doesn't work as more interesting to add, but some of that basic functionality may take quite a bit of work and knowledge to figure out how to enable, and require changes to the PSA back end in the p4c compiler.

I do not know if @antoninbas has any thoughts on this kind of functionality in psa_switch or simple_switch -- I believe at the current time both of them let you write an infinite recirculation loop if that is what your P4 program and control plane configuration happens to do, with the idea that "here is the gun, try not to shoot yourself in the foot" kind of way that most programming languages have. :-)

targets/psa_switch/psa_switch.cpp
```diff
@@ -422,6 +461,11 @@ PsaSwitch::egress_thread(size_t worker_id) {
     phv->get_field("psa_egress_input_metadata.parser_error").set(
         packet->get_error_code().get());
 
+    phv->get_field("psa_egress_input_metadata.max_recirculations").set(
```
antoninbas: I am not sure it is a good idea to place these fields into existing standard PSA headers. These headers are well-defined and IMO we should not put stuff there which is not part of the PSA specification.

jafingerhut:
PSA is an architecture that allows a developer to write a P4 program with user-defined metadata fields that are preserved with the packet across before-vs-after a resubmit or recirculation operation.  I think it would be a mistake to create an architecture with operations like resubmit/recirculate that do not support this.

Given that PSA enables this, a developer can definitely write a P4 program that maintains its own resubmit/recirculate count, timestamps, or whatever they want, and their own custom conditions for when a packet has been processed for "too long" and should be dropped, with whatever error counts you want to maintain for the reasons.

If a feature like this is added to psa_switch and/or simple_switch, it definitely seems preferable _not_ to use names that are inside of a standard metadata struct defined by the architecture.  While it is unlikely that this will ever conflict with future additions to the architecture, why risk it when it can be avoided with 100% certainty?  I would defer to Antonin's recommendations on where such data should be recorded about a packet, but one way is using the set_register() and get_register() methods that can access some 'hidden registers' associated with each packet.  You can see some modifications to bmv2 for the v1model architecture that used these for a different purpose here: https://github.com/p4lang/behavioral-model/commit/7cd8d683517739c51ecf807d1919f4f118e349db

PR #903, opened 2020-06-02
Fix undesired reordering in queueing logic

### LINKED ISSUE #900: make check:   FAIL: test_queueing
i install most of libs by using "apt install xxx-dev" ,not from source.And after "make check",only 1 test failed:test_queueing.
this is the result of "cat tests/test-suite.log":

...
FAIL: test_queueing
===================

[==========] Running 6 tests from 5 test cases.
[----------] Global test environment set-up.
[----------] 1 test from QueueingTest/0, where TypeParam = bm::QueueingLogic<std::unique_ptr<int, std::default_delete<int> >, WorkerMapper>
[ RUN      ] QueueingTest/0.ProducerConsummer
[       OK ] QueueingTest/0.ProducerConsummer (110 ms)
[----------] 1 test from QueueingTest/0 (110 ms total)

[----------] 1 test from QueueingTest/1, where TypeParam = bm::QueueingLogicRL<std::unique_ptr<int, std::default_delete<int> >, WorkerMapper>
[ RUN      ] QueueingTest/1.ProducerConsummer
test_queueing.cpp:141: Failure
      Expected: expected_queue_id
      Which is: 2
To be equal to: queue_id
      Which is: 0
...

How to fix it?

antoninbas:
C++ clocks can return the same value in consecutive calls, which was
causing potential reordering in the queueing logic, in particular for
rate-limited queues when the pps rate was set to 0.

This was causing some unit tests to fail.

Fixes #900

### DIFF
--- a/include/bm/bm_sim/queueing.h
+++ b/include/bm/bm_sim/queueing.h
@@ -274,7 +274,8 @@ class QueueingLogicRL {
     auto &w_info = workers_info.at(worker_id);
     if (q_info.size >= q_info.capacity) return 0;
     q_info.last_sent = get_next_tp(q_info);
-    w_info.queue.emplace(item, queue_id, q_info.last_sent);
+    w_info.queue.emplace(
+        item, queue_id, q_info.last_sent, w_info.wrapping_counter++);
     q_info.size++;
     w_info.q_not_empty.notify_one();
     return 1;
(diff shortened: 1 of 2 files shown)


antoninbas:
Even if the counter wraps, it may just create packet re-ordering for a very small time window, which is still strictly better than the current implementation. BTW even now, packet re-ordering is very uncommon. It seems to happen more frequently on some machines or depending on how bmv2 is run (e.g. in a docker container).

PR #904, opened 2020-06-02
Minor wording clarifications on behavior of v1model recirculation

jafingerhut:


jafingerhut:
Someone pointed out that the wording before this change was confusing -- the documentation suggested that recirculated packets did not go through the parser.  This change should make it clear that recirculated packets do go through the parser.

PR #910, opened 2020-06-14
Add notes on timestamp synchronization

jafingerhut:
Proposed addition of notes on synchronizing time stamps in BMv2, to answer those who ask about it.


jafingerhut:
I have changed it to use 'timestamp' consistently, even in English prose.  I believe the only reason that some occurrences are 'timestamp' is to match the field names like `ingress_global_timestamp`, etc.  Apparently in my English OED dictionary 'timestamp' is a word, too, so today I learned.

PR #918, opened 2020-07-01
[psa_switch] supporting meter extern

yunheL:
Supporting psa-meter extern:
1. Added a wrapper for psa-meter to use `bm_sim/meter` implementation
2. Added default `meter_rates` into psa_meter constructor so user can run `meter_execute()` without having set `meter_rates` with runtimeCmd for now.

Has not implemented psa-meter RuntimeCmd yet.

Update 7-11-2020:   

All psa-meter RuntimeCmd implemented:  
- `meter-array-set_rates`  
- `meter_set_rates`  
- `meter_get_rates`  

Color adjustment for PSA:

- bmv2 meter implementation assign higher value for busier flow, see [this](https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/meters.h#L153): GREEN = 0, YELLOW = 1, RED = 2.  
- PSA specification order enums differently, see [this](https://github.com/p4lang/p4c/blob/master/p4include/psa.p4#L649): RED = 0, GREEN = 1, YELLOW = 2.

Added mapping from bmv2-meter color to PSA specification color: (bmv2->psa) 0->1, 1->2, 2->0.

### DIFF
--- a/targets/psa_switch/externs/psa_meter.cpp
+++ b/targets/psa_switch/externs/psa_meter.cpp
@@ -1,0 +1,52 @@
+/* Copyright 2020-present Cornell University
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Yunhe Liu (yunheliu@cs.cornell.edu)
+ *
+ */
+
+#include "psa_meter.h"
+#include <iostream>
+
+namespace bm {
+
+namespace psa {
+
+void
+PSA_Meter::execute(const Data &index, Data &value) {
+    unsigned int color_out = _meter->execute_meter(get_packet(), index.get<size_t>(), (unsigned int) 0);
+    value.set((size_t)color_out);
+}
+
+Meter &
+PSA_Meter::get_Meter(size_t idx) {
+  return _meter->get_meter(idx);
+}
+
+const Meter &
+PSA_Meter::get_Meter(size_t idx) const {
+  return _meter->get_meter(idx);
+}
+
+BM_REGISTER_EXTERN_W_NAME(Meter, PSA_Meter);
+BM_REGISTER_EXTERN_W_NAME_METHOD(Meter, PSA_Meter, execute, const Data &, Data &);
+
+}  // namespace bm::psa
+
+}  // namespace bm
+
+int import_meters(){
+  return 0;
+}
--- a/targets/psa_switch/externs/psa_meter.cpp
+++ b/targets/psa_switch/externs/psa_meter.cpp
@@ -42,7 +42,7 @@ PSA_Meter::init() {
 
 void
 PSA_Meter::execute(const Data &index, Data &value) {
-    unsigned int color_out = _meter->execute_meter(get_packet(), index.get<size_t>(), static_cast<unsigned int>(0));
+    auto color_out = _meter->execute_meter(get_packet(), index.get<size_t>(), static_cast<unsigned int>(0));
 
     // color adjustment for PSA:
     // bmv2 meter implementation assign higher value for busier flow:
@@ -52,15 +52,20 @@ PSA_Meter::execute(const Data &index, Data &value) {
     // (PSA-specification) RED = 0, GREEN = 1, YELLOW = 2.
     // The following code maps color_out (bmv2-meter) to
     // psa_color_out (PSA-specification).
-    unsigned int psa_color_out = 1;
-    if (color_out == 0)
-        psa_color_out = 1;
-    else if (color_out == 1)
-        psa_color_out = 2;
-    else if (color_out == 2)
-        psa_color_out = 0;
+    auto psa_color_out = 1;
+    switch(color_out) {
+        case 0 :
+            psa_color_out = 1;
+            break;
+        case 1 :
+            psa_color_out = 2;
+            break;
+        case 2 :
+            psa_color_out = 0;
+            break;
+    }
 
-    value.set(static_cast<size_t>(psa_color_out));
+    value.set(psa_color_out);
 }
 
 Meter &
--- a/targets/psa_switch/externs/psa_meter.cpp
+++ b/targets/psa_switch/externs/psa_meter.cpp
@@ -52,7 +52,7 @@ PSA_Meter::execute(const Data &index, Data &value) {
     // (PSA-specification) RED = 0, GREEN = 1, YELLOW = 2.
     // The following code maps color_out (bmv2-meter) to
     // psa_color_out (PSA-specification).
-    auto psa_color_out = 1;
+    bm::Meter::color_t psa_color_out = 1;
     switch(color_out) {
         case 0 :
             psa_color_out = 1;
--- a/targets/psa_switch/externs/psa_meter.cpp
+++ b/targets/psa_switch/externs/psa_meter.cpp
@@ -56,6 +56,11 @@ PSA_Meter::get_meter(size_t idx) const {
   return _meter->get_meter(idx);
 }
 
+Meter::MeterErrorCode
+PSA_Meter::set_rates(const std::vector<Meter::rate_config_t> &configs) {
+    return _meter->set_rates(configs);
+}
+
 BM_REGISTER_EXTERN_W_NAME(Meter, PSA_Meter);
 BM_REGISTER_EXTERN_W_NAME_METHOD(Meter, PSA_Meter, execute, const Data &, Data &);
 
--- a/targets/psa_switch/externs/psa_meter.cpp
+++ b/targets/psa_switch/externs/psa_meter.cpp
@@ -52,7 +52,7 @@ PSA_Meter::execute(const Data &index, Data &value) {
     // (PSA-specification) RED = 0, GREEN = 1, YELLOW = 2.
     // The following code maps color_out (bmv2-meter) to
     // psa_color_out (PSA-specification).
-    unsigned int psa_color_out;
+    unsigned int psa_color_out = 1;
     if (color_out == 0)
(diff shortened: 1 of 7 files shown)


targets/psa_switch/externs/psa_meter.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "psa_meter.h"
+#include <iostream>
+
+namespace bm {
+
+namespace psa {
+
+void
+PSA_Meter::execute(const Data &index, Data &value) {
+    unsigned int color_out = _meter->execute_meter(get_packet(), index.get<size_t>(), (unsigned int) 0);
```
hanw, line 29: meter pre-color is left to future work?

antoninbas, line 29: please use `static_cast` instead of C-style casts. I'm surprised the style checker has not complained (maybe it's not enabled for this directory).

same below.

targets/psa_switch/externs/psa_meter.h
```diff
@@ -0,0 +1,90 @@
...
+
+        // Default trTriColor meter rates
+        // 2 packets per second, burst size of 3
+        Meter::rate_config_t committed_rate = {0.000002, 3};
+        // 10 packets per second, burst size of 1
+        Meter::rate_config_t peak_rate = {0.00001, 1};
+        Meter::MeterErrorCode error = _meter->set_rates({committed_rate, peak_rate});
+        if (error != bm::MeterArray::MeterErrorCode::SUCCESS) {
+            // TODO return error handling
+        }
```
antoninbas, line 67: indentation is wrong

also, not needed. If no rates are configured, execute will always return "green", which I believe is expected

These default values you have picked look very arbitrary

targets/psa_switch/externs/psa_meter.h
```diff
@@ -0,0 +1,90 @@
...
+  void init() override {
+    bm::MeterArray::MeterType meter_type;
+    if (type == "bytes") {
+        meter_type = bm::MeterArray::MeterType::BYTES;
+    } else if (type == "packets") {
+        meter_type = bm::MeterArray::MeterType::PACKETS;
+    } else {
+        // TODO: Here Meter is made default as a BYTES meter.
+        // However, what really should been done here is error handling.
+        meter_type = bm::MeterArray::MeterType::BYTES;
```
antoninbas, line 50: I don't think this should happen, but you have access to the logger here, so at least you should log an error

targets/psa_switch/externs/psa_meter.h
```diff
@@ -0,0 +1,90 @@
...
+  static constexpr p4object_id_t spec_id = 0xfffffffe;
+
+  BM_EXTERN_ATTRIBUTES {
+    BM_EXTERN_ATTRIBUTE_ADD(n_meters);
+    BM_EXTERN_ATTRIBUTE_ADD(type);
+    BM_EXTERN_ATTRIBUTE_ADD(is_direct);
+    BM_EXTERN_ATTRIBUTE_ADD(rate_count);
+  }
+  
+  void init() override {
```
antoninbas, line 41: looks like this definition could go into the .cpp file

targets/psa_switch/externs/psa_meter.cpp
yunheL: Yes, pre-color is not included in this PR.

targets/psa_switch/externs/psa_meter.h
yunheL: Thanks for the comments! All make sense, working on the changes! 

> > Has not implemented psa-meter RuntimeCmd yet.
> 
> You only need to implement the runtime interface I showed you last time, which should be a couple lines of code, in order for this to work. Then you will be able to write an STF test in the compiler.

For runtime interface, please correct me if I think wrong, I currently see `meter_array_set_rates`, `meter_set_rates` and `meter_get_rates` in [switch.h](https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/switch.h#L718), I will add corresponding runtime interfaces into [psa_switch.h](https://github.com/p4lang/behavioral-model/blob/master/targets/psa_switch/psa_switch.h#L114).

antoninbas: correct

yunheL: @antoninbas 
- RuntimeCmd Interfaces added (`meter_array_set_rates`, `meter_set_rates`, `meter_get_rates`). 
- Other change request updated as well. 

Would you please take a look and let me know if any further change is needed? Otherwise, I think this PR is ready to be merged? Thanks much!

targets/psa_switch/externs/psa_meter.cpp
```diff
@@ -0,0 +1,90 @@
...
+        new MeterArray(get_name() + ".$impl",
+                         spec_id,
+                         meter_type,
+                         rate_count.get<size_t>(),
+                         n_meters.get<size_t>()));
+}
+
+void
+PSA_Meter::execute(const Data &index, Data &value) {
+    unsigned int color_out = _meter->execute_meter(get_packet(), index.get<size_t>(), static_cast<unsigned int>(0));
```
antoninbas, line 45: the return type of `execute_meter` is `color_t`, I would recommend just using the C++11 `auto` keyword

targets/psa_switch/externs/psa_meter.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // (PSA-specification) RED = 0, GREEN = 1, YELLOW = 2.
+    // The following code maps color_out (bmv2-meter) to
+    // psa_color_out (PSA-specification).
+    unsigned int psa_color_out = 1;
+    if (color_out == 0)
+        psa_color_out = 1;
+    else if (color_out == 1)
+        psa_color_out = 2;
+    else if (color_out == 2)
+        psa_color_out = 0;
```
antoninbas, line 61: this would look more natural as a `switch` statement

targets/psa_switch/externs/psa_meter.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    value.set(static_cast<size_t>(psa_color_out));
```
antoninbas, line 63: what purpose does the `static_cast` serve here? `set` works with any integral type

tools/runtime_CLI.py
```diff
@@ -332,16 +332,42 @@ def get_json_key(key):
...
+                if name == "is_direct":
+                    meter_array.is_direct = value == True
+                # TODO set meter_array.binding for direct_meter
+                # direct_meter not supported on PSA yet
+                elif name == "n_meters":
+                    meter_array.size = value
+                elif name == "type":
+                    meter_array.type_ = value
+                elif name == "rate_count":
+                    meter_array.rate_count = value
```
antoninbas, line 358: architecture-specific extern stuff belongs in the CLI extension file. For PSA it is: https://github.com/p4lang/behavioral-model/blob/master/targets/psa_switch/pswitch_CLI.py

I suggest adding this logic after this line: https://github.com/p4lang/behavioral-model/blob/master/targets/psa_switch/pswitch_CLI.py#L91 and instantiating `MeterArray` there as appropriate

targets/psa_switch/externs/psa_meter.cpp
yunheL: I did not notice `set` works with all integral type, I will remove the static_cast, thanks!

tools/runtime_CLI.py
yunheL: This makes a lot of sense, I was thinking that PSA and v1model shouldn't just share `tools/runtime_CLI.py`, thanks

yunheL: @antoninbas

All fixed:  
* Moved psa-meter json parsing to `pswitch_CLI.py`
* Minor updates on coding style

Would you please take a look and let me know if any further change is needed? Thanks!

targets/psa_switch/pswitch_CLI.py
```diff
@@ -88,7 +88,43 @@ def main():
...
+        meter_array = runtime_CLI.MeterArray(j_meter["name"], j_meter["id"])
+        attribute_values = j_meter.get("attribute_values", [])
+        for attr in attribute_values:
+            name = attr.get("name", [])
+            val_type = attr.get("type", [])
+            value = attr.get("value", [])
+            if name == "is_direct":
+                meter_array.is_direct = value == True
+            # TODO set meter_array.binding for direct_meter
+            # direct_meter not supported on PSA yet
```
antoninbas, line 107: I think you mean "not supported by psa_switch yet", PSA implies the architecture :)

targets/psa_switch/pswitch_CLI.py
```diff
@@ -88,7 +88,43 @@ def main():
         args.thrift_ip, args.thrift_port, services
     )
 
-    runtime_CLI.load_json_config(standard_client, args.json)
+    json_ = runtime_CLI.load_json_config(standard_client, args.json)
```
antoninbas, line 91: I think that's the idea, but as for the implementation would you mind just passing a function to `load_json_config` to perform the target-specific work, so that you don't end up repeating the `SUFFIX_LOOKUP_MAP` stuff below. Something like:
```python
def load_json_psa(json_dict):
    for j_meter in get_json_key("extern_instances"):
        # ...

def main():
    runtime_CLI.load_json_config(standard_client, args.json, load_json_psa)
    # ...
```
Then the runtime_CLI.py code will invoke `load_json_psa` at the right time. You should make this an optional parameter (and default it to `None`) so that you don't need to do any modification for simple_switch.

Let me know what you think

targets/psa_switch/externs/psa_meter.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    // color adjustment for PSA:
+    // bmv2 meter implementation assign higher value for busier flow:
+    // (bmv2-meter) GREEN = 0, YELLOW = 1, RED = 2.
+    // PSA specification order enums differently:
+    // (see p4c/p4include/psa.p4)
+    // (PSA-specification) RED = 0, GREEN = 1, YELLOW = 2.
+    // The following code maps color_out (bmv2-meter) to
+    // psa_color_out (PSA-specification).
+    auto psa_color_out = 1;
```
antoninbas, line 55:
```suggestion
    bm::Meter::color_t psa_color_out = 1;
```

targets/psa_switch/pswitch_CLI.py
yunheL: This makes sense. I just made change accordingly. :D

PR #927, opened 2020-07-22
Add debug for checksum failure

hesingh:
During work on https://github.com/p4lang/p4c/pull/2474, I noticed a discrepancy between the TCP checksum computed by latest Scapy and the behavioral-model. 

This minor change prints the checksum so that when the subsequent checksum debugged message (shown below) is printed and checksum has failed, the new message of this PR has the failed checksum handy.  

`[18:56:19.264] [bmv2] [D] [thread 947] [0.0] [cxt 0] Verifying checksum 'cksum_3': false`

With this PR code changes, this is the  log

```
[18:56:19.264] [bmv2] [D] [thread 947] [0.0] [cxt 0] Checksum is 'cksum_3': 0x4796
[18:56:19.264] [bmv2] [D] [thread 947] [0.0] [cxt 0] Verifying checksum 'cksum_3': false
```

### DIFF
--- a/src/bm_sim/calculations.cpp
+++ b/src/bm_sim/calculations.cpp
@@ -310,12 +310,6 @@ struct crcCCITT {
 
 struct cksum16 {
   uint16_t operator()(const char *buf, size_t len) const {
-    BMLOG_DEBUG("Printing Data: {}", len);
-#ifdef BM_LOG_DEBUG_ON
-    for (size_t i = 0; i < len; i++)
-      std::printf("%02X", buf[i]);
-#endif
-    BMLOG_DEBUG("\nEnd\n");
     uint64_t sum = 0;
     const uint64_t *b = reinterpret_cast<const uint64_t *>(buf);
     uint32_t t1, t2;
--- a/src/bm_sim/calculations.cpp
+++ b/src/bm_sim/calculations.cpp
@@ -31,7 +31,6 @@
 #include <ostream>
 #include <string>
 #include <unordered_map>
-#include <iostream>
 
 #include "xxhash.h"
 #include "crc_tables.h"
@@ -311,12 +310,12 @@ struct crcCCITT {
 
 struct cksum16 {
   uint16_t operator()(const char *buf, size_t len) const {
+    BMLOG_DEBUG("Printing Data: {}", len);
 #ifdef BM_LOG_DEBUG_ON
-    std::cout << "Printing data: " << len << std::endl;
     for (size_t i = 0; i < len; i++)
-      fprintf(stderr, "%02X", buf[i]);
-    std::cout << "\nEnd\n";
+      std::printf("%02X", buf[i]);
 #endif
+    BMLOG_DEBUG("\nEnd\n");
     uint64_t sum = 0;
     const uint64_t *b = reinterpret_cast<const uint64_t *>(buf);
     uint32_t t1, t2;
--- a/src/bm_sim/calculations.cpp
+++ b/src/bm_sim/calculations.cpp
@@ -31,6 +31,7 @@
 #include <ostream>
 #include <string>
 #include <unordered_map>
+#include <iostream>
 
 #include "xxhash.h"
 #include "crc_tables.h"
@@ -310,6 +311,12 @@ struct crcCCITT {
 
 struct cksum16 {
   uint16_t operator()(const char *buf, size_t len) const {
+#ifdef BM_LOG_DEBUG_ON
+    std::cout << "Printing data: " << len << std::endl;
+    for (size_t i = 0; i < len; i++)
+      fprintf(stderr, "%02X", buf[i]);
+    std::cout << "\nEnd\n";
+#endif
     uint64_t sum = 0;
     const uint64_t *b = reinterpret_cast<const uint64_t *>(buf);
     uint32_t t1, t2;
--- a/src/bm_sim/checksums.cpp
+++ b/src/bm_sim/checksums.cpp
@@ -149,6 +149,9 @@ CalcBasedChecksum::update_(Packet *pkt) const {
 bool
 CalcBasedChecksum::verify_(const Packet &pkt) const {
   const uint64_t cksum = calculation->output(pkt);
+  std::stringstream ss;
+  ss << std::hex << cksum;
+  BMLOG_DEBUG_PKT(pkt, "Checksum is '{}': {}", get_name(), "0x"+ ss.str());
   const auto &f_cksum = pkt.get_phv()->get_field(header_id, field_offset);
   return (cksum == f_cksum.get<uint64_t>());
 }
--- a/src/bm_sim/checksums.cpp
+++ b/src/bm_sim/checksums.cpp
@@ -148,21 +148,21 @@ CalcBasedChecksum::update_(Packet *pkt) const {
   f_cksum.set(cksum);
 }
 
-#ifdef BM_LOG_DEBUG_ON
 std::string
 CalcBasedChecksum::convert(uint64_t val) const {
     std::stringstream ss;
     ss << std::hex << val;
     return ("0x" + ss.str());
 }
-#endif
 
 bool
 CalcBasedChecksum::verify_(const Packet &pkt) const {
   const uint64_t cksum = calculation->output(pkt);
   const auto &f_cksum = pkt.get_phv()->get_field(header_id, field_offset);
+#ifdef BM_LOG_DEBUG_ON
   BMLOG_DEBUG_PKT(pkt, "Checksum '{}': computed {} - actual {}",
                   get_name(), convert(cksum), convert(f_cksum.get<uint64_t>()));
+#endif
   return (cksum == f_cksum.get<uint64_t>());
 }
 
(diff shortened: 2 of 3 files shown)


hesingh:
@antoninbas Please review - thanks.

jafingerhut:
If a change like this is desirable, would it be worth going the extra step of printing the calculated checksum, and also the expected value, in the same log message?

hesingh:
With latest code changes, this is a sample log message.

```
[21:50:59.948] [bmv2] [D] [thread 13419] [0.0] [cxt 0] Checksum 'cksum_3': computed 0x4796 : in pkt 0x47a6
[21:50:59.948] [bmv2] [D] [thread 13419] [0.0] [cxt 0] Verifying checksum 'cksum_3': false
```

src/bm_sim/checksums.cpp
```diff
@@ -149,7 +149,12 @@ CalcBasedChecksum::update_(Packet *pkt) const {
 bool
 CalcBasedChecksum::verify_(const Packet &pkt) const {
   const uint64_t cksum = calculation->output(pkt);
+  std::stringstream ss, ess;
+  ss << std::hex << cksum;
   const auto &f_cksum = pkt.get_phv()->get_field(header_id, field_offset);
+  ess << std::hex << f_cksum.get<uint64_t>();
+  BMLOG_DEBUG_PKT(pkt, "Checksum '{}': computed {} : in pkt {}",
```
antoninbas, line 156: please do the conversion from `uint64` to hex string in a separate helper function. And make sure the function is only called when invoking the `BMLOG_DEBUG_PKT`. I want to make sure that the function is not evaluated when the debugging macros are turned off, even though the compiler should take care of that in O2.

src/bm_sim/checksums.cpp
```diff
@@ -149,7 +149,12 @@ CalcBasedChecksum::update_(Packet *pkt) const {
...
+                  get_name(), "0x"+ ss.str(), "0x"+ ess.str());
```
antoninbas, line 157: you can use `ss << "0x" << std::setfill('0') << std::setw(8) << std::hex << <value>`

src/bm_sim/checksums.cpp
antoninbas, line 156: I don't understand the use of the ":" as the second separator here. Maybe use "-". Also do you think "actual" would be more familiar than "in pkt" to people?

hesingh:
> When I see this:
> 
> ```
> [21:50:59.948] [bmv2] [D] [thread 13419] [0.0] [cxt 0] Checksum 'cksum_3': computed 0x4796 : in pkt 0x47a6
> [21:50:59.948] [bmv2] [D] [thread 13419] [0.0] [cxt 0] Verifying checksum 'cksum_3': false
> ```
> 
> The order of the log messages is a bit surprising. With the addition of this new log message, maybe replace "Verifying" with "Verified" in the second log message.

Ok. Incidentally this message prints "false" when checksum comparison has failed which is confusing.  A better term is "passed" or "failed".

hesingh:
Interesting vagary about behavioral-model tools.  in `behavioral-model/tools/` path there is `cpplin.py` which I ran.  On running I was asked to change `BM_BM_SIM_CHECKSUMS_H_` to `INCLUDE_BM_BM_SIM_CHECKSUMS_H_`.  However, the same directory has `check_style.sh` which says remove the "INCLUDE_" from the header guard.  If cppliny.py is not used, it should be removed from the repo.

hesingh:
Having worked with @sarashakeri who had file https://github.com/p4lang/p4c/issues/2472, there is no issue with TCP checksum with simple_switch.

Yesterday, I added verify_checksum_with_payload() to my p4c [PR ](https://github.com/p4lang/p4c/pull/2474) and had @sarashakeri use the code to test.  She also used the code from this PR for her simple_switch and sent me input packets.  I saw her packet included 20 bytes of TCP options which her P4 program has no parser for, nor are the TCP options added to verify and update checksums. This is why the simple_switch failed checksum.

I think we could still consider this PR to debug any futures issues reported by new folks.

I will look into adding TCP options to my p4c PR to test verify and update checksums.

hesingh:
Incidentally in this p4c [issue ](https://github.com/p4lang/p4c/issues/2472) against p4c @jafingerhut had mentioned using Wireshark to verify checksum of a packet.

Having worked on TCP checksum during the past two days, I see the simple_switch with this PR can also verify TCP checksum and so can Scapy.  Just use `pkt.show2()` in Scapy to see the checksum.

```bash
p=Ether(dst='02:74:BF:C3:55:2F', src='8A:3A:18:86:86:AB')/IP(src='192.168.41.2', dst='192.168.43.2')/TCP(sport=51338, dport=5201)/Raw("deadbeefdeadbeef")
>>> p.show2()
b'\xc8\x8a\x14Q\x00\x00\x00\x00\x00\x00\x00\x00P\x02 \x00\x00\x00\x00\x00deadbeefdeadbeef'
###[ Ethernet ]### 
  dst= 02:74:bf:c3:55:2f
  src= 8a:3a:18:86:86:ab
  type= IPv4
###[ IP ]### 
     version= 4
     ihl= 5
     tos= 0x0
     len= 56
     id= 1
     flags= 
     frag= 0
     ttl= 64
     proto= tcp
     chksum= 0xa56a
     src= 192.168.41.2
     dst= 192.168.43.2
     \options\
###[ TCP ]### 
        sport= 51338
        dport= 5201
        seq= 0
        ack= 0
        dataofs= 5
        reserved= 0
        flags= S
        window= 8192
        chksum= 0xc276
        urgptr= 0
        options= []
###[ Raw ]### 
           load= 'deadbeefdeadbeef'


```

src/bm_sim/checksums.cpp
```diff
@@ -146,10 +148,21 @@ CalcBasedChecksum::update_(Packet *pkt) const {
...
+    std::stringstream ss;
+    ss << std::hex << val;
+    return ("0x" + ss.str());
+}
+
 bool
 CalcBasedChecksum::verify_(const Packet &pkt) const {
   const uint64_t cksum = calculation->output(pkt);
   const auto &f_cksum = pkt.get_phv()->get_field(header_id, field_offset);
+#ifdef BM_LOG_DEBUG_ON
```
antoninbas, line 162: this `#ifdef` guard is not needed. If it is not defined, `BMLOG_DEBUG_PKT` will be a no-op and parameters won't be evaluated.

src/bm_sim/calculations.cpp
```diff
@@ -310,6 +311,12 @@ struct crcCCITT {
 
 struct cksum16 {
   uint16_t operator()(const char *buf, size_t len) const {
+#ifdef BM_LOG_DEBUG_ON
+    std::cout << "Printing data: " << len << std::endl;
+    for (size_t i = 0; i < len; i++)
+      fprintf(stderr, "%02X", buf[i]);
+    std::cout << "\nEnd\n";
+#endif
```
antoninbas, line 319: remove?

this is not helpful to users, not to mention you are just outputting to `std::cout`

hesingh: Would be good to not remove.

An IPv6 layer-4 checksum P4 example in p4c is still pending to test with simple_switch.  Further, there are 4-5 different ways to write P4 code for layer-4 options.  If a discrepancy is found in checksum,  a user would like to see the data simple_switch used to computer the checksum.  I will change the `std::cout` to qualified debug.

Is there a preferred way to enable BM_LOG_DEBUG_ON in behavioral-mode build?  First thought I had was to add 
`-DBM_LOG_DEBUG_ON` to Makefile `DEFS`

src/bm_sim/calculations.cpp
```diff
@@ -310,6 +310,12 @@ struct crcCCITT {
 
 struct cksum16 {
   uint16_t operator()(const char *buf, size_t len) const {
+    BMLOG_DEBUG("Printing Data: {}", len);
+#ifdef BM_LOG_DEBUG_ON
+    for (size_t i = 0; i < len; i++)
+      std::printf("%02X", buf[i]);
+#endif
+    BMLOG_DEBUG("\nEnd\n");
```
antoninbas, line 318: I am sorry but I do not think this is a good addition: if the packet is 9,000 bytes and we compute the checksum over the whole payload, are you really going to print the whole buffer?

This is also irrelevant to the typical bmv2 user. Unless there is a bug in p4c or bmv2, then printing this buffer serves no real purpose and clutters the logs.

src/bm_sim/checksums.cpp
```diff
@@ -26,11 +26,19 @@
...
 
 namespace bm {
 
 namespace {
 
+std::string
+convert(const uint64_t val) {
+    std::stringstream ss;
+    ss << "0x" << std::hex << val;
+    return (ss.str());
```
antoninbas, line 39: surrounding parentheses are not needed

src/bm_sim/checksums.cpp
```diff
@@ -26,11 +26,19 @@
...
 #include <cstdint>
 #include <string>
+#include <sstream>
 
 namespace bm {
 
 namespace {
 
+std::string
+convert(const uint64_t val) {
```
antoninbas, line 36: can we have a less generic name for this function? Maybe `convertU64ToHexStr`?

PR #931, opened 2020-07-28
[psa_switch] supporting extern psa-random

yunheL:
Supporting extern `psa-random`: 

Used a random function that is almost identical to an existing bmv2 method [modify_field_rng_uniform()](https://github.com/p4lang/behavioral-model/blob/master/targets/psa_switch/primitives.cpp#L45). The only difference is that `modify_field_rng_uniform()` uses `uint_64()` for range(min, max) while this extern use `int` for range(min, max) during random number generation.

### DIFF
--- a/targets/psa_switch/Makefile.am
+++ b/targets/psa_switch/Makefile.am
@@ -8,7 +8,12 @@ THRIFT_IDL = $(srcdir)/thrift/psa_switch.thrift
 
 noinst_LTLIBRARIES = libpsaswitch.la
 
-libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp externs/psa_random.h externs/psa_random.cpp
+libpsaswitch_la_SOURCES = \
+psa_switch.cpp psa_switch.h \
+primitives.cpp \
+externs/psa_counter.h externs/psa_counter.cpp \
+externs/psa_meter.h externs/psa_meter.cpp \
+externs/psa_random.h externs/psa_random.cpp
 
 libpsaswitch_la_LIBADD = \
 $(top_builddir)/src/bm_sim/libbmsim.la \
--- a/targets/psa_switch/Makefile.am
+++ b/targets/psa_switch/Makefile.am
@@ -8,7 +8,7 @@ THRIFT_IDL = $(srcdir)/thrift/psa_switch.thrift
 
 noinst_LTLIBRARIES = libpsaswitch.la
 
-libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp
+libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp externs/psa_random.h externs/psa_random.cpp
 
 libpsaswitch_la_LIBADD = \
 $(top_builddir)/src/bm_sim/libbmsim.la \
--- a/targets/psa_switch/externs/psa_random.cpp
+++ b/targets/psa_switch/externs/psa_random.cpp
@@ -18,30 +18,45 @@
  */
 
 #include "psa_random.h"
-#include <bm/bm_sim/_assert.h>
+#include <bm/bm_sim/logger.h>
 
 #include <random>
 #include <thread>
+#include <iostream>
 
 namespace bm {
 
 namespace psa {
 
 void
+PSA_Random::init() {
+  valid_range = true;
+  min_val = min.get_uint64();
+  max_val = max.get_uint64();
+  range = max_val - min_val + 1;
+
+  if (range <= 0) {
+    valid_range = false;
+    Logger::get()->warn("Error: PSA extern random range must be positive.");
+  }
+
+  if ((range & (range - 1)) != 0) {
+    valid_range = false;
+    Logger::get()->warn("Error: PSA extern random range must be a power of 2.");
+  }
+}
+
+void
 PSA_Random::read(Data &value) {
-    int min_val = min.get_int();
-    int max_val = max.get_int();
+  if (!valid_range)
+    return;
 
-    int range = max_val - min_val + 1;
-    _BM_ASSERT((range > 0) && "[Error] Random number range must be positive.");
-    _BM_ASSERT(((range & (range - 1)) == 0) && "[Error] Random number range must be a power of 2.");
-
-    using engine = std::default_random_engine;
-    using hash = std::hash<std::thread::id>;
-    static thread_local engine generator(hash()(std::this_thread::get_id()));
-    using distrib64 = std::uniform_int_distribution<uint64_t>;
-    distrib64 distribution(min_val, max_val);
-    value.set(distribution(generator));
+  using engine = std::default_random_engine;
+  using hash = std::hash<std::thread::id>;
+  static thread_local engine generator(hash()(std::this_thread::get_id()));
+  using distrib64 = std::uniform_int_distribution<uint64_t>;
+  distrib64 distribution(min_val, max_val);
+  value.set(distribution(generator));
 }
 
 BM_REGISTER_EXTERN_W_NAME(Random, PSA_Random);
--- a/targets/psa_switch/externs/psa_random.cpp
+++ b/targets/psa_switch/externs/psa_random.cpp
@@ -35,15 +35,18 @@ PSA_Random::init() {
   max_val = max.get_uint64();
   range = max_val - min_val + 1;
 
-  if (range <= 0) {
+  if (min_val > max_val) {
     valid_range = false;
     Logger::get()->warn("Error: PSA extern random range must be positive.");
   }
 
+  /* Even though PSA spec mentioned range should be a power of 2 for
+   * max portability, bmv2 does not have to impose this restriction.
   if ((range & (range - 1)) != 0) {
     valid_range = false;
     Logger::get()->warn("Error: PSA extern random range must be a power of 2.");
   }
+  */
 }
 
 void
--- a/targets/psa_switch/externs/psa_random.cpp
+++ b/targets/psa_switch/externs/psa_random.cpp
@@ -1,0 +1,56 @@
+/* Copyright 2020-present Cornell University
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


targets/psa_switch/externs/psa_random.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace psa {
+
+void
+PSA_Random::read(Data &value) {
+    int min_val = min.get_int();
+    int max_val = max.get_int();
+
+    int range = max_val - min_val + 1;
+    _BM_ASSERT((range > 0) && "[Error] Random number range must be positive.");
+    _BM_ASSERT(((range & (range - 1)) == 0) && "[Error] Random number range must be a power of 2.");
```
yunheL, line 37: Not sure whether this is the proper way to to error reporting. Suggestion appreicated!

targets/psa_switch/pswitch_CLI.py
```diff
@@ -101,6 +123,34 @@ def get_json_key(key):
...
+        for name, res in res_dict.items():
+            suffix = None
+            for s in reversed(name.split('.')):
+                suffix = s if suffix is None else s + '.' + suffix
+                key = (res_type, suffix)
+                runtime_CLI.SUFFIX_LOOKUP_MAP[key] = res
+                suffix_count[key] += 1
+    for key, c in suffix_count.items():
+        if c > 1:
+            del runtime_CLI.SUFFIX_LOOKUP_MAP[key]
```
yunheL, line 152: Had to duplicate this piece of code (line 140-150) from [tools/runtime_CLI.py line 380](https://github.com/p4lang/behavioral-model/blob/master/tools/runtime_CLI.py#L380) as I needed to add `RANDOMS` to `SUFFIX_LOOKUP_MAP`.

I could put `RANDOMS` in `tools/runtime_CLI.py` add avoid this duplication, however, I did not because I do not want to put `RANDOMS` (psa-random specific code) to `tools/runtime_CLI.py` which not architecture specific.

Comments on whether this code organizing decision make sense and suggestions are appreciated!

targets/psa_switch/externs/psa_random.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <thread>
+
+namespace bm {
+
+namespace psa {
+
+void
+PSA_Random::read(Data &value) {
+    int min_val = min.get_int();
+    int max_val = max.get_int();
```
antoninbas, line 33: do the conversion to integers in the `init` method so it does not have to be done for every call to `read`

antoninbas, line 33: why limit ourselves to 32-bit integers? do we need to support signed values or can we use `uint64`s?

targets/psa_switch/externs/psa_random.cpp
antoninbas, line 37: The PSA spec reads:
> Implementations are allowed to support only ranges where (max - min + 1) is a power of 2

I don't see a strong reason to have this limitation in the bmv2 implementation of PSA.

targets/psa_switch/pswitch_CLI.py
antoninbas: can we write a function in `tools/runtime_CLI.py` which does that computation for a provided list of `(type, dict)` tuples and use that function in both places?

antoninbas: more importantly, why are you modifying pswitch_CLI.py here? which runtime functions are you planning to support?

targets/psa_switch/externs/psa_random.cpp
antoninbas: I think logging an error with an early return like https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/primitives.cpp#L72 is the way to go

targets/psa_switch/pswitch_CLI.py
yunheL: Haha, you are right, there is no runtime function I am trying to support and I do not need to modify pswitch_CLI.py at all. I will remove this chunk of code.

yunheL: Haha, you are right, there is no runtime function I am trying to support and I do not need to modify pswitch_CLI.py at all. I will remove this chunk of code.

targets/psa_switch/externs/psa_random.cpp
yunheL: This refers to use `int` instead of `uint_64` right? I used `int` as I wasn't sure whether we need to support negative `min` and negative `max`.

jfingerh:
It appears that both v1model and PSA architectures support a type parameter T as the type of the min, max, and returned random value.  I have no idea if v1model was ever intended to support T as a type `int<W>`, and I do not recall the issue being raised in PSA meetings.  Certainly type T should be allowed to be a large variety of `bit<W>` types, up to some maximum value of W, where 64 seems reasonable as a maximum there.

It seems reasonable to me to restrict type T not to include `int<W>`.  I have created an issue to explicitly document the restrictions on type T for the Random extern here: https://github.com/p4lang/p4-spec/issues/882

targets/psa_switch/Makefile.am
```diff
@@ -8,7 +8,7 @@ THRIFT_IDL = $(srcdir)/thrift/psa_switch.thrift
 
 noinst_LTLIBRARIES = libpsaswitch.la
 
-libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp
+libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp externs/psa_random.h externs/psa_random.cpp
```
rst0git, line 11: Could you please split this line into multiple lines?
```
libpsaswitch_la_SOURCES = \
psa_switch.cpp \
psa_switch.h \
primitives.cpp \
externs/psa_counter.h \
externs/psa_counter.cpp \
externs/psa_meter.h \
externs/psa_meter.cpp \
externs/psa_random.h \
externs/psa_random.cpp
```

targets/psa_switch/externs/psa_random.cpp
yunheL: changed to uint64_t

targets/psa_switch/externs/psa_random.cpp
```diff
@@ -18,30 +18,45 @@
...
 namespace psa {
 
+void
+PSA_Random::init() {
+  valid_range = true;
+  min_val = min.get_uint64();
+  max_val = max.get_uint64();
+  range = max_val - min_val + 1;
+
+  if (range <= 0) {
```
jafingerhut, line 38: The only way this condition can be true for `range`, declared as type `uint64`, is if it is equal to 0.

I would recommend replacing this condition with `(min_val > max_val)`

targets/psa_switch/externs/psa_random.cpp
```diff
@@ -18,30 +18,45 @@
...
+    valid_range = false;
+    Logger::get()->warn("Error: PSA extern random range must be positive.");
+  }
+
+  if ((range & (range - 1)) != 0) {
```
jafingerhut, line 43: I know the PSA specification says that implementations are allowed to impose restrictions like this, but is there any particular reason why the BMv2 implementation should?  It seems like without this `if` statement, the rest of the code would handle non-power-of-2 ranges just fine, yes?  If so, I'd recommend removing this `if` statement.

jafingerhut: The reason this restriction is mentioned in the PSA spec at all is that for some hardware targets, they might make such a restriction in order to make the hardware implementation simpler, e.g. generating N random bits, then adding them to min_value.

yunheL: That make sense, I will remove this if statement from bmv2 end :D thanks!

yunheL: Both should be fixed now: changed negative check to `!(min > max)` and removed the restriction that range must be a power of 2.

jafingerhut (approved):
Looks good to me.

targets/psa_switch/externs/psa_random.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace bm {
+
+namespace psa {
+
+void
+PSA_Random::init() {
+  valid_range = true;
+  min_val = min.get_uint64();
+  max_val = max.get_uint64();
+  range = max_val - min_val + 1;
```
antoninbas, line 36: `range` is not really used for anything?

targets/psa_switch/externs/psa_random.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  if (min_val > max_val) {
+    valid_range = false;
+    Logger::get()->warn("Error: PSA extern random range must be positive.");
+  }
```
antoninbas, line 41: Now that the check is in the `init` function, I would actually recommend using `_BM_ASSERT` so that the error is not missed. Then you can remove `valid_range` altogether.

We need a better way to report errors from `init`, but that can be done in the future.

targets/psa_switch/externs/psa_random.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  /* Even though PSA spec mentioned range should be a power of 2 for
```
antoninbas, line 43: I think leave the explanation as a comment but remove the commented-out code.

targets/psa_switch/externs/psa_random.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <bm/bm_sim/extern.h>
+
+namespace bm {
+
+namespace psa {
+
+class PSA_Random : public bm::ExternType {
+ public:
+  static constexpr p4object_id_t spec_id = 0xfffffffd;
```
antoninbas, line 31: is this used?

targets/psa_switch/externs/psa_random.cpp
yunheL: Your are right, I have range removed.

targets/psa_switch/externs/psa_random.cpp
yunheL: Done.

targets/psa_switch/externs/psa_random.h
yunheL: no, it is not and I have it removed 👍

yunheL:
The requests of changes has been resolved and this PR is ready to be reviewed, thanks! :)

yunheL:
Just bumping up this thread: this PR Is ready to be reviewed and merged, thanks!

hanw (approved):
LGTM.

antoninbas (approved):

PR #935, opened 2020-08-17
[psa_switch] first pass at i2e cloning 

peteli3:
As the title says, this is a first attempt at implementing i2e cloning.
It should be reminiscent of ingress cloning in simple_switch.cpp, except that the pre-ingress packet is not parsed out-of-band again because the pre-ingress packet's next step is a (newly) explicit egress parser.

stf test:
https://github.com/p4lang/p4c/pull/2499

### DIFF
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -70,7 +70,49 @@ namespace bm {
 
 namespace psa {
 
+static constexpr uint16_t MAX_MIRROR_SESSION_ID = (1u << 15) - 1;
 packet_id_t PsaSwitch::packet_id = 0;
+
+class PsaSwitch::MirroringSessions {
+ public:
+  bool add_session(mirror_id_t mirror_id,
+                   const MirroringSessionConfig &config) {
+    Lock lock(mutex);
+    if (0 <= mirror_id && mirror_id <= MAX_MIRROR_SESSION_ID) {
+      sessions_map[mirror_id] = config;
+      return true;
+    } else {
+      bm::Logger::get()->error("mirror_id out of range. No session added.");
+      return false;
+    }
+  }
+
+  bool delete_session(mirror_id_t mirror_id) {
+    Lock lock(mutex);
+    if (0 <= mirror_id && mirror_id <= MAX_MIRROR_SESSION_ID) {
+      return sessions_map.erase(mirror_id) == 1;
+    } else {
+      bm::Logger::get()->error("mirror_id out of range. No session deleted.");
+      return false;
+    }
+  }
+
+  bool get_session(mirror_id_t mirror_id,
+                   MirroringSessionConfig *config) const {
+    Lock lock(mutex);
+    auto it = sessions_map.find(mirror_id);
+    if (it == sessions_map.end()) return false;
+    *config = it->second;
+    return true;
+  }
+
+ private:
+  using Mutex = std::mutex;
+  using Lock = std::lock_guard<Mutex>;
+
+  mutable std::mutex mutex;
+  std::unordered_map<mirror_id_t, MirroringSessionConfig> sessions_map;
+};
 
 PsaSwitch::PsaSwitch(bool enable_swap)
   : Switch(enable_swap),
@@ -92,7 +134,8 @@ PsaSwitch::PsaSwitch(bool enable_swap)
         this->transmit_fn(port_num, buffer, len);
     }),
     pre(new McSimplePreLAG()),
-    start(clock::now()) {
+    start(clock::now()),
+    mirroring_sessions(new MirroringSessions()) {
   add_component<McSimplePreLAG>(pre);
 
   add_required_field("psa_ingress_parser_input_metadata", "ingress_port");
@@ -200,6 +243,23 @@ PsaSwitch::reset_target_state_() {
   get_component<McSimplePreLAG>()->reset_state();
 }
 
+bool
+PsaSwitch::mirroring_add_session(mirror_id_t mirror_id,
+                                    const MirroringSessionConfig &config) {
+  return mirroring_sessions->add_session(mirror_id, config);
+}
+
+bool
+PsaSwitch::mirroring_delete_session(mirror_id_t mirror_id) {
+  return mirroring_sessions->delete_session(mirror_id);
+}
+
+bool
+PsaSwitch::mirroring_get_session(mirror_id_t mirror_id,
+                                    MirroringSessionConfig *config) const {
+  return mirroring_sessions->get_session(mirror_id, config);
+}
+
 int
 PsaSwitch::set_egress_queue_depth(size_t port, const size_t depth_pkts) {
   egress_buffers.set_capacity(port, depth_pkts);
@@ -281,6 +341,28 @@ PsaSwitch::enqueue(port_t egress_port, std::unique_ptr<Packet> &&packet) {
 }
 
 void
+PsaSwitch::multicast(Packet *packet, unsigned int mgid, PktInstanceType path) {
+  auto phv = packet->get_phv();
+  const auto pre_out = pre->replicate({mgid});
+  auto &f_instance = phv->get_field("psa_egress_input_metadata.instance");
+  auto &f_packet_path = phv->get_field("psa_egress_parser_input_metadata.packet_path");
+  auto packet_size = packet->get_register(PACKET_LENGTH_REG_IDX);
+  for (const auto &out : pre_out) {
+    auto egress_port = out.egress_port;
+    auto instance = out.rid;
+    BMLOG_DEBUG_PKT(*packet,
+                    "Replicating packet on port {} with instance {}",
+                    egress_port, instance);
+    f_instance.set(instance);
+    // TODO use appropriate enum member from JSON
+    f_packet_path.set(path);
+    std::unique_ptr<Packet> packet_copy = packet->clone_with_phv_ptr();
+    packet_copy->set_register(PACKET_LENGTH_REG_IDX, packet_size);
+    enqueue(egress_port, std::move(packet_copy));
+  }
+}
+
+void
 PsaSwitch::ingress_thread() {
   PHV *phv;
 
@@ -301,10 +383,9 @@ PsaSwitch::ingress_thread() {
        bytes are parsed by the parser ("lifted" into p4 headers). Here, we
        track the buffer_state prior to parsing so that we can put it back
        for packets that are cloned or resubmitted, same as in simple_switch.cpp
-
-       TODO */
-
+    */
     const Packet::buffer_state_t packet_in_state = packet->save_buffer_state();
+    auto ingress_packet_size = packet->get_register(PACKET_LENGTH_REG_IDX);
 
     // The PSA specification says that for all packets, whether they
     // are new ones from a port, or resubmitted, or recirculated, the
@@ -337,17 +418,53 @@ PsaSwitch::ingress_thread() {
     ingress_mau->apply(packet.get());
     packet->reset_exit();
 
-    // prioritize dropping if marked as such - do not move below other checks
-    const auto &f_drop = phv->get_field("psa_ingress_output_metadata.drop");
-    if (f_drop.get_int()) {
+    // ingress cloning - sent in their pre-ingress-thread state to either an egress
+    //                   port or a multicast group
+    //                 - dropped packets should still be cloned - do not move below drop
+    unsigned int clone = phv->get_field("psa_ingress_output_metadata.clone").get_uint();
+    if (clone) {
+      MirroringSessionConfig config;
+      unsigned int clone_session_id = phv->get_field("psa_ingress_output_metadata.clone_session_id").get_uint();
+      bool is_session_configured = mirroring_get_session(
+          static_cast<mirror_id_t>(clone_session_id), &config);
+      BMLOG_DEBUG_PKT(*packet, "Cloning packet at ingress to session id {}", clone_session_id);
+
+      if (is_session_configured) {
+        const Packet::buffer_state_t packet_out_state = packet->save_buffer_state();
+        packet->restore_buffer_state(packet_in_state);
+
+        std::unique_ptr<Packet> packet_copy = packet->clone_no_phv_ptr();
+        packet_copy->set_register(PACKET_LENGTH_REG_IDX, ingress_packet_size);
+        auto phv_copy = packet_copy->get_phv();
+        phv_copy->reset_metadata();
+        phv_copy->get_field("psa_egress_parser_input_metadata.packet_path").set(PACKET_PATH_CLONE_I2E);
+        phv_copy->get_field("psa_egress_input_metadata.packet_path").set(PACKET_PATH_CLONE_I2E);
+
+        if (config.mgid_valid) {
+          BMLOG_DEBUG_PKT(*packet_copy, "Cloning packet to multicast group {}", config.mgid);
+          multicast(packet_copy.get(), config.mgid, PACKET_PATH_CLONE_I2E);
+        }
+
+        if (config.egress_port_valid) {
+          BMLOG_DEBUG_PKT(*packet_copy, "Cloning packet to egress port {}", config.egress_port);
+          enqueue(config.egress_port, std::move(packet_copy));
+        }
+
+        packet->restore_buffer_state(packet_out_state);
+      }
+    }
+
+    // drop - packets marked via the ingress_drop action
+    unsigned int drop = phv->get_field("psa_ingress_output_metadata.drop").get_uint();
+    if (drop) {
       BMLOG_DEBUG_PKT(*packet, "Dropping packet at the end of ingress");
       continue;
     }
 
     // resubmit - these packets get immediately resub'd to ingress, and skip
     //            deparsing, do not move below multicast or deparse
-    const auto &f_resubmit = phv->get_field("psa_ingress_output_metadata.resubmit");
-    if (f_resubmit.get_int()) {
+    unsigned int resubmit = phv->get_field("psa_ingress_output_metadata.resubmit").get_uint();
+    if (resubmit) {
       BMLOG_DEBUG_PKT(*packet, "Resubmitting packet");
 
       packet->restore_buffer_state(packet_in_state);
@@ -366,46 +483,24 @@ PsaSwitch::ingress_thread() {
     const auto &f_ig_cos = phv->get_field("psa_ingress_output_metadata.class_of_service");
     const auto ig_cos = f_ig_cos.get_uint();
 
-    // handling multicast
-    unsigned int mgid = 0u;
-    const auto &f_mgid = phv->get_field("psa_ingress_output_metadata.multicast_group");
-    mgid = f_mgid.get_uint();
-
+    unsigned int mgid = phv->get_field("psa_ingress_output_metadata.multicast_group").get_uint();
     if (mgid != 0) {
       BMLOG_DEBUG_PKT(*packet,
                       "Multicast requested for packet with multicast group {}",
                       mgid);
-      const auto pre_out = pre->replicate({mgid});
-      auto &f_instance = phv->get_field("psa_egress_input_metadata.instance");
-      auto &f_eg_cos = phv->get_field("psa_egress_input_metadata.class_of_service");
-      auto packet_size = packet->get_register(PACKET_LENGTH_REG_IDX);
-      for(const auto &out : pre_out){
-        auto egress_port = out.egress_port;
-        auto instance = out.rid;
-        BMLOG_DEBUG_PKT(*packet,
-                        "Replicating packet on port {} with instance {}",
-                        egress_port, instance);
-        f_instance.set(instance);
-        // TODO use appropriate enum member from JSON
-        f_packet_path.set(PACKET_PATH_NORMAL_MULTICAST);
-        f_eg_cos.set(ig_cos);
-        std::unique_ptr<Packet> packet_copy = packet->clone_with_phv_ptr();
-        packet_copy->set_register(PACKET_LENGTH_REG_IDX, packet_size);
-        enqueue(egress_port, std::move(packet_copy));
-      }
+      multicast(packet.get(), mgid, PACKET_PATH_NORMAL_MULTICAST);
       continue;
     }
 
-    const auto &f_egress_port = phv->get_field("psa_ingress_output_metadata.egress_port");
     auto &f_instance = phv->get_field("psa_egress_input_metadata.instance");
     auto &f_eg_cos = phv->get_field("psa_egress_input_metadata.class_of_service");
-    port_t egress_port = f_egress_port.get_uint();
-    BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
     f_instance.set(0);
     // TODO use appropriate enum member from JSON
-    f_packet_path.set(PACKET_PATH_NORMAL_UNICAST);
     f_eg_cos.set(ig_cos);
 
+    f_packet_path.set(PACKET_PATH_NORMAL_UNICAST);
+    port_t egress_port = phv->get_field("psa_ingress_output_metadata.egress_port").get_uint();
+    BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
     enqueue(egress_port, std::move(packet));
   }
 }
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -194,7 +194,7 @@ PsaSwitch::receive_(port_t port_num, const char *buffer, int len) {
                                bm::PacketBuffer(len + 512, buffer, len));
 
   BMELOG(packet_in, *packet);
-  PHV *phv = packet->get_phv();
+  auto *phv = packet->get_phv();
 
   // many current p4 programs assume this
   // from psa spec - PSA does not mandate initialization of user-defined
@@ -326,7 +326,7 @@ PsaSwitch::enqueue(port_t egress_port, std::unique_ptr<Packet> &&packet) {
     packet->set_egress_port(egress_port);
 
 #ifdef SSWITCH_PRIORITY_QUEUEING_ON
-    size_t priority = phv->has_field(SSWITCH_PRIORITY_QUEUEING_SRC) ?
+    auto priority = phv->has_field(SSWITCH_PRIORITY_QUEUEING_SRC) ?
         phv->get_field(SSWITCH_PRIORITY_QUEUEING_SRC).get<size_t>() : 0u;
     if (priority >= SSWITCH_PRIORITY_QUEUEING_NB_QUEUES) {
       bm::Logger::get()->error("Priority out of range, dropping packet");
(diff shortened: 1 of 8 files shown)


targets/psa_switch/psa_switch.cpp
```diff
@@ -354,31 +372,42 @@ PsaSwitch::ingress_thread() {
...
-        auto egress_port = out.egress_port;
-        BMLOG_DEBUG_PKT(*packet, "Replicating packet on port {}", egress_port);
-        std::unique_ptr<Packet> packet_copy = packet->clone_with_phv_ptr();
-        packet_copy->set_register(PACKET_LENGTH_REG_IDX, packet_size);
-        enqueue(egress_port, std::move(packet_copy));
-      }
+      multicast(packet.get(), mgid);
       continue;
     }
+    // TODO(pete) from psa spec - else if multicast group is empty, drop packet
```
antoninbas, line 381: I am not sure I understand this comment. Why would you drop the packet is the mcast group is empty?

targets/psa_switch/psa_switch.cpp
```diff
@@ -354,31 +372,42 @@ PsaSwitch::ingress_thread() {
...
+
+    // i2e clone
+    unsigned int clone = phv->get_field("psa_ingress_output_metadata.clone").get_uint();
+    unsigned int clone_session_id = phv->get_field("psa_ingress_output_metadata.clone_session_id").get_uint();
+    if (clone && clone_session_id) {
```
antoninbas, line 386: I believe in PSA, 0 is a valid `clone_session_id`. This is why there is the `clone` boolean flag.

targets/psa_switch/psa_switch.cpp
```diff
@@ -354,31 +372,42 @@ PsaSwitch::ingress_thread() {
...
+      BMLOG_DEBUG_PKT(*packet, "Cloning packet");
+
+      const Packet::buffer_state_t packet_out_state = packet->save_buffer_state();
+      packet->restore_buffer_state(packet_in_state);
+
+      std::unique_ptr<Packet> packet_copy = packet->clone_no_phv_ptr();
+      packet_copy->set_register(PACKET_LENGTH_REG_IDX, ingress_packet_size);
```
antoninbas, line 393: I noticed that you are not consuming this value in the egress thread (see simple_switch.cpp)

targets/psa_switch/psa_switch.cpp
```diff
@@ -354,31 +372,42 @@ PsaSwitch::ingress_thread() {
...
 
-    const auto &f_egress_port = phv->get_field("psa_ingress_output_metadata.egress_port");
-    port_t egress_port = f_egress_port.get_uint();
-    BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
+      PHV *copy_phv;
+      copy_phv = packet_copy->get_phv();
```
antoninbas, line 396: I suggest this one liner: `auto copy_phv = packet_copy->get_phv();`

targets/psa_switch/psa_switch.cpp
```diff
@@ -354,31 +372,42 @@ PsaSwitch::ingress_thread() {
...
+      copy_phv->reset_metadata();
+      copy_phv->get_field("psa_egress_parser_input_metadata.packet_path")
+        .set(PACKET_PATH_CLONE_I2E);
+      copy_phv->get_field("psa_egress_input_metadata.packet_path")
+        .set(PACKET_PATH_CLONE_I2E);
 
+      BMLOG_DEBUG_PKT(*packet, "Cloning packet to CSID {}", clone_session_id);
```
antoninbas, line 403: s/CSID/session id

CSID is a confusing acronym IMO

targets/psa_switch/psa_switch.cpp
peteli3: Ah thanks - I had forgotten to ask @jafingerhut about this. It's very possible I'm reading too far into the spec.

There's a paragraph under Packet Cloning in the PSA spec that says clone session configs should work similarly to multicast group configs.
Under Multicast replication, the first paragraph says "Sending a packet to an empty group causes the packet to
be dropped."
https://github.com/p4lang/p4-spec/blob/master/p4-16/psa/PSA.mdk#L899

@antoninbas I don't have a great answer to your question since it is also a question of mine.
@jafingerhut, can you shed some light on whether I'm reading too much into the spec by adding this TODO or if there's a scenario in which it is useful?

targets/psa_switch/psa_switch.cpp
peteli3: Thanks, fixing this in next commit.

targets/psa_switch/psa_switch.cpp
peteli3: Making this change in next commit.

targets/psa_switch/psa_switch.cpp
peteli3: Making this change in next commit.

targets/psa_switch/psa_switch.cpp
peteli3: Closing the loop here with snippet from a slack conversation:
> **Antonin**
> for PSA it seems that the packet length is not exposed through metadata; plus because of the deparsing / parsing step between ingress & egress, there would be another way to access this length

We are going to leave this as is in case the register comes in handy.

targets/psa_switch/psa_switch.cpp
jafingerhut: If the multicast_group field is non-0, then it is possible that the multicast group has been configured with no (egress_port, egress_instance) pairs at all.  If so, then no copies of the packet are made, so it has a similar effect as dropping the packet, in that no packets are enqueued in the packet buffer for any output ports at all.

If a packet should be cloned, because the ostd.clone field is true when ingress is done, then the clone session configuration for clone session numbered ostd.clone_session_id.  That clone session might be configured with no (egress_port, egress_instance) pairs.  If that set of pairs is empty, then no clone packets are created.

Does that answer your question?

jafingerhut: Note that whether clones are created for a packet, is an independent decision from whether the packet is dropped or not.  See the pseudocode for end-of-ingress behavior in section https://p4.org/p4-spec/docs/PSA-v1.1.0.html#sec-after-ingress

and for end-of-egress behavior in section https://p4.org/p4-spec/docs/PSA-v1.1.0.html#sec-after-egress

Feel free to ask any questions about those or any other parts of the PSA spec they are not clear.

antoninbas: That seems pretty easy to implement, we should have the behavior that @jafingerhut is describing from the get go.

antoninbas: Specifically, the current code already implements the correct behavior for empty multicast groups, so the TODO should be removed. However, it seems that more work is required to ensure that the cloning decision can be enforced even when the packet is dropped (that includes when the multicast group is empty).

targets/psa_switch/psa_switch.cpp
```diff
@@ -354,31 +372,42 @@ PsaSwitch::ingress_thread() {
...
+      multicast(packet_copy.get(), clone_session_id);
```
antoninbas, line 404: what is this? That seems very wrong, I don't know how I missed it. Why is the `clone_session_id` used as an mgid here?

I recommend taking a look at the simple_switch implementation again. BTW, in order for P4Runtime to work correctly (but I think that's out-of-scope for now), a cloning session can either point to an egress port or to a multicast group id. Since we are not testing psa_switch with P4Runtime right now, you can ignore the multicast group id case from the simple_switch implementation.

But regardless, multicast & cloning are two different things, and you cannot call `multicast` here.

targets/psa_switch/psa_switch.cpp
antoninbas, line 396: connected to my comment below: you do not seem to be using `copy_phv` for anything here, which is a sure indicator that something is wrong

I also suggest renaming it `phv_copy` for consistency with simple_switch code

targets/psa_switch/psa_switch.cpp
peteli3: Thanks for clarifying - I'll take a look at this a bit later today.

targets/psa_switch/psa_switch.cpp
peteli3: Thanks @jafingerhut and @antoninbas, taking a look at this later today as well.

peteli3:
@antoninbas @jafingerhut I addressed Antonin's comment's from above - and also took a stab at adding support for `mirroring_*` commands in `pswitch_CLI` which were useful for the corresponding STF tests https://github.com/p4lang/p4c/pull/2499

Please review when you get a chance

targets/psa_switch/psa_switch.cpp
```diff
@@ -337,17 +418,53 @@ PsaSwitch::ingress_thread() {
...
+      if (is_session_configured) {
+        const Packet::buffer_state_t packet_out_state = packet->save_buffer_state();
+        packet->restore_buffer_state(packet_in_state);
+
+        std::unique_ptr<Packet> packet_copy = packet->clone_no_phv_ptr();
+        packet_copy->set_register(PACKET_LENGTH_REG_IDX, ingress_packet_size);
+        auto phv_copy = packet_copy->get_phv();
+        phv_copy->reset_metadata();
+        phv_copy->get_field("psa_egress_parser_input_metadata.packet_path").set(PACKET_PATH_CLONE_I2E);
+        phv_copy->get_field("psa_egress_input_metadata.packet_path").set(PACKET_PATH_CLONE_I2E);
```
jfingerh, line 441: This line is redundant and I'd recommend removing it, because egress_thread code always copies the value of psa_egress_parser_input_metadata.packet_path to psa_egress_input_metadata.packet_path for every packet, regardless of the value of packet_path.

targets/psa_switch/psa_switch.cpp
```diff
@@ -366,46 +483,24 @@ PsaSwitch::ingress_thread() {
...
-                        egress_port, instance);
-        f_instance.set(instance);
-        // TODO use appropriate enum member from JSON
-        f_packet_path.set(PACKET_PATH_NORMAL_MULTICAST);
-        f_eg_cos.set(ig_cos);
-        std::unique_ptr<Packet> packet_copy = packet->clone_with_phv_ptr();
-        packet_copy->set_register(PACKET_LENGTH_REG_IDX, packet_size);
-        enqueue(egress_port, std::move(packet_copy));
-      }
+      multicast(packet.get(), mgid, PACKET_PATH_NORMAL_MULTICAST);
```
jfingerh, line 491: Do you know whether these changes are based upon the latest version of the behavioral-model code?   I know I have (unintentionally) made your life a little more difficult here by making changes to this code recently that have been merged in while your work was in progress.

Your proposed multicast function does not include the assignments to the class_of_service metadata field, for example, and there is now a PSA test for normal multicasting of packets (not involving cloning) that will fail if class_of_service is not copied from ingress to egress for such packets, as required by the PSA specification.

For cloned packets, the class_of_service should come from the clone session configuration, but I do not know if that is something that is part of the clone session configuration in your changes right now.  It could be left for a later enhancement if that is easier, preferably with a comment that it should be done later.

Perhaps you could add a new parameter 'egress_cos' to your multicast function, and pass in the value 'ig_cos' when called from here for normal multicast packets, and put in a constant 0 for now when you call multicast from the ingress-to-egress clone code that does multicast?  That would be the place for the TBD comment to change the 0 later.

jfingerh:
@peteli3 This is looking pretty good to me.  Have you built the latest behavioral-model plus these changes on your machine, plus latest p4c plus your p4c PR's new test case, and tried running all p4c tests to see which pass and which fail?

It is OK if p4c tests with file names beginning with 'ebpf' or 'ubpf' fail (at least, I have yet to figure out how to build p4c so that they pass), but the rest should all pass before we make changes to behavioral-model repo.

jfingerh:
@antoninbas I do not know whether I am correct here in my belief, or even if I am, whether this PR needs to address it before we merge it in, since it can be improved later if it is an issue.

I believe the basic idea of these changes is that every clone session can either have an egress_port configured for it, or a multicast group id.

Later when we want to implement the P4Runtime API for PSA on bmv2, IIRC every clone session should have its own independent set of (egress_port, instance) pairs, and those should be numbered by the clone session id, independent of any multicast group ids configured in the device.

Does that sound correct to you?  And if it does, then at least a little bit of this psa_switch code will need a few changes to accomodate that?

targets/psa_switch/psa_switch.cpp
```diff
@@ -363,49 +486,25 @@ PsaSwitch::ingress_thread() {
...
     auto &f_eg_cos = phv->get_field("psa_egress_input_metadata.class_of_service");
-    port_t egress_port = f_egress_port.get_uint();
-    BMLOG_DEBUG_PKT(*packet, "Egress port is {}", egress_port);
     f_instance.set(0);
     // TODO use appropriate enum member from JSON
-    f_packet_path.set(PACKET_PATH_NORMAL_UNICAST);
     f_eg_cos.set(ig_cos);
 
+    f_packet_path.set(PACKET_PATH_NORMAL_UNICAST);
+    port_t egress_port = phv->get_field("psa_ingress_output_metadata.egress_port").get_uint();
```
antoninbas, line 506: style suggestion:
```cpp
auto egress_port = phv->get_field("psa_ingress_output_metadata.egress_port").get<port_t>();
```

targets/psa_switch/psa_switch.cpp
```diff
@@ -337,17 +420,57 @@ PsaSwitch::ingress_thread() {
...
     packet->reset_exit();
 
-    // prioritize dropping if marked as such - do not move below other checks
-    const auto &f_drop = phv->get_field("psa_ingress_output_metadata.drop");
-    if (f_drop.get_int()) {
+    const auto &f_ig_cos = phv->get_field("psa_ingress_output_metadata.class_of_service");
+    const auto ig_cos = f_ig_cos.get_uint();
+
+    // ingress cloning - sent in their pre-ingress-thread state to either an egress
+    //                   port or a multicast group
```
antoninbas, line 427: I am not sure I understand this comment. What does "pre-ingress-thread state" mean?

targets/psa_switch/psa_switch.cpp
```diff
@@ -337,17 +420,57 @@ PsaSwitch::ingress_thread() {
...
+    if (drop) {
       BMLOG_DEBUG_PKT(*packet, "Dropping packet at the end of ingress");
       continue;
     }
 
     // resubmit - these packets get immediately resub'd to ingress, and skip
     //            deparsing, do not move below multicast or deparse
-    const auto &f_resubmit = phv->get_field("psa_ingress_output_metadata.resubmit");
-    if (f_resubmit.get_int()) {
+    unsigned int resubmit = phv->get_field("psa_ingress_output_metadata.resubmit").get_uint();
```
antoninbas, line 472: `auto resubmit = ...`

targets/psa_switch/psa_switch.cpp
```diff
@@ -363,49 +486,25 @@ PsaSwitch::ingress_thread() {
...
     auto &f_packet_path = phv->get_field("psa_egress_parser_input_metadata.packet_path");
-    const auto &f_ig_cos = phv->get_field("psa_ingress_output_metadata.class_of_service");
-    const auto ig_cos = f_ig_cos.get_uint();
-
-    // handling multicast
-    unsigned int mgid = 0u;
-    const auto &f_mgid = phv->get_field("psa_ingress_output_metadata.multicast_group");
-    mgid = f_mgid.get_uint();
 
+    unsigned int mgid = phv->get_field("psa_ingress_output_metadata.multicast_group").get_uint();
```
antoninbas, line 490: same as above: `auto mgid = ...` for style consistency

targets/psa_switch/psa_switch.cpp
```diff
@@ -337,17 +420,57 @@ PsaSwitch::ingress_thread() {
...
+    const auto &f_ig_cos = phv->get_field("psa_ingress_output_metadata.class_of_service");
+    const auto ig_cos = f_ig_cos.get_uint();
+
+    // ingress cloning - sent in their pre-ingress-thread state to either an egress
+    //                   port or a multicast group
+    //                 - dropped packets should still be cloned - do not move below drop
+    unsigned int clone = phv->get_field("psa_ingress_output_metadata.clone").get_uint();
+    if (clone) {
+      MirroringSessionConfig config;
+      unsigned int clone_session_id = phv->get_field("psa_ingress_output_metadata.clone_session_id").get_uint();
```
antoninbas, line 432: use `auto`

targets/psa_switch/psa_switch.cpp
```diff
@@ -337,17 +420,57 @@ PsaSwitch::ingress_thread() {
...
+      bool is_session_configured = mirroring_get_session(
+          static_cast<mirror_id_t>(clone_session_id), &config);
+      BMLOG_DEBUG_PKT(*packet, "Cloning packet at ingress to session id {}", clone_session_id);
```
antoninbas, line 435: is having this log message confusing in case the session is not configured? I know that I didn't do it for simple_switch, but maybe a log message to indicate that a session id which hasn't been configured has been requested would be a good idea. I wouldn't make it an error or even a warning, but it could be helpful when debugging. @jafingerhut what do you think?

antoninbas:
@jafingerhut this is the correct implementation in order to reuse the existing P4Runtime implementation. That implementation actually always creates one multicast group for each clone session: https://github.com/p4lang/PI/blob/771d5c46106f14f33777e70f39df2ae772cdfe17/proto/frontend/src/pre_clone_mgr.cpp#L107. IIRC this is done regardless of whether the clone session asks for a single replica or more, to facilitate future updates to the clone session configuration.

jafingerhut:
@antoninbas This PR has a function mirroring_session_add that creates a clone session and in one case assigns it the value of an mgid -- not a set of (egress_port, instance) pairs, but an mgid that can also be used for multicast replication in a P4 program by assigning that numeric value to the ostd.multicast_group metadata field in ingress.

It seems that either that needs to change before the P4Runtime API could possibly work correctly, or perhaps mirroring_session_add is only useful for implementing the Thrift API, and it would never be called from a P4Runtime API implementation?

targets/psa_switch/psa_switch.cpp
jafingerhut: I am certainly all for easier debugging of P4 code from the contents of the bmv2 log, so having different log messages for a configured vs. not-configured session could be nice.  Perhaps have the above log message inside the 'if (is_session_configured) {' 'then' branch below, and a slightly different one in an 'else' branch that said something like "Cloning packet at ingress to unconfigured session id {} causes no clone packets to be created" ?

antoninbas:
@jafingerhut On the contrary, this is specifically to implement the P4Runtime API. The PI implementation will create a multicast group, using a mgid value from a reserved number space. The multicast group will be configured with the correct set of (egress_port, instance), then bmv2 will be configured through `mirroring_session_add` to map the clone session id to this mgid. When the clone session id is used for a packet, this packet should be multicast. All the copies sent to egress will thus have the correct egress_port and instance value. Am I misunderstanding the question? This code was pretty much copied over from simple_switch as is, but I may have missed something when I was reviewing.

jafingerhut:
Ah, so the mgid values used for clone sessions come from a range completely outside of the [1, M] range that are used for multicast, and thus never conflict with them (for whatever maximum mgid value M that bmv2 supports)?

antoninbas:
This is the relevant code in the implementation:
```cpp
  // user-defined multicast group ids must be in the range
  // ]0,first_reserved_group[; ideally this should be configurable based on the
  // target.
  static constexpr GroupId first_reserved_group_id() { return 1 << 15; }
```
So the users can use up to 32K mgids. Everything else is reserved. The assumption here being that a target would typically support 64K mgids.

jafingerhut:
Thanks for the explanation.  That all makes sense to me now.  I wouldn't be surprised if some targets might not use an mgid value at all internally for clone sessions that multicast, but for targets that do, this looks like a good way to do it, while still preserving the contiguous range of mgid's usable for non-clone mgid values.

targets/psa_switch/psa_switch.cpp
peteli3: I accidentally started a review instead of commenting on this, then removed the review and thereby my response to this question.

I added in your suggested class_of_service handling to multicast() + a TBD comment.

jfingerh: Yep, I saw that, and looks good to me.

As I mentioned in another comment, it is important not only that with these changes, p4lang/behavioral-model automated tests on p4lang/behavioral-model repository pass, which is automatically initiated and recorded by Github when you commit here, but there is currently nothing automated to tell whether these changes will cause p4lang/p4c automated tests to pass or fail, until AFTER the p4lang/behavioral-model changes are merged in, and a few hours pass, and the next p4lang/p4c automated tests run for one of its PR commits.

If you know how to locally run the p4lang/p4c tests with these changes locally in your p4lang/behavioral-model repository, please let us know if those all pass (again, perhaps with ebpf and ubpf tests failing, but that is OK if you haven't done the necessary steps p4lang/p4c install steps for those to work, which I do not currently know what they are).

targets/psa_switch/pswitch_CLI.py
```diff
@@ -37,41 +50,69 @@ def __init__(self, pre_type, standard_client, mc_client, pswitch_client):
...
 
+    @handle_bad_input
     def do_set_queue_depth(self, line):
         "Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port>]"
         args = line.split()
-        depth = int(args[0])
+        depth = self.parse_int(args[0])
         if len(args) > 1:
-            port = int(args[1])
+            port = self.parse_int(args[1])
```
rst0git, line 59: `self.parse_int(args[1], "egress_port")`

targets/psa_switch/pswitch_CLI.py
```diff
@@ -37,41 +50,69 @@ def __init__(self, pre_type, standard_client, mc_client, pswitch_client):
                                         standard_client, mc_client)
         self.pswitch_client = pswitch_client
 
+    @handle_bad_input
     def do_set_queue_depth(self, line):
         "Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port>]"
         args = line.split()
-        depth = int(args[0])
+        depth = self.parse_int(args[0])
```
rst0git, line 57: `self.parse_int(args[0], "nb_pkts")`

targets/psa_switch/pswitch_CLI.py
```diff
@@ -37,41 +50,69 @@ def __init__(self, pre_type, standard_client, mc_client, pswitch_client):
...
             self.pswitch_client.set_egress_queue_depth(port, depth)
         else:
             self.pswitch_client.set_all_egress_queue_depths(depth)
 
+    @handle_bad_input
     def do_set_queue_rate(self, line):
         "Set rate of one / all egress queue(s): set_queue_rate <rate_pps> [<egress_port>]"
         args = line.split()
-        rate = int(args[0])
+        rate = self.parse_int(args[0])
```
rst0git, line 68: `self.parse_int(args[0], "rate_pps")`

targets/psa_switch/pswitch_CLI.py
```diff
@@ -37,41 +50,69 @@ def __init__(self, pre_type, standard_client, mc_client, pswitch_client):
...
         if len(args) > 1:
-            port = int(args[1])
+            port = self.parse_int(args[1])
```
rst0git, line 70: `self.parse_int(args[1], "egress_port")`

targets/psa_switch/pswitch_CLI.py
```diff
@@ -37,41 +50,69 @@ def __init__(self, pre_type, standard_client, mc_client, pswitch_client):
...
+        mgrp = self.parse_int(args[1], "mgrp")
+        config = MirroringSessionConfig(mgid=mgrp)
+        self.pswitch_client.mirroring_session_add(mirror_id, config)
+        print "Associating multicast group", mgrp, "to mirroring session", mirror_id
+
+    @handle_bad_input
     def do_mirroring_delete(self, line):
         "Delete mirroring mapping: mirroring_delete <mirror_id>"
-        mirror_id = int(line)
+        mirror_id = self.parse_int(line)
```
rst0git, line 98: `self.parse_int(line, "mirror_id")`

targets/psa_switch/pswitch_CLI.py
peteli3: thanks for catching - fixing this in next commit

targets/psa_switch/pswitch_CLI.py
peteli3: thanks for catching - fixing this in next commit

targets/psa_switch/pswitch_CLI.py
peteli3: thanks for catching - fixing this in next commit

targets/psa_switch/pswitch_CLI.py
peteli3: thanks for catching - fixing this in next commit

targets/psa_switch/pswitch_CLI.py
peteli3: thanks for catching - fixing this in next commit

targets/psa_switch/psa_switch.cpp
peteli3: Hmm I meant to say "packet as it was before it was picked up by the ingress thread" - but that seems incorrect now that I look at the PSA spec again. 

I changed this comment to say "each cloned packet is a copy of the packet as it entered the ingress parser" - what the PSA spec says.

targets/psa_switch/psa_switch.cpp
peteli3: making this in next commit

targets/psa_switch/psa_switch.cpp
peteli3: making this in next commit

targets/psa_switch/psa_switch.cpp
peteli3: making this change in next commit

targets/psa_switch/psa_switch.cpp
peteli3: making this change in next commit

peteli3:
> @peteli3 This is looking pretty good to me. Have you built the latest behavioral-model plus these changes on your machine, plus latest p4c plus your p4c PR's new test case, and tried running all p4c tests to see which pass and which fail?
> 
> It is OK if p4c tests with file names beginning with 'ebpf' or 'ubpf' fail (at least, I have yet to figure out how to build p4c so that they pass), but the rest should all pass before we make changes to behavioral-model repo.

@jafingerhut 
I ran all tests on clean builds of both p4c and behavioral-model:
```bash
# p4c built with these flags
$ cmake .. -DCMAKE_BUILD_TYPE=DEBUG -DENABLE_EBPF=false && make -j4

# output from p4c tests
100% tests passed, 0 tests failed out of 1574

Label Time Summary:
XFAIL        =  20.14 sec*proc (17 tests)
bmv2         = 1420.44 sec*proc (439 tests)
cpplint      =  17.57 sec*proc (1 test)
err          =  37.20 sec*proc (231 tests)
gtest        =  13.81 sec*proc (1 test)
p14_to_16    = 505.02 sec*proc (191 tests)
p4           = 685.80 sec*proc (696 tests)
ubpf         =   8.37 sec*proc (15 tests)

Total Test time (real) = 2691.05 sec
Built target check-all
Scanning dependencies of target check
Built target check
```

targets/psa_switch/psa_switch.cpp
peteli3: +1 for andy's suggestion. Adding in next commit.

peteli3:
@antoninbas please re-review when you get a chance, thanks!

PR #936, opened 2020-08-19
Fix to ingress_port assigned before ingress control is executed

jafingerhut:
The PSA spec says that ingress_port during ingress control should be
the same as it was sent into the ingress parser.  Without this fix,
ingress_port sent into the ingress control is always the original
input port on which the packet was received, which is incorrect for
recirculated packets.


jafingerhut:
@hanw @yunheL @peteli3 FYI, and if you are confident reviewing, please do so.  I have a slightly modified version of the PSA recirculate test in the p4lang/p4c repository that helped me catch this, which I will create a separate PR for in that repo.

PR #937, opened 2020-08-20
PSA small fixes for unicast and multicast forwarding

jafingerhut:
Make psa_switch was assign the required values to the 'instance' field
for multicast replicated packets.

Also assign the correct NORMAL_UNICAST or NORMAL_MULTICAST values to
packet_path for packets before they begin their egress processing.


jafingerhut:
@hanw @yunheL @peteli3 pinging you simply so you are aware of these changes.  It does not seem likely to me that it directly affects anything you are working on.

I have new and modified PSA test programs with STF packet tests for p4c that helped catch these issues, which I can merge into p4c after these changes, or similarly effective ones, are in psa_switch.

PR #945, opened 2020-09-09
[PSA] add support for extern psa-internetChecksum

yunheL:
Implemented `internetChecksum` following the suggested implemented given by PSA specification [[please see here](https://p4.org/p4-spec/docs/PSA-v1.1.0.html#appendix-internetchecksum-implementation)].

TODO:
1. Currently `internetChecksum.add()` can only take 1 field instead of a list of fields. For example:

```
InternetChecksum ck;
ck.add(fieldA); // this is supported
ck.add({fieldA, fieldB}) // this is not supported yet
```

I have not found an elegant way to support this. I am not entirely sure whether this should be handled on the p4c end or the bmv2 end. Advices and ideas are greatly appreciated!

### DIFF
--- a/targets/psa_switch/externs/psa_internetChecksum.cpp
+++ b/targets/psa_switch/externs/psa_internetChecksum.cpp
@@ -1,0 +1,119 @@
+/* Copyright 2020-present Cornell University
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Yunhe Liu (yunheliu@cs.cornell.edu)
+ *
+ */
+
+#include "psa_internetChecksum.h"
+#include <bm/bm_sim/logger.h>
+#include <execinfo.h>
+
+#define SUM_LENGTH 16
+
+namespace bm {
+
+namespace psa {
+
+uint16_t
+ones_complement_sum(uint16_t x, uint16_t y) {
+  uint32_t _x, _y, s;
+  uint16_t ret;
+  _x = x;
+  _y = y;
+  s = _x + _y;
+  if (s >> 16 == 1) {
+    s = s + 1;
+  }
+  ret = s;
+  return ret;
+}
+
+void
+PSA_InternetChecksum::init() {
+  sum = 0;
+}
+
+void
+PSA_InternetChecksum::clear() {
+  sum = 0;
+}
+
+void
+PSA_InternetChecksum::add(const Field &field) {
+  int len = field.get_nbits();
+  // TODO give a error if len is not a multiple of 16
+  Data field_val(field);
+
+  int i = 0;
+  while (i < len) {
+    Data tmp(field_val);
+    Data mask(0xffff);
+    tmp.bit_and(tmp, mask);
+    uint16_t d = tmp.get<uint16_t>();
+    sum = ones_complement_sum(sum, d);
+    field_val.bm::Data::shift_right(field_val, SUM_LENGTH);
+    i += SUM_LENGTH;
+  }
+}
+
+void
+PSA_InternetChecksum::subtract(const Field &field) {
+  int len = field.get_nbits();
+  // TODO give a error if len is not a multiple of 16
+  Data field_val(field);
+
+  int i = 0;
+  while (i < len) {
+    Data tmp(field_val);
+    Data mask(0xffff);
+    tmp.bit_and(tmp, mask);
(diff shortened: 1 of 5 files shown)


targets/psa_switch/externs/psa_internetChecksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+  uint16_t ret;
+  _x = x;
+  _y = y;
+  s = _x + _y;
+  if (s >> 16 == 1) {
+    s = s + 1;
+  }
+  ret = s;
+  return ret;
+}
```
antoninbas, line 42: put this in an anonymous namespace

targets/psa_switch/externs/psa_internetChecksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void
+PSA_InternetChecksum::clear() {
+  sum = 0;
+}
+
+void
+PSA_InternetChecksum::add(const Field &field) {
+  int len = field.get_nbits();
+  // TODO give a error if len is not a multiple of 16
```
antoninbas, line 57: I would recommend using `_BM_ASSERT` for that check

targets/psa_switch/externs/psa_internetChecksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+    sum = ones_complement_sum(sum, d);
+    field_val.bm::Data::shift_right(field_val, SUM_LENGTH);
+    i += SUM_LENGTH;
+  }
+}
+
+void
+PSA_InternetChecksum::subtract(const Field &field) {
+  int len = field.get_nbits();
+  // TODO give a error if len is not a multiple of 16
```
antoninbas, line 75: same as above

targets/psa_switch/externs/psa_internetChecksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  int i = 0;
+  while (i < len) {
+    Data tmp(field_val);
+    Data mask(0xffff);
+    tmp.bit_and(tmp, mask);
+    uint16_t d = tmp.get<uint16_t>();
+    sum = ones_complement_sum(sum, d);
+    field_val.bm::Data::shift_right(field_val, SUM_LENGTH);
+    i += SUM_LENGTH;
```
antoninbas, line 68: that looks like a very inefficient implementation, I don't understand why this has to be done bit by bit

targets/psa_switch/externs/psa_internetChecksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * limitations under the License.
+ */
+/*
+ * Yunhe Liu (yunheliu@cs.cornell.edu)
+ *
+ */
+
+#include "psa_internetChecksum.h"
+#include <bm/bm_sim/logger.h>
+#include <execinfo.h>
```
antoninbas, line 22: forgot to remove?

antoninbas (changes requested):
I think for an extern like this one, Gtest unit tests are needed to ensure the implementation is correct and doesn't break, and we shouldn't rely just on STF tests in the compiler, since the 2 are separate pieces of software and can be updated independently.

IIRC, we had a conversation on Slack on how the general case can be supported with some "simple" transformations in the p4c bmv2 backend. I think the key pieces of that discussion should be included in the PR description. Also, if the decision is to handle this in the compiler, the "TODO" is not really a TODO for bmv2, but for p4c.

targets/psa_switch/externs/psa_internetChecksum.cpp
jfingerh: When you say bit by bit, you mean that it is doing it in 16-bit word chunks?

I have no argument that there might be more efficient ways to implement it.  In practice I suspect this function will be called with lists of fields that are typically no more than 18 bytes (for IPv4 headers, minus checksum) or 18+2*16+2 = 42 bytes for TCP headers with IPv6 pseudo-headers, so having a correct implementation with perhaps a TODO for considering performance measurement and enhancement might be appropriate?

antoninbas: actually it's my bad, I read this too fast. I saw the while loop and didn't see that the increment was `i += SUM_LENGTH` at the end. Maybe we can replace it with `for (int i = 0; i < len; i+= SUM_LENGTH)`. It's less verbose and less confusing to me :)

on an unrelated node, maybe @yunheL can explain why the explicit reference to `bm::Data::shift_right` is needed in `field_val.bm::Data::shift_right(field_val, SUM_LENGTH);`? `field_val` is already a `Data` object, and even if it was a `Field`, I believe it sill would not be required.

antoninbas: However, regarding implementation performance, we should avoid creating all these temporary `Data` objects, which are kind of expensive. The internal state (`sum`) is already a uint16, so we can extract 16-bit chunks from the input `field` and do all the arithmetic on that. Looking at the current code, I only see 16-bit arithmetic, so there is no need for arbitrary-precision arithmetic / big nums.

That's one more reason for adding some C++ unit tests. We can quickly modify them to act as a benchmark.

antoninbas: And to clarify "we can extract 16-bit chunks from the input `field`", I think potentially the best is to get the binary string for the input `field` and read it 2 bytes at a time.

PR #947, opened 2020-09-13
add support for e2e cloning in psa_switch

peteli3:
Looking for feedback on this first attempt at implementing e2e cloning for psa_switch. I'll add a link to the associated STF test here as soon as I get a chance.

Thanks!


targets/psa_switch/psa_switch.h
```diff
@@ -212,7 +212,10 @@ class PsaSwitch : public Switch {
   //              by psa_switch and log an error msg / return error code
 
  private:
-  static constexpr size_t nb_egress_threads = 4u;
+  // temporarily setting this to 1 helps reduce nondeterminism fails in STF tests
+  // TODO figure out if we can make stf work around this
+  // TODO unrelated, make ingress threads configurable as well
+  static constexpr size_t nb_egress_threads = 1u;
```
jafingerhut, line 218: Out of curiosity, what kinds of nondeterministic behavior have you seen in STF tests?

I am wondering if there might be a relatively small and simple change to backends/bmv2/bmv2stf.py to implement a command called 'wait' that exists in several STF test files in the p4c/testdata/p4_14_samples directory.  I believe the basic idea of that command is to perform all command that appear before the first one, including any sending of packets from 'packet' commands, and waiting for and checking of output packets against 'expect' command contents, before proceeding past the 'wait' statement to look for more commands.

Right now I believe the 'wait' command is ignored and skipped, as are all unrecognized commands in STF files.  If the 'wait' command were correctly implemented, then most or all of the nondeterminism of packet processing order could be avoided by adding 'wait' commands between every 'packet' followed by 'expect' command.

antoninbas: also curious about this, since IIRC the order of `expect` commands in STF tests should not matter, and that's the only non-determinism I can think of that would be caused by the number of egress threads

if this is required for STF tests to work somehow (but once again, not sure why), you need to make `nb_egress_threads` a target-specific command-line option for psa_switch, and set it to 1 when creating the `simple_switch` process in the STF implementation for bmv2 in p4c.

targets/psa_switch/psa_switch.cpp
```diff
@@ -563,6 +563,39 @@ PsaSwitch::egress_thread(size_t worker_id) {
     Deparser *deparser = this->get_deparser("egress_deparser");
     deparser->deparse(packet.get());
 
+    // egress cloning - each cloned packet is a copy of the packet as output by the egress deparser
+    auto clone = phv->get_field("psa_egress_output_metadata.clone").get_uint();
+    if (clone) {
+      MirroringSessionConfig config;
+      auto clone_session_id = phv->get_field("psa_egress_output_metadata.clone_session_id").get_uint();
+      auto is_session_configured = mirroring_get_session(static_cast<mirror_id_t>(clone_session_id), &config);
```
antoninbas, line 571: should be:
```cpp
auto clone_session_id = phv->get_field("psa_egress_output_metadata.clone_session_id").get< mirror_id_t>();
auto is_session_configured = mirroring_get_session(clone_session_id, &config);
```

please "fix" it for ingress cloning as well

antoninbas (commented):
one minor comment, otherwise LGTM but the issue raised by @jafingerhut needs to be addressed before merging

targets/psa_switch/psa_switch.cpp
peteli3: done. adding in next commit

targets/psa_switch/psa_switch.h
peteli3: I'll create a PR with the stf test case I'm referring to here.

On 4 egress threads, each time I run the test, I get a different expectation match error. Here are a few of them:
```bash
RuntimeCmd:
*** Received packet  00000000000800000000BEEFFACE
*** Packet different at position 11 : expected 9 , received 8
*** Full expected packet is  00000000000900000000CAFEFACE
*** Full received packet is  00000000000800000000BEEFFACE
*** Packet 1 on port 7 differs
*** Test failed

RuntimeCmd:
*** Received packet  00000000000900000000CAFEFACE
*** Packet different at position 11 : expected 2 , received 9
*** Full expected packet is  00000000000200000000CAFEFACE
*** Full received packet is  00000000000900000000CAFEFACE
*** Packet 0 on port 7 differs
*** Test failed

RuntimeCmd:
*** Received packet  00000000000800000000BEEFFACE
*** Packet different at position 11 : expected 9 , received 8
*** Full expected packet is  00000000000900000000CAFEFACE
*** Full received packet is  00000000000800000000BEEFFACE
*** Packet 1 on port 7 differs
*** Test failed
```

Switching to 1 egress thread immediately fixes the issue, and tests have been passing for 10+ consecutive runs (still possible though, but I just haven't observed it yet)

peteli3: https://github.com/p4lang/p4c/pull/2549/files

peteli3: That said, it's also perfectly plausible that my test case is incorrect...

antoninbas: Looks like the STF framework wants packets to appear in the same order in the pcap output as the order of the `expect` statements in the .stf: https://github.com/p4lang/p4c/blob/557b77f8c41fc5ee1158710eda1073d84f5acf53/backends/bmv2/bmv2stf.py#L772-L776. I would say that this is not ideal and was not the intention for STF tests, but I don't know if @jafingerhut and @mbudiu-vmw share this opinion.

mbudiu-vmw: "S" is from "simple". So far all the tests that we wrote were deterministic. Perhaps some new syntax would be needed to describe non-deterministic behaviors in STF.

antoninbas: I am not sure the scope is that large here. I don't think we are trying to cover cases like ECMP where a packet may be sent to any port in a set. My understanding is that one of the initial design goals of STF was that `expect` statements could be fulfilled in any order, with the ability to define "barriers" using `wait` statements when desired.

mbudiu-vmw: It depends whether you are asking about ordering of packets between different ports and ordering of packets on the same port. Yes, initially there were no ordering requirements between ports; I don't remember whether we lost that at some point. If we don't care about supporting "unordered sets of packets within the same port" then indeed, it may be easier.

mbudiu-vmw: BTW: I think @ChrisDodd wrote the first implementation of STF; Barefoot may have a different one internally. There is no specification of STF.

antoninbas: In this case, the issue is indeed with packets on the same port. Looking at the STF code, the ordering does not matter across ports. If the intention was that ordering is important when looking at packets received on the same port, then I think we should go with @jafingerhut's idea and implement the `wait` command. In Peter's test, a `wait` statement will be needed before every `packet` statement except for the first one. It may not be sufficient for the last case actually ("RECIRCULATE PACKET CASE"), but maybe the test can be updated to avoid doing multicast on the recirculated packet.

mbudiu-vmw: Maybe it's not that complicated; we could add a command "EXPECT_SET" followed by a port and a list of packets (using some separator - comma?).

jafingerhut: I have not worked it out in detail, but I suspect that when there are multiple egress threads, the egress code for ports 1-8 (or some small range like that) are running in a different thread from egress code for the recirculate port.  Thus it is possible for recirculate port packets to "go ahead" or "go behind" of non-recirculate port packets, and when both the before-recirculate and after-recirculate egress code can create egress-to-egress cloned packets, the egress-to-egress cloned packets from the recirculated packet can then go out the ports either before or after the egress-to-egress cloned packets from the first pass packets.

Perhaps if you change the test program so that first pass packets and after-recirculated packets always clone to a different clone session id, with a different set of output ports, things will test more repeatably even with multiple egress threads?  In general, an STF test where at most one packet comes out of each output port seems like it should be immune from such non-deterministic behavior.

jafingerhut: Sorry, I hadn't read the latest comments before writing my previous one.

Yes, 'wait' would help.  'expect_set' could also be made to work.

WIth the test program linked, I suspect that 'wait' by itself would not be enough, because for that test program, it looks like even a single input packet could lead to non-deterministic output packet order when there are multiple threads in bmv2.

antoninbas: My favorite option would be to implement `wait` as it's already an STF command theoretically, and avoid introducing a new one. This will work for most cases and in the case of Peter's test, a small adjustment to the test will make it work. But it may not be straightforward to implement `wait` correctly for the bmv2 implementation of STF, which currently relies on pcap files for packet IO.

jafingerhut: I've thought briefly about the current bmv2stf.py code for 'executing' STF tests, and it doesn't look like a lot of work to implement 'wait', if once a pcap file is opened, one can read all packets written to it so far, but then leave the file open and read more packets from it later when the next 'wait' is reached (or the end of the STF test).

For this particular STF test, I spoke with Peter and it should be straightforward to make the test cover the functionality we want, but send no more than one packet to each output port.  That should avoid all packet reordering issues.

jafingerhut (approved):
LGTM

jafingerhut (changes requested):
Actually, I take it back.  I think the number of egress threads should be left unchanged, and then I think these changes look good to me.

targets/psa_switch/psa_switch.h
antoninbas: To correctly honor the `wait`, one would need to stop sending packets when reaching a `wait`. So the input pcap is dynamic as well. I don't know how well that will work. I believe bmv2 may support a single input pcap file per interface and will stop when EOF is reached, even though STF may not be done writing to the file.

jafingerhut: Got it.  I had not considered the possibility that bmv2 might forever stop reading from a pcap input file when EOF is reached.  That would make the implementation of 'wait' more difficult, e.g. requiring a change in bmv2, or using virtual Ethernet interfaces instead, which would also then require root privileges for running p4c tests.

antoninbas: Got an idea: we can switch to using a named pipe in bmv2stf.py. This way we can just keep the pipe open and write pcap data as needed, with no change needed in bmv2 and no potential for a race condition. I'll take a stab at it.

antoninbas: Actually bmv2stf.py already uses named pipes for inputs, but we may want to use them for outputs as well

peteli3: Will follow up on this bmv2stf stuff separately. I've added links to Antonin's comments in the P4 arch wg meeting notes, so we don't lose track of it.

For now, I took Andy's workaround to make the stf test send no more than one packet to each output port.

jafingerhut (approved):
Looks good to me.  Thanks for removing the number of egress threads changes, and adding the egress drop handling.

antoninbas (approved):
LGTM

PR #951, opened 2020-09-22
Add support for systems that have Python3 only

### LINKED ISSUE #709: Thoughts on transitioning to Python 3?
I get this ominous-looking message when running `pip install ...` commands based on Python 2.7 on my Ubuntu Linux 16.04 machine now:

```
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020.
Please upgrade your Python as Python 2.7 won't be maintained after that date.
A future version of pip will drop support for Python 2.7.
```

I am sure this issue is shared by many other projects, and I suspect a lot of them will simply continue using Python 2.7 after it is no longer maintained by whoever is doing it now (and perhaps maintenance might be taken up by some other organization), but the writing seems to be on the wall.

fruffy:
Let's see what happens... 

Fixes #709 if everything works out.

### DIFF
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,2 +1,6 @@
 set noparent
 filter=-build/include_dirname,-build/c++11
+filter=-build/include_subdir
+filter=-build/include_order
+filter=-build/include_what_you_use
+filter=-runtime/string
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,6 +1,2 @@
 set noparent
-filter=-build/c++11
-filter=-build/include_subdir
-filter=-build/include_order
-filter=-build/include_what_you_use
-filter=-runtime/string
+filter=-build/include_dirname,-build/c++11
--- a/CPPLINT.cfg
+++ b/CPPLINT.cfg
@@ -1,5 +1,5 @@
 set noparent
-filter=-build/include_dirname,-build/c++11
+filter=-build/c++11
 filter=-build/include_subdir
 filter=-build/include_order
 filter=-build/include_what_you_use
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,6 @@ RUN apt-get update && \
     apt-get update && \
     apt-get install -y --no-install-recommends $BM_DEPS $BM_RUNTIME_DEPS && \
     ln -sf /usr/bin/python3 /usr/bin/python && \
-    ln -sf /usr/bin/pip3 /usr/bin/pip && \
     ./autogen.sh && \
     if [ "$GCOV" != "" ]; then ./configure --with-pdfixed --with-pi --with-stress-tests --enable-debugger --enable-coverage --enable-Werror; fi && \
     if [ "$GCOV" = "" ]; then ./configure --with-pdfixed --with-pi --with-stress-tests --enable-debugger --enable-Werror; fi && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,8 @@ RUN apt-get update && \
     add-apt-repository -y ppa:ubuntu-toolchain-r/test && \
     apt-get update && \
     apt-get install -y --no-install-recommends $BM_DEPS $BM_RUNTIME_DEPS && \
+    ln -sf /usr/bin/python3 /usr/bin/python && \
+    ln -sf /usr/bin/pip3 /usr/bin/pip && \
     ./autogen.sh && \
     if [ "$GCOV" != "" ]; then ./configure --with-pdfixed --with-pi --with-stress-tests --enable-debugger --enable-coverage --enable-Werror; fi && \
     if [ "$GCOV" = "" ]; then ./configure --with-pdfixed --with-pi --with-stress-tests --enable-debugger --enable-Werror; fi && \
--- a/targets/simple_switch/sswitch_CLI.py
+++ b/targets/simple_switch/sswitch_CLI.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # Copyright 2013-present Barefoot Networks, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/targets/simple_switch/sswitch_CLI.py
+++ b/targets/simple_switch/sswitch_CLI.py
@@ -1,5 +1,4 @@
-#!/usr/bin/env python2
-
+#!/usr/bin/env python3
 # Copyright 2013-present Barefoot Networks, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -38,7 +37,7 @@ def handle_bad_input(f):
             return f(*args, **kwargs)
         except InvalidMirroringOperation as e:
             error = MirroringOperationErrorCode._VALUES_TO_NAMES[e.code]
-            print "Invalid mirroring operation (%s)" % error
+            print("Invalid mirroring operation (%s)" % error)
     return handle
 
 class SimpleSwitchAPI(runtime_CLI.RuntimeAPI):
@@ -110,17 +109,17 @@ class SimpleSwitchAPI(runtime_CLI.RuntimeAPI):
         self.exactly_n_args(args, 1)
         mirror_id = self.parse_int(args[0], "mirror_id")
         config = self.sswitch_client.mirroring_session_get(mirror_id)
-        print config
+        print(config)
 
     @handle_bad_input
     def do_get_time_elapsed(self, line):
         "Get time elapsed (in microseconds) since the switch started: get_time_elapsed"
-        print self.sswitch_client.get_time_elapsed_us()
+        print(self.sswitch_client.get_time_elapsed_us())
 
     @handle_bad_input
     def do_get_time_since_epoch(self, line):
         "Get time elapsed (in microseconds) since the switch clock's epoch: get_time_since_epoch"
-        print self.sswitch_client.get_time_since_epoch_us()
+        print(self.sswitch_client.get_time_since_epoch_us())
 
 def main():
     args = runtime_CLI.get_parser().parse_args()
--- a/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
+++ b/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
@@ -1,5 +1,5 @@
 ????
-my_indirect_counter[0]=  BmCounterValue(packets=0, bytes=0)
+my_indirect_counter[0]= 0 bytes 0 packets
 ????
 Invalid counter operation (INVALID_INDEX)
 ????
--- a/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
+++ b/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
@@ -1,5 +1,5 @@
 ????
-my_indirect_counter[0]=  BmCounterValue(bytes=0, packets=0)
+my_indirect_counter[0]=  BmCounterValue(packets=0, bytes=0)
 ????
 Invalid counter operation (INVALID_INDEX)
 ????
--- a/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
+++ b/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
@@ -1,5 +1,5 @@
 ????
-my_indirect_counter[0]= 0 bytes 0 packets
+my_indirect_counter[0]= (0 bytes, 0 packets)
 ????
 Invalid counter operation (INVALID_INDEX)
 ????
--- a/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
+++ b/targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
@@ -1,5 +1,5 @@
 ????
-my_indirect_counter[0]=  BmCounterValue(packets=0, bytes=0)
+my_indirect_counter[0]=  BmCounterValue(bytes=0, packets=0)
 ????
 Invalid counter operation (INVALID_INDEX)
 ????
--- a/tools/nanomsg_client.py
+++ b/tools/nanomsg_client.py
@@ -1,5 +1,4 @@
-#!/usr/bin/env python2
-
+#!/usr/bin/env python3
 # Copyright 2013-present Barefoot Networks, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -46,7 +45,7 @@ class NameMap:
         self.names = {}
         json_ = json.loads(json_cfg)
         # special case where the switch was started with an empty config
-        if len(json_.keys()) == 0:
+        if len(list(json_.keys())) == 0:
             return
 
         for type_ in {"header_type", "header", "parser",
@@ -79,7 +78,7 @@ class MSG_TYPES:
      CHECKSUM_UPDATE,
      PIPELINE_START, PIPELINE_DONE,
      CONDITION_EVAL, TABLE_HIT, TABLE_MISS,
-     ACTION_EXECUTE) = range(15)
+     ACTION_EXECUTE) = list(range(15))
     CONFIG_CHANGE = 999
 
     @staticmethod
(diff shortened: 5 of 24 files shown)


antoninbas:
The very first step for this should be to update the base Docker image for p4lang to support both Python2 and Python3: https://github.com/p4lang/third-party. That's a pretty tedious process because the testing has to be done manually first: by building all the images locally (third-party -> pi -> bmv2 -> p4c), making sure the build succeeds and that tests pass when using the new image.

fruffy:
Uff that is quite the handful. I will see if I can get the third party image to work. I guess PI also has to be converted to support both Python 2 and 3?

jafingerhut:
@fruffy Regarding the BmCounterValue thing, if I do an interactive Python2 session on a machine where I've installed bmv2 the Python2 packages it installs, I can see what I've copied below.  I guess I could try out your changes locally on my machine, but if you could try an interactive Python3 session on a machine where you've built bmv2 with these changes, what output do you see?
```
$ python
Python 2.7.17 (default, Jul 20 2020, 15:37:01) 
[GCC 7.5.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import bm_runtime.standard.ttypes
>>> c1=bm_runtime.standard.ttypes.BmCounterValue()
>>> c2=bm_runtime.standard.ttypes.BmCounterValue(bytes=1,packets=2)
>>> c3=bm_runtime.standard.ttypes.BmCounterValue(packets=2,bytes=1)
>>> c1.__dict__
{'packets': None, 'bytes': None}
>>> c2.__dict__
{'packets': 2, 'bytes': 1}
>>> c3.__dict__
{'packets': 2, 'bytes': 1}
>>> c2 == c3
True
>>> c3 == c2
True
```
As far as I can tell, BmCounterValue(bytes=1,packets=2) and BmCounterValue(packets=2,bytes=1) should implement `==` comparison by comparing their `.__dict` values to each other, which you can see in Python2 are the same dictionary, and they are not ordered dicts so I think it should be returning true if they have the same keys (regardless of insertion order), and the same value corresponding with each key.

Maybe the Python3 version has other keys inside of the `.__dict` dictionaries that are preventing them from comparing equal?

fruffy:
Thanks @jafingerhut this put me on the right track. Essentially, the variables in this struct here 
https://github.com/p4lang/behavioral-model/blob/b2b86662060f6c843a01cd2996822e4280528fd7/thrift_src/standard.thrift#L82-L85
need to be swapped 

Python 2 does not respect the order of declaration in the init function, whereas Python 3 does. 
```python3
class BmCounterValue(object):
    """
    Attributes:
     - bytes
     - packets

    """


    def __init__(self, packets=None, bytes=None,):
        self.bytes = bytes
        self.packets = packets

```
However, I do not know what the fallout of this change might be.

jafingerhut:
@fruffy Would you mind trying out the interactive Python3 session with the commands I mentioned above, and paste what output you get from trying those things?  It seems to me that rather than tweaking the Thrift API definitions, it would be good to figure out why the two BmCounterValue objects are not `==` to each other when the constructor method is called with parameters in different order.  That seems like it should not happen.

fruffy:
Sure this was the output

```python3
Python 3.8.2 (default, Jul 16 2020, 14:00:26) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bm_runtime.standard.ttypes
>>> c1=bm_runtime.standard.ttypes.BmCounterValue()
>>> c2=bm_runtime.standard.ttypes.BmCounterValue(bytes=1,packets=2)
>>> c3=bm_runtime.standard.ttypes.BmCounterValue(packets=2,bytes=1)
>>> c1.__dict__
{'bytes': None, 'packets': None}
>>> c2.__dict__
{'bytes': 1, 'packets': 2}
>>> c3.__dict__
{'bytes': 1, 'packets': 2}
>>> c2 == c3
True
>>> c3 == c2
True
>>> 
```

jafingerhut:
@fruffy - Oh, I think I see what is going on -- that expected output file is text, and it is expected to be character-for-character identical to the actual output, whereas between Python2 and Python3 the printing order of keys for that dict might be semi-arbitrary?  And thus even though `==` returns true, the test still fails.

Antonin probably has better suggestions, but another option might be to sort the keys in a deterministic order in the print representation method of class BmCounterValue.  Although if that BmCounterValue definition is auto-generated Python code from Thrift, that might not be possible to do and consistently get into the auto-generated code, I guess.

Anyway, thanks for indulging my curiosity.

antoninbas:
I don't know if there is a requirement to work for both Python2 and Python3. Really, if we add support for Python3 and make it the default, then the test only needs to work for Python3.

That being said, I think this function can also be updated to print a different output, that does not depend on which version of Python is used: https://github.com/p4lang/behavioral-model/blob/e121539d558569224f48818e76cac6af6d3f968c/tools/runtime_CLI.py#L1962

fruffy:
Travis is being weird, I cannot reproduce these issues locally. But the Python 3 tests should work now. I will clean this up tomorrow.

CPPLINT.cfg
```diff
@@ -1,2 +1,6 @@
 set noparent
 filter=-build/include_dirname,-build/c++11
+filter=-build/include_subdir
+filter=-build/include_order
+filter=-build/include_what_you_use
+filter=-runtime/string
```
fruffy, line 6: I had to add these filters after updating cpplint.py to a newer version.

antoninbas: I think it would have been nice to update cpplint.py in a separate PR to keep the diff small and targeted
That being said, if you do it as part of this PR, I think we should try to address these warnings as much as possible instead of silencing them (and since this requires changing C++ code potentially, I think that's another argument in favor of doing it in a separate PR).

fruffy: Unfortunately, `2to3` does not really work well on `cpplint.py`. The changes necessary are a lot more involved. So I thought it would be best to just use a recent version.

Maybe the C++ changes can be addressed in a separate PR? At least `runtime/string`, `build/include_what_you_use`, and `build/include_what_you_use` can be done quickly. `build/include_subdir` corresponds to the old `build/include_dirname` I believe.

antoninbas: we could have left it for a future PR actually. There was no need to update cpplint.py to work with Python3 at the same time as the rest of the Python code.

fruffy: Reverted the changes. cpplint.py is now run with Python 2 again.

targets/simple_switch/sswitch_CLI.py
```diff
@@ -38,7 +37,7 @@ def handle(*args, **kwargs):
             return f(*args, **kwargs)
         except InvalidMirroringOperation as e:
             error = MirroringOperationErrorCode._VALUES_TO_NAMES[e.code]
-            print "Invalid mirroring operation (%s)" % error
+            print("Invalid mirroring operation (%s)" % error)
```
antoninbas, line 40: I believe `%` is deprecated now and will be removed in a future version. We can replace it with `format` in a future PR.

tools/runtime_CLI.py
```diff
@@ -94,11 +102,13 @@ def __call__(self, parser, namespace, values, option_string=None):
                         type=str, action="store", required=False)
 
     parser.add_argument('--pre', help='Packet Replication Engine used by target',
-                        type=str, choices=['None', 'SimplePre', 'SimplePreLAG'],
+                        type=str, choices=[
+                            'None', 'SimplePre', 'SimplePreLAG'],
```
antoninbas, line 106: what led to this formatting change?

tools/nanomsg_client.py
```diff
@@ -46,7 +49,7 @@ def load_names(self, json_cfg):
         self.names = {}
         json_ = json.loads(json_cfg)
         # special case where the switch was started with an empty config
-        if len(json_.keys()) == 0:
+        if len(list(json_.keys())) == 0:
```
antoninbas, line 52: I think `len(json_)` works for both Python2 & Python3?

tools/p4dbg.py
```diff
@@ -62,7 +72,7 @@ def load_names(self, json_cfg):
         json_ = json.loads(json_cfg)
 
         # special case where the switch was started with an empty config
-        if len(json_.keys()) == 0:
+        if len(list(json_.keys())) == 0:
```
antoninbas, line 75: same as above

tools/p4dbg.py
```diff
@@ -123,8 +133,8 @@ def get_all_fields():
     return field_map.get_all_fields()
 
 def enum(type_name, *sequential, **named):
```
antoninbas, line 135: we should consider dropping support for Python2 altogether in a future PR, then we can use `enum.Enum`

tools/p4dbg.py
```diff
@@ -167,14 +177,14 @@ def reset(self):
             7: "deparser",
         }
         self.ids_rev = {}
-        for k, v in self.ids.items():
+        for k, v in list(self.ids.items()):
```
antoninbas, line 180: `list` not needed

tools/p4dbg.py
```diff
@@ -167,14 +177,14 @@ def reset(self):
...
             self.ids_rev[v] = k
 
     def load_names(self, json_cfg):
         json_ = json.loads(json_cfg)
 
         # special case where the switch was started with an empty config
-        if len(json_.keys()) == 0:
+        if len(list(json_.keys())) == 0:
```
antoninbas, line 187: `len(json_)`

Dockerfile
```diff
@@ -50,6 +50,8 @@ RUN apt-get update && \
     add-apt-repository -y ppa:ubuntu-toolchain-r/test && \
     apt-get update && \
     apt-get install -y --no-install-recommends $BM_DEPS $BM_RUNTIME_DEPS && \
+    ln -sf /usr/bin/python3 /usr/bin/python && \
+    ln -sf /usr/bin/pip3 /usr/bin/pip && \
```
antoninbas, line 54: does anything use `pip` in the Docker image?

antoninbas (commented):
Thanks for taking care of this and sorry for the late review

`list()` is not needed when iterating over dictionary items (`items()` works for both Python2 and Python3). I flagged some of the occurrences but not all of them.

I think we should consider dropping support for Python2 altogether in a future PR.

targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
```diff
@@ -1,5 +1,5 @@
 ????
-my_indirect_counter[0]=  BmCounterValue(packets=0, bytes=0)
+my_indirect_counter[0]= 0 bytes 0 packets
```
antoninbas, line 2: can we use a more "structured" output, e.g. `(bytes=0, packets=0)`

antoninbas (commented):
one more nit, otherwise LGTM

targets/simple_switch/tests/CLI_tests/testdata/indirect_res.out
fruffy: I guess once the transition is complete we can also switch back to the normal printing style?

antoninbas: Not necessarily, I don't see a reason to change it again. There was no reason to have the Thrift type name there in the first place

antoninbas (approved):

jafingerhut:
I just wanted to say a big THANK YOU to you two for working through these changes.  I know such things can end up being 3 to 4 times more work than you guessed when you started on them.  It will definitely be nice to no longer see warnings from pip2 that support ended 2020-Jan-01, when you have this all done.  Let me know your favorite beverage and I'll see whether it is legal to ship it to you :-)

fruffy:
I am still in Germany so this might be a little too expensive ;-). But I appreciate the offer ha.

edgar-costa:
I also want to thank you @fruffy its been awhile I wanted to do that myself or with some student here at the university, and you will save us a good amount of time!!

I tried to build the `bmv2` in an ubuntu 18.04, and I get this error during the `make install`.

``` /bin/mkdir -p '/usr/local/bin'
  /bin/bash ../../libtool   --mode=install /usr/bin/install -c simple_switch '/usr/local/bin'
libtool: install: /usr/bin/install -c .libs/simple_switch /usr/local/bin/simple_switch
 /bin/mkdir -p '/usr/local/bin'
 /usr/bin/install -c simple_switch_CLI '/usr/local/bin'
 /bin/mkdir -p '/usr/local/include'
 /bin/mkdir -p '/usr/local/include/bm/simple_switch'
 /usr/bin/install -c -m 644  bm/simple_switch/runner.h '/usr/local/include/bm/simple_switch'
 /bin/mkdir -p '/usr/local/include/bm'
 /usr/bin/install -c -m 644 gen-cpp/bm/simple_switch_constants.h gen-cpp/bm/SimpleSwitch.h gen-cpp/bm/simple_switch_types.h '/usr/local/include/bm'
 /bin/mkdir -p '/usr/local/lib/python3.6/site-packages/sswitch_runtime'
 /usr/bin/install -c -m 644 gen-py/sswitch_runtime/constants.py gen-py/sswitch_runtime/__init__.py gen-py/sswitch_runtime/SimpleSwitch.py gen-py/sswitch_runtime/ttypes.py '/usr/local/lib/python3.6/site-packages/sswitch_runtime'
Byte-compiling python modules...
constants.py__init__.pySimpleSwitch.py  File "/usr/local/lib/python3.6/site-packages/sswitch_runtime/SimpleSwitch.py", line 554
    except InvalidMirroringOperation, ouch:
                                    ^
SyntaxError: invalid syntax

ttypes.py
```

Same for PSA.

```
/bin/mkdir -p '/usr/local/bin'
  /bin/bash ../../libtool   --mode=install /usr/bin/install -c psa_switch '/usr/local/bin'
libtool: install: /usr/bin/install -c .libs/psa_switch /usr/local/bin/psa_switch
 /bin/mkdir -p '/usr/local/bin'
 /usr/bin/install -c psa_switch_CLI '/usr/local/bin'
 /bin/mkdir -p '/usr/local/include/bm'
 /usr/bin/install -c -m 644 gen-cpp/bm/psa_switch_constants.h gen-cpp/bm/PsaSwitch.h gen-cpp/bm/psa_switch_types.h '/usr/local/include/bm'
 /bin/mkdir -p '/usr/local/lib/python3.6/site-packages/pswitch_runtime'
 /usr/bin/install -c -m 644 gen-py/pswitch_runtime/constants.py gen-py/pswitch_runtime/__init__.py gen-py/pswitch_runtime/PsaSwitch.py gen-py/pswitch_runtime/ttypes.py '/usr/local/lib/python3.6/site-packages/pswitch_runtime'
Byte-compiling python modules...
constants.py__init__.pyPsaSwitch.py  File "/usr/local/lib/python3.6/site-packages/pswitch_runtime/PsaSwitch.py", line 401
    except InvalidMirroringOperation, ouch:
                                    ^
SyntaxError: invalid syntax
```

fruffy:
Hmm looks like Python 3 is supported starting with Thrift 0.10.0:

https://issues.apache.org/jira/browse/THRIFT-1857

This repo uses Thrift 0.11.0. Maye the minimum thrift version needs to be upgraded?

edgar-costa:
Another thing, aren't all the `utils` in the https://github.com/p4lang/tutorials using python2? Also in `p4app`?

jafingerhut:
Short response: I believe updating p4lang/behavioral-model travis/install-thrift.sh to install version 0.11.0 instead of the current 0.9.2 is a safe change to make, and seems like now is as good a time as any to make that change.

Gory details:
@fruffy When you say "this repo uses Thrift 0.11.0", do you mean p4lang/behavioral-model uses it?   Because this script https://github.com/p4lang/behavioral-model/blob/master/travis/install-thrift.sh still installs 0.9.2.

I have changed that script in my local builds to use 0.11.0 for months now, and although I doubt I have tested every code path that might be Thrift-related, table entry adds/deletes/etc. seem to work fine.  Definitely all p4c STF tests do, and they use the Thrift API today for adding table entries.

I think @antoninbas might have made updates to Thrift version to 0.11.0 in some p4lang public repo somewhere, but do not recall which ones right now.

antoninbas:
The base Docker image used for CI includes Thrift 0.11.0. But yes it seems that it would be a good idea to update https://github.com/p4lang/behavioral-model/blob/master/travis/install-thrift.sh and the information in the README.

jafingerhut:
I will create a PR very soon for Thrift 0.11.0 update for review.

fruffy:
@edgar-costa I do not know about these repos but considering Mininet already supports Python 3 the migration should be straightforward. I personally just used `2to3 -W` on the root folder and then made custom edits.

PR #954, opened 2020-10-09
Update build script and README to Thrift version 0.11.0

jafingerhut:
This version has been officially supported for a while now, and will
likely help in moving from Python2 to Python3.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -6,6 +6,8 @@ then
     # This older package libssl1.0-dev enables compiling Thrift 0.9.2
     # on Ubuntu 18.04.  Package libssl-dev exists, but Thrift 0.9.2
     # fails to compile when it is installed.
+    # TBD whether using this package makes a difference for Ubuntu
+    # 18.04 and Thrift 0.11.0.
     LIBSSL_DEV="libssl1.0-dev"
 else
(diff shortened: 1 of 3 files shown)


install_deps.sh
```diff
@@ -6,6 +6,8 @@ then
     # This older package libssl1.0-dev enables compiling Thrift 0.9.2
     # on Ubuntu 18.04.  Package libssl-dev exists, but Thrift 0.9.2
     # fails to compile when it is installed.
+    # TBD whether using this package makes a difference for Ubuntu
+    # 18.04 and Thrift 0.11.0.
```
jafingerhut, line 10: Let me know if anyone considers it important to determine whether we can use same libssl-dev package across Ubuntu 16.04 and 18.04 here.  The script as written works, but might also work with a common package for all supported Ubuntu versions.

antoninbas (approved):
LGTM

install_deps.sh
antoninbas: as long as it works :)

adhicode:
Is Thrift 0.13.0 supported? When i built BMv2 with thrift-0.13.0, i could run successfully "make check". Pls specify if we have any impact. Thanks

PR #963, opened 2020-10-28
Update cpplint to latest version

fruffy:
This pull request updates `cpplint.py` to the latest version from [here](https://github.com/cpplint/cpplint/blob/develop/cpplint.py).
There were a bunch of changes necessary to make the test pass again but most of them were cosmetic such as reordering of header includes. 

In a couple instances I had to change `const std::string` to `const char,` for example [here](https://github.com/p4lang/behavioral-model/pull/963/files#diff-0a3836fa670f1665489e832b6d05bcda4dd9f6e0ce907adc4373b5d38b3ef552R160).

This is hopefully the last change needed for Python 3 support.


src/bm_sim/tables.cpp
```diff
@@ -14,16 +14,17 @@
  */
 
 /*
- * Antonin Bas (antonin@barefootnetworks.com)
+ *  Antonin Bas (antonin@barefootnetworks.com)
```
antoninbas, line 17: why the extra space just in this file?

src/bm_sim/tables.cpp
```diff
@@ -14,16 +14,17 @@
...
  */
-
 #include <bm/bm_sim/tables.h>
+
 #include <bm/bm_sim/debugger.h>
 #include <bm/bm_sim/logger.h>
 
 #include <string>
 
+
```
antoninbas, line 27: why the extra empty line just for that specific `namespace` declaration?

src/bm_sim/tables.cpp
```diff
@@ -14,16 +14,17 @@
  */
 
 /*
- * Antonin Bas (antonin@barefootnetworks.com)
+ *  Antonin Bas (antonin@barefootnetworks.com)
  *
  */
-
```
antoninbas, line 20: why?

src/bm_sim/tables.cpp
```diff
@@ -14,16 +14,17 @@
...
 #include <bm/bm_sim/tables.h>
+
 #include <bm/bm_sim/debugger.h>
 #include <bm/bm_sim/logger.h>
```
antoninbas, line 23: I would understand a re-ordering to be in lexicographic order but I don't understand the extra empty line?

src/bm_sim/tables.cpp
```diff
@@ -35,7 +36,7 @@ MatchActionTable::MatchActionTable(
 const ControlFlowNode *
 MatchActionTable::operator()(Packet *pkt) const {
   // TODO(antonin)
-  // this is temporary while we experiment with the debugger
+  //  this is temporary while we experiment with the debugger
```
antoninbas, line 39: this also doesn't look tight

maybe the script wants: `// TODO(antonin): this is temporary while we experiment with the debugger`

antoninbas (commented):
Thanks for taking care of this. However, a few things don't look right (see comments for tables.cpp) and deserve an explanation. Similar things may be happening in other files.

src/bm_sim/tables.cpp
fruffy: `tables.cpp:36:  Too many spaces before TODO  [whitespace/todo] [2]`
This is the particular complaint it has. Will put it all in one line.

fruffy:
I updated `tables.cpp` and rechecked the other files.

src/bm_sim/options_parse.cpp
```diff
@@ -18,22 +18,22 @@
  *
  */
 
-#include <bm/config.h>
 #include <bm/bm_sim/options_parse.h>
+
```
antoninbas, line 22: I think this is another instance where includes should be re-ordered instead of adding an empty line?

src/bm_sim/switch.cpp
```diff
@@ -18,25 +18,25 @@
...
 
 #include <cassert>
 #include <fstream>
-#include <string>
-#include <vector>
 #include <iostream>
 #include <streambuf>
+#include <string>
+#include <vector>
+#include <boost/filesystem.hpp>
```
antoninbas, line 39: add empty line before this include statement to match other files

targets/simple_switch_grpc/switch_sysrepo.cpp
```diff
@@ -33,6 +31,9 @@
 #include <string>
 #include <unordered_map>
 
+#include <boost/optional.hpp>
+
+
```
antoninbas, line 36: remove extra empty line

antoninbas (commented):
couple more nits, otherwise LGTM

antoninbas (approved):
LGTM

PR #989, opened 2021-03-24
Debian/Ubuntu packaging

dreibh:
Debian/Ubuntu packaging:
- Added Debian/Ubuntu packaging files (debian/ directory)
- Added scripts to create custom Debian/Ubuntu packages, by using pbuilder or Ubuntu Launchpad PPA upload (details in README.Debian).
- Added Linux Software Map (LSM) entry file.


dreibh:
Note: The build fixes from https://github.com/p4lang/behavioral-model/pull/988 are necessary, in order to have all required files for a build in a "make dist" tarball!

dreibh:
Built packages for Ubuntu Focal, Groovy and Hirsuite can be found in my Launchpad PPA: https://launchpad.net/~dreibh/+archive/ubuntu/ppa/+packages?field.name_filter=bm

I filed a Debian RFS for the packaging as well: https://mentors.debian.net/package/bm/

build-deb
```diff
@@ -0,0 +1,123 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Contact: dreibh@iem.uni-due.de
+
+# Bash options:
+set -e
```
jafingerhut, line 22: I do not know if there is an overall policy here for this repository, but most p4.org code in the p4lang organization is released under the Apache 2.0 license.  I doubt having a single build file like this under GPL would necessarily cause an issue, but wanted to bring it to the attention of @antoninbas in case there is a policy.

jafingerhut: And same comment for other files proposed to be added in this PR with GPL license.

debian/control
```diff
@@ -0,0 +1,132 @@
...
+               python3-thrift,
+               thrift-compiler
+Standards-Version: 4.5.1.0
+Rules-Requires-Root: no
+Homepage: https://github.com/simula/behavioral-model
+Vcs-Git: https://github.com/simula/behavioral-model
+Vcs-Browser: https://github.com/simula/behavioral-model/tree/p4-packaging
+
+Package: p4lang-bmv2-bin
+Architecture: amd64 arm64
```
jafingerhut, line 41: Has this been tested to work on arm64?

debian/control
```diff
@@ -0,0 +1,132 @@
...
+         ${python3:Depends},
+         ${shlibs:Depends}
+Recommends: hipercontracer,
+            netperfmeter,
+            rsplib-tools,
+            subnetcalc,
+            tcpdump,
+            tsctp,
+            tshark
+Description: p4lang behavioral-model (binaries)
```
jafingerhut, line 54: I believe this list of packages is only "recommended", not "required", but hipercontracer for example does not even exist in Ubuntu 18.04 package system.  Some of the others I've never heard of before either.  Would it be better to recommend some of them in a README or example of other interesting tools to help debug things, rather than auto-recommended when installing this package?

debian/python3-bmv2.install
```diff
@@ -0,0 +1,54 @@
...
+usr/lib/python*/site-packages/pswitch_runtime/__init__.py
+usr/lib/python*/site-packages/pswitch_runtime/constants.py
+usr/lib/python*/site-packages/pswitch_runtime/ttypes.py
+usr/lib/python*/site-packages/runtime_CLI.py
+usr/lib/python*/site-packages/sswitch_CLI.py
+usr/lib/python*/site-packages/sswitch_runtime/
+usr/lib/python*/site-packages/sswitch_runtime/SimpleSwitch.py
+usr/lib/python*/site-packages/sswitch_runtime/__init__.py
+usr/lib/python*/site-packages/sswitch_runtime/constants.py
+usr/lib/python*/site-packages/sswitch_runtime/ttypes.py
```
jafingerhut, line 54: I am not sure, but believe that the default Ubuntu installs of Python3 ignore the site-packages directory, using its own dist-packages directory instead.  Good to test an install of this at least to see if it can find these packages.

build-deb
dreibh: Apache 2.0 is fine as well. I just updated the scripts' headers.

debian/control
dreibh: The ARM64 builds pass the unit tests, which run as part of the package building procedure. This works for Ubuntu 20.04, 20.10 and 21.04. However, I did not make further ARM64 tests besides having the packages build for ARM64 in my Ubuntu Launchpad PPA.

debian/control
dreibh: I just removed the "Recommended" part.

debian/python3-bmv2.install
dreibh: This is the location where the debian/rules "install" target puts these files. It should be the same as for a regular build with "make install DESTDIR=/usr".

PR #993, opened 2021-04-09
Add a command line option to enable dropping packets in v1model ingress

jafingerhut:
... unless you explicitly assign a value to standard_metadata.egress_spec

### DIFF
--- a/targets/simple_switch/bm/simple_switch/runner.h
+++ b/targets/simple_switch/bm/simple_switch/runner.h
@@ -39,7 +39,8 @@ class SimpleSwitchRunner {
   static constexpr uint32_t default_drop_port = 511;
 
   explicit SimpleSwitchRunner(uint32_t cpu_port = 0,
-                              uint32_t drop_port = default_drop_port);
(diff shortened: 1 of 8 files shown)


antoninbas (commented):
Each field has a `get_written_to` method to detect if a field was set during packet processing. This may be a better approach as it would not actually change the value of the egress spec field (one could imagine a program testing the value of egress spec ). It also allows someone to still use 0 as a valid port number value if they wish to. What do you think @jafingerhut?

jafingerhut:
I am not sure I understand one of your comments.  With these changes, 0 is still a perfectly valid port number.  One can assign 0 to stdmeta.egress_spec in their P4 program during ingress, and then stdmeta.egress_spec will change from its initial value of drop_port to 0, which should be different values (drop_port default value is 511, but can be changed via a command line option to a different value).

antoninbas:
@jafingerhut you're right. Sorry it was late for me and I was thinking of something else. The suggestion of using `get_written_to` to avoid modifying egress spec is still valid though.

PR #994, opened 2021-04-20
Make dataplane_interface reusable between targets

rst0git:
The `p4::bm::DataplaneInterface` service was introduced with commit 2877c05 (Support for injecting / receiving pkts with gRPC).  It was initially added under `simple_switch_grpc` as this is the first BMv2 target to support P4Runtime. However, this service is not target dependent and it could be reused with other targets (e.g., `psa_switch`).


PI/Makefile.am
```diff
@@ -28,3 +28,77 @@ src/device_state.cpp
 lib_LTLIBRARIES = libbmpi.la
 
 nobase_include_HEADERS = bm/PI/pi.h
+
+# dataplane_interface.proto
```
antoninbas, line 32: I don't think this should go under the PI/ folder. It's really orthogonal.
I would be fine with a new directory at the top-level, although I am not sure about the name. Maybe "services", or "api"/"apis", with the possibility of moving other RPC service definitions there in the future.

configure.ac
```diff
@@ -216,6 +216,23 @@ AC_LINK_IFELSE(
...
 
+# Check for protobuf compiler
+AC_PATH_PROG([PROTOC], [protoc], [])
+AS_IF([test "x$PROTOC" = x], [AC_MSG_ERROR([protoc not found])])
+
+# Check for gRPC protocol buffers compiler plugins
+AC_PATH_PROG([GRPC_CPP_PLUGIN], [grpc_cpp_plugin])
+AS_IF([test "x$GRPC_CPP_PLUGIN" = x], [
+    AC_MSG_ERROR([grpc_cpp_plugin not found])
+])
```
antoninbas, line 227: If gRPC is a requirement now, I think the README needs to be updated

antoninbas (commented):
Thanks for working on this

configure.ac
rst0git: I dropped this change for now and I will address this is issue in a different PR.

PI/Makefile.am
rst0git: Thank you @antoninbas for your feedback. I've updated the PR to use "services" instead of PI/.

antoninbas (approved):
LGTM

antoninbas:
@rst0git could you look at the CI failures?

rst0git:
Thanks, I've updated the pull request and all of the tests should pass now.

PR #997, opened 2021-04-24
Use Python 3 with simple_switch_grpc

rst0git:
Support for Python 3 was introduced with commit 13d7267 and most of the occurrences of `python` have been replaced with `python3` with commit a633059. This patch updates the autoconf configuration in simple_switch_grpc to use Python 3 as well.

### DIFF
--- a/targets/simple_switch_grpc/configure.ac
+++ b/targets/simple_switch_grpc/configure.ac
@@ -9,7 +9,7 @@ AC_CONFIG_HEADERS([config.h])
 AC_PROG_CC
 AC_PROG_CXX
 
-AM_PATH_PYTHON([2.7],, [:])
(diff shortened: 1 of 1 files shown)


jafingerhut:
@rst0git Basic question from someone who does not know autoconf (i.e. me).  I see you replaced Python version 2.7 with 3.5.  Will that also find any later version of Python3, e.g. Python3.8, that is installed on the system?  If so, great.

rst0git:
> Will that also find any later version of Python3, e.g. Python3.8, that is installed on the system?

Yes, the first argument (3.5) is the [minimum version](https://www.gnu.org/software/automake/manual/html_node/Python.html#index-AM_005fPATH_005fPYTHON) of Python required.

antoninbas (approved):

PR #998, opened 2021-04-24
Improve CI build

rst0git:
This pull request replaces the hard-coded number of make jobs (`-j2`) with a number corresponding to the processing units available. This allows `docker build` to complete faster on systems with more than 2 CPU cores. In addition, the verbosity of `apt-get` has been reduced to avoid logging messages such as download progress in Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,6 @@ install:
 
 script:
   - ci_env=`bash <(curl -s https://codecov.io/env)`
-  - docker run $ci_env --env GCOV -w /behavioral-model bm /bin/bash -c "make check -j$(nproc) && ./travis/codecov.sh"
-  - if [ "$sswitch_grpc" = "yes" ]; then docker run -w /behavioral-model/targets/simple_switch_grpc bm make check -j$(nproc); fi
+  - docker run --rm $ci_env --env GCOV -w /behavioral-model bm /bin/bash -c "make check -j$(nproc) && ./travis/codecov.sh"
+  - if [ "$sswitch_grpc" = "yes" ]; then docker run --rm -w /behavioral-model/targets/simple_switch_grpc bm make check -j$(nproc); fi
   - bash tools/check_style.sh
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -2,11 +2,6 @@ ARG PARENT_VERSION=latest
 FROM p4lang/pi:${PARENT_VERSION}
 LABEL maintainer="Antonin Bas <antonin@barefootnetworks.com>"
 
-# Default to using 2 make jobs, which is a good default for CI. If you're
-# building locally or you know there are more cores available, you may want to
-# override this.
-ARG MAKEFLAGS=-j2
```
antoninbas, line 8: do you think you could update https://github.com/p4lang/behavioral-model/blob/main/Dockerfile.noPI as well?

.travis.yml
```diff
@@ -47,6 +47,6 @@ install:
 
 script:
   - ci_env=`bash <(curl -s https://codecov.io/env)`
-  - docker run $ci_env --env GCOV -w /behavioral-model bm /bin/bash -c "make check -j2 && ./travis/codecov.sh"
-  - if [ "$sswitch_grpc" = "yes" ]; then docker run -w /behavioral-model/targets/simple_switch_grpc bm make check -j2; fi
+  - docker run --rm $ci_env --env GCOV -w /behavioral-model bm /bin/bash -c "make check -j$(nproc) && ./travis/codecov.sh"
+  - if [ "$sswitch_grpc" = "yes" ]; then docker run --rm -w /behavioral-model/targets/simple_switch_grpc bm make check -j$(nproc); fi
```
antoninbas, line 51: well the CI VMs are not reused, so I would say `--rm` doesn't add value, but there is probably no harm in adding it either

Dockerfile
rst0git: done

.travis.yml
rst0git: I agree, it is helpful when copy/paste the docker command from CI logs to replicate an error.

antoninbas (approved):

Issue #1001, opened 2021-05-07
Action profile installation fails with "serious errors" due to undefined behavior in Judy library


smolkaj:
I noticed that BMv2 seems to occasionally replicate group members of  `action_profile_action_set`s. Concretely, when installing the following entry (along with several other entries),
```
table_id: 33554499
match {
  field_id: 1
  exact {
    value: "group-4294934549"
  }
}
action {
  action_profile_action_set {
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-8"
        }
      }
      weight: 15
      watch_port: "\r"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-9"
        }
      }
      weight: 14
      watch_port: "\023"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-11"
        }
      }
      weight: 14
      watch_port: "\027"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-12"
        }
      }
      weight: 14
      watch_port: "\033"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-21"
        }
      }
      weight: 14
      watch_port: "-"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-23"
        }
      }
      weight: 14
      watch_port: "1"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-24"
        }
      }
      weight: 14
      watch_port: "5"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-25"
        }
      }
      weight: 14
      watch_port: "7"
    }
    action_profile_actions {
      action {
        action_id: 16777221
        params {
          param_id: 1
          value: "nexthop-27"
        }
      }
      weight: 14
      watch_port: "="
    }
  }
}
```
and then injecting a packet that hits this table entry, I see the following in the BMv2 log:
```
[12:14:40.301] [bmv2] [T] [thread 1142658] [1.0] [cxt 0] Applying table 'ingress.routing.wcmp_group_table'
[...]

PR #1002, opened 2021-05-22
Remove dependency on Judy1

antoninbas:
We have been experiencing issues with some builds of bmv2, with memory
errors in Judy when building it from source (it's possible that it's an
issue with how we build Judy and not with Judy itself...). Judy is very
complex and not actively maintained any more apparently, so it's
probably a good idea to remove the dependency on Judy altogether. In
this commit, we replace usage of Judy1 with alternative implementations:

* HandleMgr now uses a custom implementation, inspired from
  boost::dynamic_bitset.
* RandomAccessUIntSet is now implemented using
  boost::container::flat_set (a sorted vector which exposes a set
  interface). Note that this may introduce a dependency on a new Boost
  library (libboost-container-dev) for some distros.

Originally I was planning on supporting both the Judy-based
implementation and the alternative implementation (with a configure
switch). However I strongly believe that there is no significant
difference in performance, and it doesn't seem necessary to maintain
both versions.

We still need to remove the dependency on JudyL (LPM trie implementation
and BMI) which is likely to be a bit more difficult to do.

See #1001

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -46,7 +46,7 @@ AC_ARG_WITH([pdfixed],
 AM_CONDITIONAL([COND_PDFIXED], [test "$want_pdfixed" = yes])
 
 AC_ARG_WITH([nanomsg],
-    AS_HELP_STRING([--with-nanomsg], [Build Nanomsg RPC service, if disabled then you must have some other way of controlling the switch]),
+    AS_HELP_STRING([--with-nanomsg], [Support generating Nanomsg events]),
     [want_nanomsg="$withval"], [want_nanomsg=yes])
 
 AM_CONDITIONAL([COND_NANOMSG], [test "$want_nanomsg" = yes])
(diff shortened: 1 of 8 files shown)


jafingerhut:
For the new package `libboost-container-dev`, should some install instructions in one of the README files mention adding another package with this or a similar name, before building this project from source code?

antoninbas:
@jafingerhut I didn't update the instructions because these still target Ubuntu 16.04, for which libboost-container-dev does not exist. Installing libboost-dev provides the necessary files in that case. I also did a quick check with Ubuntu 18.04 and Ubuntu 20.04, and it seems that installing libboost-dev is also sufficient there. libboost-container-dev exists for these distributions and according to the docs it is supposed to provide the flat_set container, but in practice it seems installing libboost-dev is sufficient.

jafingerhut:
@antoninbas Thanks for the info.  I see libboost-dev installed in the install_deps.sh script in this repo, so if that covers it, then all should still be well.  I run a full build from latest source of p4c, behavioral-model, PI, and a few other repos every month, so that will catch anything that might have gone amiss, but it sounds like this change should be fine.

PR #1007, opened 2021-06-10
[psa_switch] supporting checksum extern

### LINKED ISSUE #974: bmv2 does not support psa checksum
I'm trying to use bmv2 to compile a psa code with checksum. however, it semms that behavioral-model can not deal with psa checksum. I'm confused because I don't konw whether behavioral-model can'suppoet this checksum or something was wrong in my code. Currently if I comple my code, I got these following error message:
```
parser IngressParserImpl(packet_in buffer,
                         out headers parsed_hdr,
                         inout metadata user_meta,
                         in psa_ingress_parser_input_metadata_t istd,
                         in empty_t resubmit_meta,
                         in empty_t recirculate_meta)
{
    CommonParser() p;
    Checksum<bit<16>>(PSA_HashAlgorithm_t.CRC16) ck;
    state start {
        p.apply(buffer, parsed_hdr, user_meta);

	ck.clear();
	ck.update({ parsed_hdr.ipv4.version,
	      parsed_hdr.ipv4.ihl,
              parsed_hdr.ipv4.diffserv,
              parsed_hdr.ipv4.totalLen,
              parsed_hdr.ipv4.identification,
              parsed_hdr.ipv4.flags,
              parsed_hdr.ipv4.fragOffset,
              parsed_hdr.ipv4.ttl,
              parsed_hdr.ipv4.protocol,
              parsed_hdr.ipv4.srcAddr,
              parsed_hdr.ipv4.dstAddr });
	parsed_hdr.ipv4.hdrChecksum = ck.get();
	ck.clear();

        transition accept;
    }
}
```
```
forwarding_task_psa_untilnow.p4(93): [--Werror=unsupported] error: ck.clear: not supported in parser on this target
 ck.clear();
 ^^^^^^^^^^
forwarding_task_psa_untilnow.p4(94): [--Werror=unsupported] error: ck.update: not supported in parser on this target
 ck.update({ parsed_hdr.ipv4.version,
 ^
In file: /home/myp4/p4/p4c/backends/bmv2/common/expression.cpp:97
Compiler Bug: forwarding_task_psa_untilnow.p4(105): ck_0/ck;: could not convert to Json
 parsed_hdr.ipv4.hdrChecksum = ck.get();
                               ^^
```

what's more, my Compilation options are "p4c-bm2-psa" because "p4c --arch psa" cannot run correctly and compiler shows that "arch psa not implement"
` sudo p4c-bm2-psa forwarding_task_psa_untilnow.p4 -o success.json`

DusanKrdzic:
Fixes #974 in behavioral-model.

*[psa_switch] supporting checksum extern and it's methods

link to p4c pull request: https://github.com/p4lang/p4c/pull/2794

### DIFF
--- a/src/bm_sim/P4Objects.cpp
+++ b/src/bm_sim/P4Objects.cpp
@@ -2168,11 +2168,11 @@ P4Objects::init_objects(std::istream *is,
 
     init_errors(cfg_root);  // parser errors
 
+    init_calculations(cfg_root);
+
     init_parsers(cfg_root, &init_state);
 
     init_deparsers(cfg_root);
-
-    init_calculations(cfg_root);
 
     init_counter_arrays(cfg_root);
 
--- a/targets/psa_switch/externs/psa_checksum.cpp
+++ b/targets/psa_switch/externs/psa_checksum.cpp
@@ -1,0 +1,48 @@
+
+
+#include "psa_checksum.h"
+#include <iostream>
+
+namespace bm {
+
+namespace psa {
+
+void
+PSA_Checksum::init() {
+  this->clear();
+}
+
+void PSA_Checksum:: get(Field& dst) {
+  dst.set(internal.get<uint64_t>());
+}
+
+void PSA_Checksum::get_verify(Field& dst, Field& equOp) {
+  if (equOp.get<uint64_t>()!=internal.get<uint64_t>()) {
+    dst.set(false);
+  } else {
+    dst.set(true);
+  }
+}
+
+void PSA_Checksum::clear() {
+  internal.set(0);
+}
+
+void PSA_Checksum::update(const NamedCalculation& calculation) {
+  const uint64_t cksum = calculation.output(get_packet());
+  this->internal.set(cksum);
+}
+
+BM_REGISTER_EXTERN_W_NAME(Checksum,PSA_Checksum);
+BM_REGISTER_EXTERN_W_NAME_METHOD(Checksum, PSA_Checksum,update, const NamedCalculation&);
+BM_REGISTER_EXTERN_W_NAME_METHOD(Checksum, PSA_Checksum,get, Field &);
+BM_REGISTER_EXTERN_W_NAME_METHOD(Checksum, PSA_Checksum,get_verify, Field &, Field &);
+BM_REGISTER_EXTERN_W_NAME_METHOD(Checksum, PSA_Checksum,clear);
+
+}  // namespace bm::psa
+
+}  // namespace bm
+
+int import_checksums() {
+  return 0;
+}
--- a/targets/psa_switch/externs/psa_checksum.h
+++ b/targets/psa_switch/externs/psa_checksum.h
@@ -1,0 +1,41 @@
+
+
+#ifndef PSA_SWITCH_PSA_CHECKSUM_H_
+#define PSA_SWITCH_PSA_CHECKSUM_H_
+
+#include <bm/bm_sim/extern.h>
+#include <bm/bm_sim/calculations.h>
+
+
+namespace bm {
+
+namespace psa {
+
+class PSA_Checksum : public bm::ExternType {
+ public:
+  static constexpr p4object_id_t spec_id = 0xfffffffd;
+
+  BM_EXTERN_ATTRIBUTES {
+    BM_EXTERN_ATTRIBUTE_ADD(hash);
+  }
+
+  void init() override;
+
+  void get(Field& dst);
+
+  void get_verify(Field& dst, Field& equOp);
+  
+  void clear();
+
+  void update(const NamedCalculation& calculation);
+
+ private:
+  std::string hash;
+  Data internal;
+
+};
+
+}  // namespace bm::psa
+
+}  // namespace bm
+#endif
--- a/targets/psa_switch/Makefile.am
+++ b/targets/psa_switch/Makefile.am
@@ -8,7 +8,7 @@ THRIFT_IDL = $(srcdir)/thrift/psa_switch.thrift
 
 noinst_LTLIBRARIES = libpsaswitch.la
 
-libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp
+libpsaswitch_la_SOURCES = psa_switch.cpp psa_switch.h primitives.cpp externs/psa_counter.h externs/psa_counter.cpp externs/psa_meter.h externs/psa_meter.cpp externs/psa_checksum.h externs/psa_checksum.cpp
 
 libpsaswitch_la_LIBADD = \
 $(top_builddir)/src/bm_sim/libbmsim.la \
--- a/targets/psa_switch/psa_switch.cpp
+++ b/targets/psa_switch/psa_switch.cpp
@@ -65,6 +65,7 @@ REGISTER_HASH(bmv2_hash);
 extern int import_primitives();
 extern int import_counters();
 extern int import_meters();
+extern int import_checksums();
 
 namespace bm {
 
@@ -181,6 +182,7 @@ PsaSwitch::PsaSwitch(bool enable_swap)
   import_primitives();
   import_counters();
   import_meters();
+  import_checksums();
 }
 
 #define PACKET_LENGTH_REG_IDX 0
--- a/targets/psa_switch/psa_switch.h
+++ b/targets/psa_switch/psa_switch.h
@@ -36,6 +36,7 @@
 
 #include "externs/psa_counter.h"
 #include "externs/psa_meter.h"
+#include "externs/psa_checksum.h"
 
 // TODO(antonin)
 // experimental support for priority queueing


src/bm_sim/P4Objects.cpp
```diff
@@ -2168,12 +2168,12 @@ P4Objects::init_objects(std::istream *is,
...
     init_errors(cfg_root);  // parser errors
 
+    init_calculations(cfg_root);
+
     init_parsers(cfg_root, &init_state);
 
     init_deparsers(cfg_root);
 
-    init_calculations(cfg_root);
-
```
antoninbas, line 2176: what's the reason for this change?

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,48 @@
+
```
antoninbas, line 1: missing apache copyright header

targets/psa_switch/externs/psa_checksum.h
```diff
@@ -0,0 +1,41 @@
+
```
antoninbas, line 1: missing apache copyright header

targets/psa_switch/externs/psa_checksum.h
```diff
@@ -0,0 +1,41 @@
...
+#include <bm/bm_sim/calculations.h>
+
+
+namespace bm {
+
+namespace psa {
+
+class PSA_Checksum : public bm::ExternType {
+ public:
+  static constexpr p4object_id_t spec_id = 0xfffffffd;
```
antoninbas, line 16: unused

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,48 @@
...
+namespace bm {
+
+namespace psa {
+
+void
+PSA_Checksum::init() {
+  this->clear();
+}
+
+void PSA_Checksum:: get(Field& dst) {
```
antoninbas, line 15: extra whitespace

targets/psa_switch/externs/psa_checksum.h
```diff
@@ -0,0 +1,41 @@
...
+
+  BM_EXTERN_ATTRIBUTES {
+    BM_EXTERN_ATTRIBUTE_ADD(hash);
+  }
+
+  void init() override;
+
+  void get(Field& dst);
+
+  void get_verify(Field& dst, Field& equOp);
```
antoninbas, line 26: looks like these could be `const` methods

targets/psa_switch/externs/psa_checksum.h
```diff
@@ -0,0 +1,41 @@
...
+  
+  void clear();
+
+  void update(const NamedCalculation& calculation);
+
+ private:
+  std::string hash;
```
antoninbas, line 33: this is not used anywhere, which is confusing

targets/psa_switch/externs/psa_checksum.h
```diff
@@ -0,0 +1,41 @@
...
+
+  void init() override;
+
+  void get(Field& dst);
+
+  void get_verify(Field& dst, Field& equOp);
+  
+  void clear();
+
+  void update(const NamedCalculation& calculation);
```
antoninbas, line 30: The PSA `update` extern method takes data to add to the checksum (incremental checksum computation), but this takes a `NamedCalculation` object which also includes a hash algorithm. It seems that the hash algorithm is a property of the extern instance, not of the calculation. Could you comment on the implementation choices in the code (there isn't much info to that effect in the commit message)?

src/bm_sim/P4Objects.cpp
DusanKrdzic: Because in *init_parsers* or *init deparsers* we can have Chekcsum update primitive which needs a calculation object as parameter to compute checksum.

targets/psa_switch/externs/psa_checksum.cpp
DusanKrdzic: Yes. I am sorry. I will delete id.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: Yes. I will change it.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: Yes. Actually it represents the parameter of psa_checksum constructor from psa.p4. But that information is used in p4c when calculation is made in JSON for update parameter. So here it is not necessary.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: In p4c actual `update` parameters are encapsulated in calculation instance. So I made a calculation instance in JSON (in calculations block) that have information about update parameters (input field in calculation JSON) and about hash algorithm that this checksum instance use. So this parameter NamedCalculation is actually that parameter from JSON and it has information about hash algorithm and actual update parameters. It use hash algorithm which is defined in Checksum constructor on actual upadate parameters and calculate the checksum. I implemented like this because I thought it would be nice to use something that is already in code. BM project already has Calculation classes (NamedCalculation) where are hash algorithms already implemented and it has methods that are only used for calculating checksums. I didn't want to write something that already exists.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: Yes. I will delete it. I thought I would use it but I didn't.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: When we use something like this in P4 code parsers:
	`ck.update({ parsed_hdr.ipv4.version,
                parsed_hdr.ipv4.ihl,
                parsed_hdr.ipv4.diffserv,
                parsed_hdr.ipv4.totalLen,
                parsed_hdr.ipv4.identification,
                parsed_hdr.ipv4.flags,
                parsed_hdr.ipv4.fragOffset,
                parsed_hdr.ipv4.ttl,
                parsed_hdr.ipv4.protocol,
                parsed_hdr.ipv4.srcAddr,
                parsed_hdr.ipv4.dstAddr });`
The JSON representation i `parsers` block is:
                  `"op" : "_Checksum_update",
                  "parameters" : [
                    {
                      "type" : "extern",
                      "value" : "IngressParserImpl.ck"
                    },
                    {
                      "type" : "calculation",
                      "value" : "calc"
                    }
                  ]`
Where calculation `calc` is reference to:
`{
      "name" : "calc",
      "id" : 0,
      "algo" : "crc16",
      "input" : [
        {
          "type" : "field",
          "value" : ["ipv4", "version"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "ihl"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "diffserv"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "totalLen"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "identification"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "flags"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "fragOffset"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "ttl"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "protocol"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "srcAddr"]
        },
        {
          "type" : "field",
          "value" : ["ipv4", "dstAddr"]
        }
      ]
    }`
In `calculations` JSON block.

targets/psa_switch/externs/psa_checksum.cpp
DusanKrdzic: Yes, I am sorry. I will add it.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: Yes, I am sorry. I will add it.

DusanKrdzic:
I made changes that have been requested. I added `license`, `const` to `get` methods, and removed unused variables. Please, tell me if there is anything else to do so this pr can be merged.

targets/psa_switch/externs/psa_checksum.cpp
antoninbas: are you making this contribution on behalf of Cornell University? I'm asking because that's the copyright you are using but it doesn't match your email address / Github profile.

On a side note, I also found out recently that `<YEAR>-present` is not a valid mention, so please just use `YEAR`

targets/psa_switch/externs/psa_checksum.h
antoninbas: Thanks for all the detailed information. My only concern was that the hash function should ideally be a property of the extern. However, I understand that in the absence of a `field_list` construct in the bmv2 JSON, it is convenient to use a `calculation` for this. I think this is fine as an initial implementation, which can be refined later if we do introduce a `field_list` construct.

antoninbas: although thinking about this a bit more (and remembering why I was opposed to adding a `field_list` construct to bmv2 in the first place), it seems that using a `parameters_vector` of integral values (search for `parameters_vector` in https://github.com/p4lang/behavioral-model/blob/main/docs/JSON_format.md) would have made more sense. I won't block the PR for this since these changes are limited to the PSA target, but I think a `parameters_vector` is the right approach as: 1) it may be applicable to other extern types as well, 2) `NamedCalculation` are basically legacy objects that predate P4_16, PSA and p4c.

targets/psa_switch/externs/psa_checksum.cpp
DusanKrdzic: Thank you for paying me attention to this. I am making this contribution on behalf of SYRMIA LCC. I have just changed the copyright.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: `Is this implementation with NamedCalculation good enough for merging?` I understand the solution with `parameters_vector`. But there are a few problems with that and that solution is not so straightforward. First hash algorithms and other structures that are used for checksum calculations are implemented in `calculations.cpp` in anonymous namespace so it is not available from other files. So, it is necessary to implement again the same code in `psa_checksum.cpp` file, or to change anonymous namespace (I don't know if is it ok to change it.). Second, if we use `paramameters_vector` in JSON, `PSA_Checksum::update` method need to look like this: `update(const std::vector<Data> fileds) `. JSON will look like: 
parameters_vector
So when we have package like this (`IPv4`): 4(version)5(ihl)00(diffserv)0038(totallen)...(in hex format), the elements of vector `fields` (which represent those fields `version`, `ihl` and so on) will have values 4,5,0,56 (decimal format). I think that there is no way to find the actual size of fields in `update` method, for example `diffserv`. Here it has value 0 which is 1B and actual size is 2B. Because of this, there is no way in method `update `to represent those fields as one single data to calculate checksum. This problem is solved with `Calculation` because when `calculations` are parsed in JSON it use BufBuilder and other structures to resolve this problem (BufBuilder saves info about header, offset).  When `parameters_vector` is parsed in JSON it only takes actual fields values and represents them as `Data`. So, in method `update` with vector<Data> I only have actual values. There is no information about actual field (size, header, field name etc.).

DusanKrdzic: Also, `PSA_Checksum::update` shouldn't have vector of fields as parameter because when we use construct `{field1,field2,field3,...}` in `update` in P4, it represents kind of concatenation and those fields are packed in one object. It would be necessary to introduce new JSON construct which would `tie` all fields from `update` JSON in one `Data` in bm. This has already been done with `calculations`. `Calculations` construct was made because of this problem. Only thing that is problematic with `calculation` implementation is that `PSA_Checksum` doesn't have information about `hash` as attribute. But it is not really necessary. Hash is only important for `update`, so in p4c `hash` attribute is taken from extern checksum constructor and given to corresponding calculation to calculate checksum.

antoninbas: I agree with your first point (missing size information is an issue). I am not sure I agree with the second statement though. Ultimately, it's all serialized as one single byte buffer, the structure of the input doesn't matter. It's a valid compiler transformation IMO to flatten the input so it's just a sequence of fields, and only the order matters.

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,69 @@
...
+void
+PSA_Checksum::clear() {
+  internal.set(0);
+}
+
+void
+PSA_Checksum::update(const NamedCalculation& calculation) {
+  const uint64_t cksum = calculation.output(get_packet());
+  this->internal.set(cksum);
+}
```
antoninbas, line 55: @DusanKrdzic I was looking at this one more time and I am not sure this implements the actual PSA specification. `update` is a method you can call multiple times for a given checksum calculation, in order to support incremental computation. Unless I am mistaken, it seems that your current implementation assumes a single call to `update`, as every call overrides the `internal` value. Am I missing something?

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: I have only tried to say that it is important to save actual sizes of fields in package, so the actual value from package could be recreated in `update` or to make new JSON construct that would "tie" the fields in one single data while parsing JSON in `P4Objects`. Parameter `std::vector<Data> fields` doesn't support that.
Maybe the checksum computation in PSA should be done with `Hash` extern, which is used to calculate hashes in PSA. For example `PSA_Checksum` should have extern `Hash` object as attribute. `PSA_Chekcsum` attribute `hash` which is a `string` should be used as `Hash` object's constructor parameter in `PSA_Checksum::init` where it should be initialized. Then in `PSA_Checksum::update`, the extern `Hash` object will be used for calculating checksum over the package. But extern `Hash` isn't supported yet in PSA.
This pull request is about issue number #974 and it fix it. Now user can use extern checksum and its methods to calculate checksum. This implementation has its weakness (hash isn't `PSA_Checksum` attribute in bm) but for actual PSA architecture that is supported, I think this implementation with `NamedCalculation` as `PSA_Checksum::update` parameter is the best approach.

targets/psa_switch/externs/psa_checksum.cpp
DusanKrdzic: This is the `basic checksum` implementation with PSA extern `extern Checksum`. I couldn't find anywhere in the `PSA spec` that the `basic checksum` uses incremental computation to compute checksum. I think it use basic computation as the name says.
There is a special type of checksum in PSA `extern InternetChecksum` which uses incremental computation. It has special method `add` for that kind of calculation.

DusanKrdzic: `Basic checksum -  Checksum`
basic_checksum

`Incremental checksum -  InternetChecksum`
incremental

antoninbas: let's double-check with @jafingerhut.

@jafingerhut I want to make sure what the expected behavior is for the `update` extern method:
1. overwrite the checksum value, using only the data provided in the `update` call
2. append the data to an internal buffer (preserving the data provided in previous `update` calls), the checksum is computed on the entire buffer when `get` is called

My understanding of the spec is that the second one is the expected behavior:
```
  /// Add data to checksum
  void update<T>(in T data);

  /// Get checksum for data added (and not removed) since last clear
  W    get();
```

The "and not removed" is a bit confusing, since this extern doesn't support removing data.

DusanKrdzic: My understanding of the spec is that the first is expecting behavior because it is the basic checksum. If basic checksum also uses incremental computation why is extern InternetChecksum needed? I understood that this update is some kind of computation that is used in simple switch which uses basic checksum computation. In PSA spec it says "PSA also provides an incremental checksum" after defining the basic checksum. So there must be a difference in computation.

antoninbas: Maybe incremental is not the right term for what the basic checksum does. It doesn't have a `remove` method. With the `InternetChecksum`, if the value of a field changes, you can update the checksum by removing the old value and adding the new value. That's not possible with `Checksum`, but I still believe the intent is to support multiple `update` calls. Anyway, I will wait for @jafingerhut to chime in.

jafingerhut: InternetChecksum is explicitly incremental, not only for adding more data, but also for removing data, so that it can even be used for implementing things like updating TCP and UDP header checksums in certain situations.  There are both add and subtract methods for InternetChecksum for this very reason.

Antonin, I agree that the "and not removed" part is confusing, and to me it is not terribly explicit in the latest PSA spec whether behavior 1 or 2 is being specified.  It has been about 3 years since there were discussions on this topic, and I do not now recall any discussions that would favor interpretation 1 over 2, or vice versa.  Let me check on the details of at least one PSA implementation of Checksum extern in this regard and report back.

DusanKrdzic: If it is not terribly explicit why it shouldn't be the first interpretation:
`1. overwrite the checksum value, using only the data provided in the update call`
because it has already been implemented with this pull-request. 
I think that the name `basic` defines that it uses the first interpretation. If someone wants to use incremental computation then there is special extern type `InternetChecksum` for that kind of computation.

antoninbas: @DusanKrdzic "because this PR does it this way" is not a good enough reason. Let's wait to hear back from Andy on the status of current implementations.

@jafingerhut we should probably update the spec and the method comments for `update` and `get`.

jafingerhut: @antoninbas Agreed that the spec should be clarified here, regardless of the answer.

DusanKrdzic: To improve the computation of large chunk of data such as in TCP, RFC 1071 (https://datatracker.ietf.org/doc/html/rfc1071) introduced `InternetChecksum` with incremental computing with `1’s complement` algorithm which is especially used for incremental computation. This technique is covered in RFC 1071 and further polished over two other RFCs: RFC 1141 (https://datatracker.ietf.org/doc/html/rfc1141) and RFC 1624 (https://datatracker.ietf.org/doc/html/rfc1624).
For the same reason `InternetChecksum` is also introduced in PSA as incremental checksum with `1's complement` to perform incremental computation. `InternetChecksum` uses `only` `1's complement`  algorithm because of explicit incremental computation (like in RFC).
In PSA, `Basic checksum` can use all other hash algorithms which are not explicitly used for incremental computation.
This is one of the reasons why I think that there is a difference in computation between `basic` and `internet` checksum.
If `basic` checksum also uses incremental computation why is it needed? Then `InternetChecksum` would do the same even much more (because of `substract` and other methods that `basic` checksum doesn't have). In this case `basic` checksum would be a kind of subset of `InternetChecksum`. This would make `basic` checksum redundant.

targets/psa_switch/externs/psa_checksum.h
DusanKrdzic: I made new implementation for `PSA_Checksum::update` as you requested @antoninbas. Now, PSA_Checksum::update method has  declaration: `update(const std::vector<Field> fields)`. 
JSON looks like: 
fields_list
PSA Checksum JSON instance has information about its `hash` attribute in constructor.
`PSA_Checksum` class in bm has the `hash`attribute. In method `update`, checksum will be calculated based on which hash the current checksum instance uses. The hash implementations are used from `calculations.cpp`.

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,181 @@
...
+std::string prepareDataForHash(bm::Data &input) {
+  std::string hex_ascii;
+  std::string hex_hex;
+  const uint8_t base = 16;
+
+  while (input != 0) {
+    hex_ascii += hex_digits[input % base];
+    input.divide(input, base);
+  }
+  std::reverse(hex_ascii.begin(), hex_ascii.end());
```
DusanKrdzic, line 43: Conversion from decimal format to hex string. For example `451` will be converted to "1C3".

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  if(hex_ascii.size() % 2 != 0) {
+    hex_ascii.insert(hex_ascii.size()-1, 1, '0');
+  }
```
DusanKrdzic, line 47: For odd number of digits in hex string insert '0'. For example "1C3" -> "1C03". This is important because hash algorithm takes 1B (2 hex digits). "1C3" will be processed as 0x1C 0x03 in hash algorithm.

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  for (size_t i = 0; i < hex_ascii.size(); i+=2) {
+    std::string hex_byte;
+    hex_byte += hex_ascii[i];
+    hex_byte += hex_ascii[i+1];
+    uint8_t num = convertHexStrToU8(hex_byte);
+    hex_hex += num;
+  }
```
DusanKrdzic, line 55: Hash algorithms process the data in chunks of bytes (chars). If the hex value from input is "4F", it semantically means 0x4F. If "4F" is directly passed to hash algorithm, it will process it as 0x34 0x46. So, it is important to convert the hex value 0x4F to ASCII char which is 'O'. Then the hash algorithm will process it as 0x4F.

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,181 @@
...
+    hex_ascii.insert(hex_ascii.size()-1, 1, '0');
+  }
+
+  for (size_t i = 0; i < hex_ascii.size(); i += 2) {
+    std::string hex_byte;
+    hex_byte += hex_ascii[i];
+    hex_byte += hex_ascii[i+1];
+    uint8_t num = convertHexStrToU8(hex_byte);
+    hex_hex += num;
+  }
```
DusanKrdzic, line 55: Hash algorithms process the data in chunks of bytes (chars). If the hex value from input is "4F", it semantically means 0x4F. If "4F" is directly passed to hash algorithm, it will process it as 0x34 0x46. So, it is important to convert "4F" to char with value 0x4F in ASCII table, which is 'O'. Then the hash algorithm will process it as 0x4F.

targets/psa_switch/externs/psa_checksum.cpp
DusanKrdzic, line 43: Conversion from decimal format to hex string. For example 451 in dec will be converted to "1C3" in hex. It is necessary because .stf packet is in hex format and user expects hash from hex value.

targets/psa_switch/externs/psa_checksum.cpp
DusanKrdzic, line 47: For odd number of digits in hex string insert '0'. For example "1C3" -> "1C03". This is important because hash algorithm takes 1B (2 hex digits). "1C3" will be processed as 0x1C 0x03 in hash algorithm.

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,188 @@
...
+  std::string hex_hex;
+  const uint8_t base = 16;
+  uint8_t num_hex_digits = 0;
+
+  while (input != 0) {
+    hex_ascii += hex_digits[input % base];
+    num_hex_digits++;
+    input.divide(input, base);
+  }
+  std::reverse(hex_ascii.begin(), hex_ascii.end());
```
DusanKrdzic, line 45: Conversion from decimal format to hex string. For example 451 in dec will be converted to "1C3" in hex. It is necessary because .stf packet is in hex format and user expects hash from hex value.

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  for (size_t i = 0; i < hex_ascii.size(); i += 2) {
+    std::string hex_byte;
+    hex_byte += hex_ascii[i];
+    if(i != hex_ascii.size() - 1)
+      hex_byte += hex_ascii[i+1];
+    uint8_t num = convertHexStrToU8(hex_byte);
+    hex_hex += num;
+  }
```
DusanKrdzic, line 59: Hash algorithms process the data in chunks of bytes (chars). If the hex value from input is "4F", it semantically means 0x4F. If "4F" is directly passed to hash algorithm, it will process it as 0x34 0x46. So, it is important to convert "4F" to char with value 0x4F in ASCII table, which is 'O'. Then the hash algorithm will process it as 0x4F.

DusanKrdzic:
@antoninbas Is there anything else you want me to do in order to fix and merge this pr?

targets/psa_switch/externs/psa_checksum.cpp
```diff
@@ -0,0 +1,186 @@
...
+  // So, it is important to convert "4F" to char with value 0x4F in ASCII table,
+  // which is 'O'. Then the hash algorithm will process it as 0x4F.
+  for (size_t i = 0; i < hex_ascii.size(); i += 2) {
+    std::string hex_byte;
+    hex_byte += hex_ascii[i];
+    if(i != hex_ascii.size() - 1)
+      hex_byte += hex_ascii[i + 1];
+    uint8_t num = convert_hex_str_to_u8(hex_byte);
+    hex_hex += num;
+  }
```
DusanKrdzic, line 80: Hash algorithms process the data in chunks of bytes (chars). If the hex value from input is "4F", it semantically means 0x4F. If "4F" is directly passed to hash algorithm, it will process it as 0x34 0x46. So, it is important to convert "4F" to char with value 0x4F in ASCII table, which is 'O'. Then the hash algorithm will process it as 0x4F.

antoninbas:
@DusanKrdzic I have been waiting to see if @jafingerhut can give us an answer as to what the expected behavior is. Maybe this is something he had an opportunity to bring up at a recent PSA meeting?

DusanKrdzic:
@antoninbas  @jafingerhut  I updated `the basic Checksum` implementation PR and resolved conflicts with `the InternetChecksum` implementation.
Maybe you determined the behavior of the `update` method?
I still think it should be basic behavior, which I explained in this comment: https://github.com/p4lang/behavioral-model/pull/1007#discussion_r654763590

DusanKrdzic:
@antoninbas @jafingerhut  Any updates?

Issue #917, opened 2020-06-24
_constants.cpp: No such file or directory


fruffy:
Just a heads up for anyone using the latest version of Thrift. There has been a recent merge (https://github.com/apache/thrift/pull/2095) that omits the generation of `_types.cpp `and `_constants.cpp` files. I had to edit all the automake makefiles that include these (for example, in [pdfixed](https://github.com/p4lang/behavioral-model/blob/master/pdfixed/Makefile.am#L8)). After removing these includes the behavioral models compiles fine again. This may not be the right way to fix this or may not become an issue but might be useful for anyone that encounters the same problem in the meantime.

bluecmd:
Given that the thrift used in this repo is pretty old by now, would it make sense to upgrade to relying on the latest Thrift these days? That way people can use the thrift version installed by their up-to-date OSes.

jafingerhut:
@bluecmd Have you tried and successfully built the latest BMv2 code and other open source P4 dev tools with a recent version of Thrift?  I have no idea how strongly the Thrift project maintains backwards compatibility, but at least with the portion of the Thrift APIs that this project uses, it would be good to find out.  If you are interested, there is a bash script that I maintain that can install thrift, grpc, p4lang/PI, p4lang/behavioral-model, p4lang/p4c, and mininet, on which the tutorials basic exercise runs without problems, at least, which you could tweak to use more recent versions of Thrift:

+ stand-alone script that runs on Ubuntu 18.04 or 20.04 Desktop Linux, AMD64 arch, from fresh install: https://github.com/jafingerhut/p4-guide/blob/master/bin/README-install-troubleshooting.md in particular the install-p4dev-v4.sh script, not the older install-p4dev-v2.sh one mentioned there.

+ Vagrantfile and other files that can be used on a host OS to create an Ubuntu 20.04 Desktop Linux VM with the same versions of the P4 source code: https://github.com/jafingerhut/tutorials/tree/add-2021-mar-vm-based-on-ubuntu-20.04. (note that this is on a branch named add-2021-mar-vm-based-on-ubuntu-20.04, not the main branch, because it hasn't been merged into the main branch yet, as it is still somewhat early in its testing)

bluecmd:
I am currently building `simple_switch` to use together with `mininet`, but that's about it - I will see if it works in the end.
[...]

PR #1031, opened 2021-08-24
Fix build with recent versions of Thrift.

blp:
Recent versions of Thrift (such as the one that comes with recent Fedora
distros) don't generate "_constants" files for C++, so the build fails.
Empty files work fine, so this commit fixes the problem by creating them.

Fixes bug #917.


antoninbas (approved):
LGTM, thanks for the patch Ben

PR #1033, opened 2021-09-03
[psa_switch] Supporting Hash extern

DusanKrdzic:
* [psa_switch] supporting Hash extern and its methods.
* Based on PSA spec, pages 31-32: https://p4.org/p4-spec/docs/PSA.pdf
* link to p4c pull request: https://github.com/p4lang/p4c/pull/2896


targets/psa_switch/externs/psa_hash.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::string buf = prepare_data_for_algo(fields);
+    hash = compute(buf.data(), buf.size());
+
+    uint64_t result = base.get<uint64_t>() + (hash % max.get<uint64_t>());
+
+    dst.set(result);
+}
+
+uint64_t
+PSA_Hash::compute(const char *buf, size_t s) {
```
antoninbas, line 140: I disagree with the current approach. I may be missing something but I think this could be implemented very easily, and without hardcoding the algorithms name anywhere in the extern implementation:
```cpp
void
PSA_Hash::init() {
    // using this->calc to emphasize that calc would be a class member
    this->calc = bm::CalculationsMap::get_instance()->get_copy(algo);
}

uint64_t
PSA_Hash::compute(const char *buf, size_t s) {
    return this->calc->output(buf, s);
}
```

With this approach I would expect most of the changes to the core `bm_sim` files to not be necessary?

DusanKrdzic: I've done this with hardcoding the algorithms name because the algo names for the same algorithm are different in PSA arch and in `bm_sim` (beacauce in `bm_sim` algos are for `v1_model`)current implementation. But I will convert names in p4c so I could do this in this new approach.

DusanKrdzic:
@antoninbas  I've made changes that have been requested.

targets/psa_switch/externs/psa_hash.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // For an odd number of hex digits.
+    // 0x456 -> 0x45 0x06
+    if (n_bits % 8 != 0) {
+        help.bit_and(input, mask4);
+        buf.insert(buf.begin(), 1, help.get<uint8_t>());
+        input.shift_right(input, 4);
+        n_bits -= 4;
+    }
```
antoninbas, line 49: not sure I understand this, but maybe it's because of the comment?
* comment says "For an odd number of hex digits" but then you test for an odd number of bits
* "0x456 -> 0x45 0x06" just doesn't seem like a correct transformation. I see no reason to modify the last byte in the buffer ever.

DusanKrdzic: I've found this byte extraction approach in some documentation and CRC-16 calculator on the Internet. It extracts bytes from left to right. But I agree that it doesn't seem like a correct transformation. I will remove it.

targets/psa_switch/externs/psa_hash.cpp
```diff
@@ -0,0 +1,118 @@
...
+ *
+ */
+
+
+#include "psa_hash.h"
+
+namespace {
+
+std::string
+prepare_data_for_algo(const std::vector<bm::Field> &fields) {
```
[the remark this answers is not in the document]
DusanKrdzic: I know about this approach and have already tried it this way. However, it does not work for fields whose length is indivisible by 8. e.g. bit<12> a = 0x456 will give hash `0xb832` which is not correct. It should be `0xfe82`. 
If you think that field length must be divisible with 8, I will push new implementation with `BufBuilder` help class

DusanKrdzic:
@antoninbas I've made changes that have been requested. If you still think that it should be done with `BufBuilder` class, I will push that implementation.

targets/psa_switch/externs/psa_hash.cpp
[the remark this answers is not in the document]
antoninbas: `BufBuilder` pads to right. It's possible that CRCs expects padding to the left (at least that's what I see here: https://stackoverflow.com/questions/3411654/best-way-to-generate-crc8-16-when-input-is-odd-number-of-bits-not-byte-c-or-p). But what about non-CRC hash algorithms? I guess padding to the left is reasonable, so I am fine with modifying the implementation to do that. As long as `field.deparse` is used to avoid big numbers or custom code it works for me.

That may mean iterating over the field list twice: once to compute the required length and once to deparse (serialize) fields. But that's probably not a big issue, as it also lets us allocate the required memory ahead of time. Hope that works:

```cpp
int nbits = 0;
for (const auto &field : fields) {
  nbits += field.get_nbits();
}
int nbytes = (nbits + 7) / 8;
ByteContainer buf(nbytes); // 0-initialized
char *ptr = buf->data();
nbits = (nbytes * 8 - nbits);  // pad to the left with 0s
for (const auto &field : fields) {
  ptr += (nbits + 7) / 8;
  field.deparse(ptr, nbits % 8);
  nbits += field.get_nbits();
}
compute(buf.data(), buf.size());
```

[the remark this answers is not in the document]
DusanKrdzic: I just tried this approach. It doesn't seem to work. It calculates bad hash again now even for bit size divisible by 8. I’m not sure that 2 `Data` vars could degrade performance that much. They make code simpler, readable and it works what is important.

DusanKrdzic:
@antoninbas I implemented as required: `ByteContainer` used, big numbers avoided.

targets/psa_switch/externs/psa_hash.cpp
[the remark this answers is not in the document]
DusanKrdzic: I modified this approach a bit and now it works.

DusanKrdzic:
@antoninbas Is there anything else you want me to do in order to fix and merge this pr?

targets/psa_switch/externs/psa_hash.cpp
```diff
@@ -0,0 +1,88 @@
...
+ *
+ */
+
+
+#include "psa_hash.h"
+
+namespace {
+
+void
+build_buffer(const std::vector<bm::Field> &fields, bm::ByteContainer &buf) {
```
antoninbas, line 26: according to the coding style guide for this repo, we use pointers for out parameters (parameters that are modified by the function), so it should be `bm::ByteContainer *buf`. However, it seems that here `build-buffer` could return the `bm::ByteContainer` object directly, benefiting from NRVO? `bm::ByteContainer build_buffer(const std::vector<bm::Field> &fields)`?

targets/psa_switch/externs/psa_hash.cpp
```diff
@@ -0,0 +1,88 @@
...
+  int nbits = 0;
+  int nbytes;
+  for (const auto &field : fields) {
+    nbits += field.get_nbits();
+  }
+  nbytes = (nbits + 7) / 8;
```
antoninbas, line 32: since we know the number of bytes needed for storage at this point, could we resize the buffer a single time here? otherwise we potentially allocate memory and move data several times

antoninbas (commented):
sorry for the delay, I took some time off recently

targets/psa_switch/externs/psa_hash.cpp
DusanKrdzic: Ok. I will do in that way.

targets/psa_switch/externs/psa_hash.cpp
antoninbas: Ok we can merge it like this, and I can take a look later

DusanKrdzic:
@antoninbas I've made changes that have been requested.

antoninbas (approved):

PR #1042, opened 2021-10-06
Stop using libJudy in BMI library

antoninbas:
Instead we use a sorted vector to store port information.
This is part of the effort to remove any dependency on libJudy.
See #1001.

### DIFF
--- a/src/BMI/bmi_port.c
+++ b/src/BMI/bmi_port.c
@@ -1,4 +1,5 @@
 /* Copyright 2013-present Barefoot Networks, Inc.
+ * Copyright 2021 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


src/BMI/bmi_port.c
```diff
@@ -68,21 +67,56 @@ static inline int port_in_use(bmi_port_t *port) {
 }
 
 static inline bmi_port_t *get_port(bmi_port_mgr_t *port_mgr, int port_num) {
-  Word_t *pinfo;
-  JLG(pinfo, port_mgr->ports_map, (Word_t) port_num);
```
rst0git, line 72: `JLG` is also used in `src/bf_lpm_trie/bf_lpm_trie.c`.
Is it possible to remove the dependency there as well?

PR #1045, opened 2021-10-08
Remove remaining references to libJudy in scripts and docs

antoninbas:


antoninbas:
Going to hold off on merging this until the libJudy dependency can be removed from p4lang/PI as well

PR #1050, opened 2021-11-09
[v1model] Doc bit width mismatch: enq_qdepth and deq_qdepth

yunheL:
In simple_switch.md: `enq_qdepth` and `deq_qdepth` are documented as `16` bits.
In [p4c/p4include/v1model.p4](https://github.com/p4lang/p4c/blob/main/p4include/v1model.p4#L97): `enq_qdepth` and `deq_qdepth` are `bit<19>`. 

Thanks for reviewing.

### DIFF
--- a/docs/simple_switch.md
+++ b/docs/simple_switch.md
@@ -166,9 +166,9 @@ recommend that you use the following P4_14 code:
 header_type queueing_metadata_t {
     fields {
         enq_timestamp : 48;
-        enq_qdepth : 16;
+        enq_qdepth : 19;
         deq_timedelta : 32;
-        deq_qdepth : 16;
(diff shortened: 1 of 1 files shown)


antoninbas (approved):
I don't think there is any harm in merging this, even though the documentation section you are updating here is specific to P4_14, not P4_16 (so, not directly related to v1model.p4).

antoninbas:
Ignoring Travis failure, as this is a doc change

yunheL:
> I don't think there is any harm in merging this, even though the documentation section you are updating here is specific to P4_14, not P4_16 (so, not directly related to v1model.p4).

I see. Thanks for clarifying! :-)

PR #1057, opened 2021-12-12
Updates to documentation on metdata preservation after recent p4c cha…

jafingerhut:
…nges

In particular after this pull request was merged into p4c:
https://github.com/p4lang/p4c/pull/2902

### DIFF
--- a/docs/simple_switch.md
+++ b/docs/simple_switch.md
@@ -50,8 +50,7 @@ Here are the fields:
   ingress port on which the packet arrived to the device.  Read only.
 - `packet_length` (sm14, v1m) - For new packets from a port, or
   recirculated packets, the length of the packet in bytes.  For cloned
-  or resubmitted packets, you may need to include this in a list of
-  fields to preserve, otherwise its value will become 0.
+  or resubmitted packets, its value will become 0.
 - `egress_spec` (sm14, v1m) - Can be assigned a value in ingress code to
   control which output port a packet will go to.  The P4_14 primitive
   `drop`, and the v1model primitive action `mark_to_drop`, have the side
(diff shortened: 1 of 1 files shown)


jafingerhut:
I moved the section named "Restrictions on recirculate, resubmit, and clone operations" to the end, since it will become older and older history as time goes on.  I can move it back to a place that makes the changes easier to review if that is preferable.

jafingerhut:
Pinging @smolkaj @kheradmandG who were involved in the p4c changes that led to these documentation changes, to see if they have any comments or suggestions here.

docs/simple_switch.md
```diff
@@ -50,8 +50,7 @@ Here are the fields:
   ingress port on which the packet arrived to the device.  Read only.
 - `packet_length` (sm14, v1m) - For new packets from a port, or
   recirculated packets, the length of the packet in bytes.  For cloned
-  or resubmitted packets, you may need to include this in a list of
-  fields to preserve, otherwise its value will become 0.
+  or resubmitted packets, its value will become 0.
```
antoninbas, line 53: is that true?

this seems to contradict what you write below:

> Each cloned packet will have the its standard_metadata fields initialized in the same way that all packets that begin egress processing do, e.g. egress_port, egress_spec, egress_global_timestamp, etc.

I also see the field being set there: https://github.com/p4lang/behavioral-model/blob/982dec26720dee35735939ed699e8510d5f9f768/targets/simple_switch/simple_switch.cpp#L674-L675

I haven't looked at the code in a while, maybe it doesn't work as expected?

smolkaj:
Based on a quick skim, this looks good. Thanks for updating this!

docs/simple_switch.md
jafingerhut: I can try writing a little test program to check this.  I assumed the statement was correct when I wrote it before, but perhaps I was wrong when I wrote it before, and it has changed since then.  If anyone beats me to writing a test program to check what the actual behavior is, please add a comment, otherwise I'll probably get to it over the winter holidays some time.

jafingerhut: I ran an existing p4c test program I wrote a few years ago to exercise many packet paths in the v1model architecture, and found the results reported in this issue: https://github.com/p4lang/behavioral-model/issues/1062

jafingerhut: I will update this PR with a small change in this area to bring it up to date with Anton's recent changes in this PR: https://github.com/p4lang/behavioral-model/pull/1063

jafingerhut: I have updated the documentation of packet_length.  Nice when the documentation gets even shorter :-)

antoninbas (approved):
LGTM, thanks for updating this Andy

PR #1060, opened 2021-12-14
travis: update branch name in deploy stage

rst0git:
The deploy stage in travis should update the content of http://bmv2.org, however it looks like it doesn't work currently because the `master` branch has been renamed to `main`.


antoninbas (approved):
Thanks for spotting this

PR #1061, opened 2021-12-14
bmi_port: Fix memory leak and incorrect return type

rst0git:
This pull request fixes a memory leak and incorrect return type in `src/BMI/bmi_port.c` reported by `cppcheck`.

### DIFF
--- a/src/BMI/bmi_port.c
+++ b/src/BMI/bmi_port.c
@@ -186,9 +186,11 @@ int bmi_start_mgr(bmi_port_mgr_t* port_mgr) {
 }
 
 int bmi_port_create_mgr(bmi_port_mgr_t **port_mgr, int max_port_count) {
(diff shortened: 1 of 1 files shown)


src/BMI/bmi_port.c
```diff
@@ -188,7 +188,10 @@ int bmi_start_mgr(bmi_port_mgr_t* port_mgr) {
 int bmi_port_create_mgr(bmi_port_mgr_t **port_mgr, int max_port_count) {
   bmi_port_mgr_t *port_mgr_ = malloc(sizeof(bmi_port_mgr_t));
   int exitCode;
-  if (!port_mgr) return -1;
```
antoninbas, line 191: looks like this could just be moved before the `malloc` call?

rst0git: Thanks, I've updated the commit.

antoninbas (approved):

PR #1063, opened 2021-12-15
[simple_switch] Fix standard_metadata.packet_length for copies

### LINKED ISSUE #1062: v1model packet_length standard metadata field sometimes 0 or a value that might be uninitialized garbage
I added a line `#define ENABLE_DEBUG_TABLES` to this test program in the p4c repo: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/v1model-special-ops-bmv2.p4

There is an STF test for that test (which is not very simple to understand and follow, I know) in that same directory.  If you build p4c from source code, then run these commands, it will create a directory p4c/build/tmp<some_random_string> that contains a simple_switch log file with many details.

```
cd p4c/build
./bmv2/testdata/p4_16_samples/v1model-special-ops-bmv2.p4.test -b -v
```

With the latest versions of p4c and behavioral-model code:

Good news:
+ new input packets that were unicast to egress had a reasonable looking value of packet_length
+ recirculated packets had the packet_length field equal to the Ethernet frame length in bytes
+ packets cloned during ingress appeared to have a correct packet_length when the clone was processed next in egress

Less good news, if you care about the packet_length field in these cases:
+ resubmitted packets had the packet_length field equal to 0 in ingress, even though the packet_length was equal to the length of the Ethernet frame length in bytes when the packet first arrived.
+ packets cloned during egress appeared to have some garbage value in the packet_length field when the clone was processed next in egress

If someone wanted to fix the problematic looking cases, I would suspect that all of the changes would be in this file: https://github.com/p4lang/behavioral-model/blob/main/targets/simple_switch/simple_switch.cpp

You would probably want to create a simpler P4 program and test packets than the one mentioned above to assist you in verifying whether you had fixed the issue.

antoninbas:
Set the field correctly for resubmitted and I2E packet copies.

Fixes #1062

### DIFF
--- a/targets/simple_switch/simple_switch.cpp
+++ b/targets/simple_switch/simple_switch.cpp
@@ -1,4 +1,5 @@
 /* Copyright 2013-present Barefoot Networks, Inc.
+ * Copyright 2021 VMware, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,7 +15,7 @@
  */
 
 /*
- * Antonin Bas (antonin@barefootnetworks.com)
+ * Antonin Bas
  *
  */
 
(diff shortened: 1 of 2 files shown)


jafingerhut:
I re-ran the p4c STF test mentioned in this issue: https://github.com/p4lang/behavioral-model/issues/1062 with these changes to behavioral-model, and the packet lengths do look consistently good in all cases tested now.

jafingerhut:
I also ran all tests included with p4c with these behavioral-model changes, and there were no failures, so all good there, too.

Issue #1071, opened 2022-01-02
Idle timeout notifications for unmatched entries seem more frequent than recommend by P4Runtime API spec


jafingerhut:
I created a simple PTF test for a P4_16+v1model architecture program with one table that supports idle timeout.  You can find the P4 program and PTF test code here: https://github.com/jafingerhut/p4-guide/tree/master/ptf-tests/idletimeout

In that test, I configured the idle timeout for a table entry to be 5 seconds.  After sending in a few packets that matched the entry, the controller stops and waits for idle timeout notifications.  One does appear about 5 to 6 seconds after the last packet matches the entry, as expected.  After that the test continues waiting for another 10 seconds, and the switch sends another idle timeout notification for the entry about once every 2 seconds.

This is far better than never sending another idle timeout notification.  I am only creating the issue because the P4Runtime API specification recommends that the later idle timeout notifications occur once every TTL, i.e. 5 seconds in my test case, according to the following sentence in Section 16.4 (see https://p4.org/p4-spec/p4runtime/v1.3.0/P4Runtime-Spec.html#sec-table-idle-timeout-notification ):

"After generating an idle notification, the P4Runtime server must “reset” the timer for the corresponding entry, which means a new notification will be generated after another TTL if the entry is not hit"

antoninbas:
I agree that the behavior is a bit surprising to someone familiar with the P4Runtime spec. However, it's a bit tricky to fix properly IMO. I believe that the purpose of the spec is to ensure that if a notification doesn't make it to the controller, there will eventually be a new notification generated. The current implementation does ensure that since there will be a new notification generated after 2 seconds (which is twice the default sweep interval). So I am tempted to leave it as it is...

jafingerhut:
Agreed that this does not seem like a serious issue.  Maybe I can simply add a note to the docs for bmv2 explaining the existing behavior, then?

antoninbas:
sounds good to me, you may have to start a new doc

jafingerhut:
OK.  I was just thinking of adding a new section somewhere in this file, which already has lots of bmv2+v1model-specific details and limitations: https://github.com/p4lang/behavioral-model/blob/main/docs/simple_switch.md

antoninbas:
A new section in this doc (for simple_switch_grpc) works for me

antoninbas:
See https://github.com/p4lang/behavioral-model/pull/1074

PR #1074, opened 2022-01-05
Add some details on how BMv2 v1model implements idle timeout option o…

jafingerhut:
…n tables

### DIFF
--- a/docs/simple_switch.md
+++ b/docs/simple_switch.md
@@ -1069,17 +1069,20 @@ of IdleTimeoutNotification messages sent from the switch to a
 P4Runtime controller when a table entry has not been matched for
 longer than its configured `idle_timeout_ns` duration.
 
+Note that the P4Runtime API server/switch side is only implemented by
+the simple_switch_grpc process, not by simple_switch.
+
 The BMv2 v1model implementation of this feature does a "background
 sweep" of all entries in all tables with `support_timeout = true`
-every 2 seconds, so effectively idle timeout configurations are
-rounded up to the next multiple of 2 seconds.
+every 1 second, so effectively idle timeout configurations are rounded
+up to the next multiple of 1 second.
 
 If a table entry has not been matched for long enough, and thus sends
 an IdleTimeoutNotification message to the controller for that entry,
 and the entry continues not to be matched after that, BMv2 will
 attempt to send another IdleTimeoutNotification for the same entry
-every 2 seconds afterwards, regardless of the `idle_timeout_ns`
-configured for the entry.
+every 2 seconds afterwards (every other sweep interval), regardless of
+the `idle_timeout_ns` configured for the entry.
 
 
 ## Restrictions on recirculate, resubmit, and clone operations
--- a/docs/simple_switch.md
+++ b/docs/simple_switch.md
@@ -34,10 +34,13 @@ our knowledge is not supported by other switches, unless they are
 based on simple_switch.
 
 The simple_switch_grpc process was developed based on simple_switch,
-and supports both the P4Runtime API, as a server/switch, and the
-simple_switch Thrift API.  The Thrift API is still supported because
-it supports some configuration changes that the P4Runtime API does
-not, e.g. the `set_queue_rate` command (among others).
+and supports the P4Runtime API, as a server/switch.
+simple_switch_grpc can optionally also support the simple_switch
+Thrift API (see
+[here](https://github.com/p4lang/behavioral-model/tree/main/targets/simple_switch_grpc#enabling-the-thrift-server)).
+Enabling this option might be useful to someone because it supports
+some configuration changes that the P4Runtime API does not, e.g. the
+`set_queue_rate` command (among others).
 
 Thus if you wish to use the P4Runtime API to control the configuration
 of the switch, you must use simple_switch_grpc.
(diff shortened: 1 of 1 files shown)


jafingerhut:
Intended to resolve this issue: https://github.com/p4lang/behavioral-model/issues/1071

Antonin, note that some things I have written in this PR are my best guesses.  If you want me to do more double checking to verify those guesses, let me know.  If you do that double-checking, and have any corrections to make to my wrong guesses, I am happy to update this PR.

docs/simple_switch.md
```diff
@@ -1031,6 +1031,57 @@ options:
...
+in the P4Runtime Specification for how a controller can configure the
+time interval for each entry it adds to such a table, and the contents
+of IdleTimeoutNotification messages sent from the switch to a
+P4Runtime controller when a table entry has not been matched for
+longer than its configured `idle_timeout_ns` duration.
+
+The BMv2 v1model implementation of this feature does a "background
+sweep" of all entries in all tables with `support_timeout = true`
+every 2 seconds, so effectively idle timeout configurations are
+rounded up to the next multiple of 2 seconds.
```
antoninbas, line 1075: technically it sweeps every second, but it won't send a notification for the same entry in 2 consecutive sweeps. I am not totally sure what the rationale was for implementing it this way... Maybe account for slower controllers and give them time to act as intended (delete the entry).

The end result is the same though: a notification is sent every 2 seconds until the entry is matched or the entry is deleted.

docs/simple_switch.md
```diff
@@ -1031,6 +1031,57 @@ options:
...
+
+See Sections
+["Idle-timeout"](https://p4.org/p4-spec/p4runtime/v1.3.0/P4Runtime-Spec.html#sec-idle-timeout)
+and ["Table Idle Timeout
+Notification"](https://p4.org/p4-spec/p4runtime/v1.3.0/P4Runtime-Spec.html#sec-table-idle-timeout-notification)
+in the P4Runtime Specification for how a controller can configure the
+time interval for each entry it adds to such a table, and the contents
+of IdleTimeoutNotification messages sent from the switch to a
+P4Runtime controller when a table entry has not been matched for
+longer than its configured `idle_timeout_ns` duration.
```
antoninbas, line 1070: I wonder if this doc should make it clear that only simple_switch_grpc exposes a P4Runtime API?

docs/simple_switch.md
jafingerhut: I have attempted to make the text closer to your description of how things work.

docs/simple_switch.md
jafingerhut: I have added an explicit mention that only simple_switch_grpc implements the P4Runtime API, not simple_switch.

docs/simple_switch.md
```diff
@@ -1031,6 +1031,60 @@ options:
...
+
+Note that the P4Runtime API server/switch side is only implemented by
+the simple_switch_grpc process, not by simple_switch.
```
antoninbas, line 1073: We may want to add this to the beginning of the doc as a general disclaimer. I think you have been mentioning P4Runtime in a couple of other places. What do you think?

docs/simple_switch.md
```diff
@@ -1031,6 +1031,60 @@ options:
...
+
+The BMv2 v1model implementation of this feature does a "background
+sweep" of all entries in all tables with `support_timeout = true`
+every 1 second, so effectively idle timeout configurations are rounded
+up to the next multiple of 1 second.
+
+If a table entry has not been matched for long enough, and thus sends
+an IdleTimeoutNotification message to the controller for that entry,
```
antoninbas, line 1081: that doesn't sound grammatically correct? maybe "and thus BMv2 has sent an IdleTimeoutNotification message to the controller for that entry"

docs/simple_switch.md
jafingerhut: I have removed this sentence where you commented on it, and added a section near the beginning of the document that explicitly describes control plane API differences between simple_switch and simple_switch_grpc

docs/simple_switch.md
jafingerhut: Edited to use your suggested wording, which does sound better.

docs/simple_switch.md
```diff
@@ -25,6 +25,24 @@ separate executable program named `psa_switch`, separate from the
...
+## Differences between simple_switch and simple_switch_grpc processes
+
+The simple_switch process was developed first, and supports a
+Thrift-based control API that was custom developed for BMv2, and to
+our knowledge is not supported by other switches, unless they are
+based on simple_switch.
+
+The simple_switch_grpc process was developed based on simple_switch,
+and supports both the P4Runtime API, as a server/switch, and the
+simple_switch Thrift API.  The Thrift API is still supported because
```
antoninbas, line 38: This new section is great, but I think we should make it sure that Thrift is _not_ available by default in simple_switch_grpc to avoid confusion: https://github.com/p4lang/behavioral-model/tree/main/targets/simple_switch_grpc#enabling-the-thrift-server

jafingerhut: Eventually you will teach me to say less, rather than more incorrect things, in my documentation, but not today :-). Thanks for the correction.

antoninbas (approved):
LGTM

jafingerhut:
If you want this merged, I don't know who has approval to do so other than you, Antonin :-)

antoninbas:
@jafingerhut I was just waiting for CI to complete, even though it's only a doc change :)

PR #1077, opened 2022-01-11
Log an error in case of mismatch between device ids

antoninbas:
It is pretty easy to provide a different device id for
simple_switch_grpc (default is 0) and when using P4Runtime tools
(default is usually 1). In that case one gets a cryptic "Error when
trying to assign device" P4Runtime error. This log message should
help...


jafingerhut (approved):
LGTM

PR #1079, opened 2022-01-12
Do not ignore drop port in simple_switch_grpc

antoninbas:
The command-line argument was never used when instantiating the switch...


jafingerhut (approved):
I have not tested it, but the change looks reasonable.

Issue #995, opened 2021-04-22
HashAlgorithm - xor16


machadoatz:
Hi,

I was playing with the Hash algorithms defined in [v1model.p4](https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4) and I noticed that I could not use the `xor16`

`enum HashAlgorithm {
    crc32,
    crc32_custom,
    crc16,
    crc16_custom,
    random,
    identity,
    csum16,
    xor16
}`

I was trying to run this piece of code:

` hash(xor_value, HashAlgorithm.xor16,(bit<32>) MIN, {hdr.ipv4.srcAddr,hdr.tcp.srcPort,meta.originalPort},(bit<32>)MAX); `

I did not test crc16_custom and crc32_custom but all the others algorithms seem to be working fine. The reported errors are a mininet message saying: 

> "P4 switch s1 did not start correctly. Check the switch log file."

 and the log message saying 

> "Unknown hash algorithm: xor16"

I tried with multiple versions of the behavior model, including 1.14.0. I am using the repository from [ETH Zurique](https://github.com/nsg-ethz/p4-learning).

machadoatz:
Additionally I found that at the time of this [comment](https://github.com/p4lang/behavioral-model/issues/570#issuecomment-376249477) (March 2018), the xor algorithm was not implemented in bmv2.

antoninbas:
I tagged this as a good first issue. If someone wants to take care of this, it should probably be quite straightforward to implement.

KenCox94:
Can I take this issue?

antoninbas:
> Can I take this issue?

Of course, let me assign it to you. Thanks!

Jithin-K-Sojan:
It looks like this issue has not been resolved yet. Can I take it?

jafingerhut:
@Jithin-K-Sojan I would say go for it!  It would be highly unlikely that the person who claimed it before is just about ready to create a PR now, if they claimed it last May with no further communication since then.

antoninbas:
Done in https://github.com/p4lang/behavioral-model/pull/1081

PR #1081, opened 2022-01-14
Adding the xor16 hash function to bm_sim

Jithin-K-Sojan:
Added the xor16 hash to calculations.cpp.
Added a unit test for the same.

Tested it out with calls to both hash() and update_checksum().

### DIFF
--- a/src/bm_sim/calculations.cpp
+++ b/src/bm_sim/calculations.cpp
@@ -360,6 +360,22 @@ struct csum16 {
   }
 };
 
+struct xor16 {
+  uint16_t operator()(const char *buf, size_t len) const {
+    uint16_t mask = 0x00ff;
+    uint16_t final_xor_value = 0x0000;
+    /* Main loop - 2 bytes at a time */
+    for (unsigned int byte = 0; byte < len; byte += 2) {
+      uint16_t t1 = static_cast<uint16_t>(buf[byte]) << 8;
+      uint16_t t2 = 0x0000;
+      /* In case there is a 2nd byte */
+      if ((byte + 1) < len) t2 = static_cast<uint16_t>(buf[byte + 1]);
+      final_xor_value = final_xor_value ^ (t1 + (t2 & mask));
+    }
+    return final_xor_value;
+  }
+};
+
 struct identity {
   uint64_t operator()(const char *buf, size_t len) const {
     uint64_t res = 0ULL;
@@ -420,6 +436,7 @@ REGISTER_HASH(csum16);
 REGISTER_HASH(identity);
 REGISTER_HASH(round_robin);
 REGISTER_HASH(round_robin_consistent);
+REGISTER_HASH(xor16);
 
 using crc8_custom = crc_custom<uint8_t>;
 REGISTER_HASH(crc8_custom);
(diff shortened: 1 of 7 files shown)


Jithin-K-Sojan:
Addresses issue #995 (HashAlgorithm - xor16)

src/bm_sim/calculations.cpp
```diff
@@ -360,6 +360,22 @@ struct csum16 {
...
+    uint16_t mask = 0x00ff;
+    uint16_t final_xor_value = 0x0000;
+    /* Main loop - 2 bytes at a time */
+    for (unsigned int byte = 0; byte < len; byte += 2) {
+      uint16_t t1 = static_cast<uint16_t>(buf[byte]) << 8;
+      uint16_t t2 = 0x0000;
+      /* In case there is a 2nd byte */
+      if ((byte + 1) < len) t2 = static_cast<uint16_t>(buf[byte + 1]);
+      final_xor_value = final_xor_value ^ (t1 + (t2 & mask));
+    }
```
antoninbas, line 374: Even though this is a xor16 implementation, it's technically possible for len to be odd (e.g. 1, with a single-byte buffer), and you should handle this case (please add a test for this case as well).

antoninbas (changes requested):

src/bm_sim/calculations.cpp
Jithin-K-Sojan: Thanks for the comment.
The case of len being odd is handled. I have added a test for the same in a new commit.

src/bm_sim/calculations.cpp
```diff
@@ -360,6 +360,22 @@ struct csum16 {
...
+struct xor16 {
+  uint16_t operator()(const char *buf, size_t len) const {
+    uint16_t mask = 0x00ff;
+    uint16_t final_xor_value = 0x0000;
+    /* Main loop - 2 bytes at a time */
+    for (unsigned int byte = 0; byte < len; byte += 2) {
+      uint16_t t1 = static_cast<uint16_t>(buf[byte]) << 8;
+      uint16_t t2 = 0x0000;
+      /* Condition returns false on the last byte of an odd-length sequence. */
+      if ((byte + 1) < len) t2 = static_cast<uint16_t>(buf[byte + 1]);
```
antoninbas, line 372: do we need to have this test in the loop body? There is no need to check whether we are at the end for every iteration.
You could write something like:
```cpp
while (len >= 2) {
  // ...
  len -= 2;
}
if (len > 0) {
  // handle last byte
}
```

Jithin-K-Sojan: The loop body has been changed. The check now happens only once after the loop.

antoninbas (commented):
Code lgtm, but please address the style check failures reported by CI

Jithin-K-Sojan:
So the style check failure seems to be coming from test/utils.cpp, the source of the error being the ordering of the header files.
I have resolved the error, but wanted to confirm if it is ok to add it to this PR since it is not related to the rest of the code in the PR.

antoninbas:
@Jithin-K-Sojan the source of the failure is introduced by your PR:
```
/home/travis/build/p4lang/behavioral-model/tools/../src/bm_sim/calculations.cpp:368:  Missing space after ,  [whitespace/comma] [3]
```
I don't see another failure in the Travis logs.

Jithin-K-Sojan:
Thanks @antoninbas the error you mentioned resolved the style check issue.

antoninbas (approved):
LGTM, thanks for contributing this

PR #1083, opened 2022-01-19
Update dates in simple_switch docs to note which things are current

jafingerhut:
as of January 2022, which is nearly all of them, except that the bugs
with preserving metadata for resubmit, recirculate, and clone have
been fixed recently.


jafingerhut:
For all of the limitations mentioned with earlier dates that I have updated to now, I have confirmed with the latest version of p4c for the v1model architecture and bmv2 back end that they are still true as of 2022-Jan-18 version of p4c.

antoninbas (approved):

PR #1105, opened 2022-04-06
Integrate simple_switch_grpc with main configure

rst0git:
This patch integrates the `simple_switch_grpc` target with the main `autogen.sh` and `configure.ac` files. With this change, `simple_switch_grpc` would be compiled when the `--with-pi` option is used. This allows to reuse the automake/autoconf definitions and rules with other BMv2 targets implementing P4Runtime support.

PR #1106, opened 2022-04-08
Add Makefile for services

rst0git:
This pull request moves the automake definitions used to build `dataplane_interface.proto` from `simple_switch_grpc` into `services`.

This change allows to compile the proto files in a target independent and to reuse the dataplane_interface in other targets.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -274,6 +274,45 @@ AC_TYPE_SIZE_T
 AC_TYPE_UINT64_T
 AC_LANG_POP(C++)
 
+AS_IF([test "$want_pi" = yes], [
+	AC_PATH_PROG([PROTOC], [protoc], [])
+	AS_IF([test "x$PROTOC" = x], [AC_MSG_ERROR([protoc not found])])
+
+	PKG_CHECK_MODULES([PROTOBUF], [protobuf >= 3.0.0])
+	AC_SUBST([PROTOBUF_CFLAGS])
+	AC_SUBST([PROTOBUF_LIBS])
+
+	PKG_CHECK_MODULES([GRPC], [grpc++ >= 1.3.0 grpc >= 3.0.0])
+	AC_SUBST([GRPC_CFLAGS])
+	AC_SUBST([GRPC_LIBS])
+
+	AC_PATH_PROG([GRPC_CPP_PLUGIN], [grpc_cpp_plugin])
+	AS_IF([test "x$GRPC_CPP_PLUGIN" = x], [
+		AC_MSG_ERROR([grpc_cpp_plugin not found])
+	])
+
+	AS_IF([test "$PYTHON" != :], [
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -274,21 +274,61 @@ AC_TYPE_SIZE_T
 AC_TYPE_UINT64_T
 AC_LANG_POP(C++)
 
+AS_IF([test "$want_pi" = yes], [
```
antoninbas, line 277: If we are going to end up with all of this here anyway, perhaps my earlier suggestion of using `AC_CONFIG_SUBDIRS` and keeping `targets/simple_switch_grpc/configure.ac` as is doesn't make sense any more?

rst0git: Using `AC_CONFIG_SUBDIRS` makes sense in the context of the previous pull request (https://github.com/p4lang/behavioral-model/pull/1105). It reduces the amount of changes necessary to integrate simple_switch_grpc with `configure.ac` at the top-level directory, and the risk of breaking something or introducing unexpected bugs. Once we add P4Runtime support in psa_switch, we can reuse most of the code from `targets/simple_switch_grpc/configure.ac` for both targets. IMHO reusing common code instead of duplicating it would make it more maintainable and easier to test in the future. If you think this is a good idea, I can add this change in the pull request adding P4Runtime support to psa_switch.

antoninbas: > If you think this is a good idea, I can add this change in the pull request adding P4Runtime support to psa_switch.

That sounds good to me

services/Makefile.am
```diff
@@ -0,0 +1,85 @@
...
+
+AM_CPPFLAGS += -Icpp_out -Igrpc_out \
+-I$(builddir)/cpp_out \
+-I$(builddir)/grpc_out
+
+nodist_libbm_grpc_dataplane_la_SOURCES = $(proto_cpp_files) $(proto_grpc_files)
+# no warnings or Werror for auto-generated sources
+libbm_grpc_dataplane_la_CXXFLAGS =
+
+CLEANFILES = $(BUILT_SOURCES) proto_files.ts
```
antoninbas, line 85: nit: missing new line at the end of file

services/Makefile.am
```diff
@@ -0,0 +1,85 @@
+ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS} -I m4
+
+lib_LTLIBRARIES = libbm_grpc_dataplane.la
+
+EXTRA_DIST = p4/bm/dataplane_interface.proto \
+switch_runner/switch_runner.cpp \
+switch_runner/switch_runner.h
```
antoninbas, line 7: Am I missing something, that doesn't look right?

rst0git: Thank you for catching this. I have another commit that moves `switch_runner` from `targets/simple_switch_grpc/` to `services/` to make it reusable with other targets. I will include this commit in the next pull request.

antoninbas (approved):

PR #1109, opened 2022-04-14
Update Dockerfile.noPI

rst0git:
This pull request removes a few redundant lines in `configure.ac` and moves the declaration for the `HAVE_GRPC_PY_PLUGIN` conditional outside the test for `--with-pi` to ensure that it is always defined. In addition, the `pkg-config` package has been added as dependency in Dockerfile.noPI to enable support for the `PKG_CHECK_MODULES` macro. This fixes the following error:

```
./configure: line 14028: syntax error near unexpected token `PROTOBUF,'
./configure: line 14028: `	PKG_CHECK_MODULES(PROTOBUF, protobuf >= 3.0.0)'
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -297,21 +297,9 @@ AS_IF([test "$want_pi" = yes], [
 			AC_MSG_WARN([grpc_python_plugin not found, Python code won't be generated])
 		])
 	])
-	AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
-
-	AC_PATH_PROG([GRPC_CPP_PLUGIN], [grpc_cpp_plugin])
-	AS_IF([test "x$GRPC_CPP_PLUGIN" = x], [
-		AC_MSG_ERROR([grpc_cpp_plugin not found])
-	])
-
-	AS_IF([test "$PYTHON" != :], [
-		AC_PATH_PROG([GRPC_PY_PLUGIN], [grpc_python_plugin])
-		AS_IF([test "x$GRPC_PY_PLUGIN" = x], [
-			AC_MSG_WARN([grpc_python_plugin not found, Python code won't be generated])
-		])
-	])
-	AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
 ])
+
+AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
 
 # Generate makefiles
 AC_CONFIG_FILES([Makefile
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -290,28 +290,15 @@ AS_IF([test "$want_pi" = yes], [
...
-			AC_MSG_WARN([grpc_python_plugin not found, Python code won't be generated])
-		])
-	])
-	AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
+AS_IF([test "$PYTHON" != :], [
+    AC_PATH_PROG([GRPC_PY_PLUGIN], [grpc_python_plugin])
+    AS_IF([test "x$GRPC_PY_PLUGIN" = x], [
+        AC_MSG_WARN([grpc_python_plugin not found, Python code won't be generated])
+    ])
 ])
```
antoninbas, line 300: I don't think it makes sense not to condition this on `test "$want_pi" = yes`.

rst0git: When the the condition is on `test "$want_pi" = yes`, `./autogen.sh && ./configure` fails with:

```
configure: error: conditional "HAVE_GRPC_PY_PLUGIN" was never defined.
Usually this means the macro was only invoked conditionally.
```

configure.ac
```diff
@@ -297,22 +297,10 @@ AS_IF([test "$want_pi" = yes], [
...
-	AS_IF([test "$PYTHON" != :], [
-		AC_PATH_PROG([GRPC_PY_PLUGIN], [grpc_python_plugin])
-		AS_IF([test "x$GRPC_PY_PLUGIN" = x], [
-			AC_MSG_WARN([grpc_python_plugin not found, Python code won't be generated])
-		])
-	])
-	AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
 ])
 
+AM_CONDITIONAL([HAVE_GRPC_PY_PLUGIN], [test "x$GRPC_PY_PLUGIN" != x])
```
antoninbas, line 302: yes, this is what I had in mind

configure.ac
```diff
@@ -297,22 +297,10 @@ AS_IF([test "$want_pi" = yes], [
...
-	AS_IF([test "x$GRPC_CPP_PLUGIN" = x], [
-		AC_MSG_ERROR([grpc_cpp_plugin not found])
-	])
-
-	AS_IF([test "$PYTHON" != :], [
-		AC_PATH_PROG([GRPC_PY_PLUGIN], [grpc_python_plugin])
-		AS_IF([test "x$GRPC_PY_PLUGIN" = x], [
-			AC_MSG_WARN([grpc_python_plugin not found, Python code won't be generated])
-		])
-	])
```
antoninbas, line 312: hmm, I didn't notice the duplicates before...

antoninbas (approved):

PR #1110, opened 2022-04-14
ci: Upload Doxygen documentation only from p4lang

rst0git:
The GitHub action used to upload the Doxygen documentation should run only in the p4lang/behavioral-model repository.

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -7,8 +7,9 @@ on:
 
 jobs:
   upload-doc:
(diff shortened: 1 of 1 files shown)


.github/workflows/doxygen.yml
```diff
@@ -7,8 +7,9 @@ on:
 
 jobs:
   upload-doc:
+    if: ${{ github.repository == 'p4lang/behavioral-model' && github.ref == 'refs/heads/main' }}
```
antoninbas, line 10: I think the `github.ref` is redundant with the workflow definition, as the workflow only runs for the main branch already. But it is harmless.

antoninbas (approved):

Issue #1131, opened 2022-08-10
Question about adding new hash functions


qobilidop:
We are currently using the `v1model` architecture and the `bmv2` target to build a P4 pipeline. But some hash functions we need are missing from them. In a previous [comment](https://github.com/p4lang/behavioral-model/issues/884#issuecomment-621624830), it's said that to use a new hash function:

> You can either follow the instructions in #570 or you can always manually edit the JSON file generated by the p4c compiler to use the desired hash function instead of "random" or whatever you use in the P4 program.
> 
> There isn't really any other alternative, unless you can convince the P4 Language Design Working Group which owns v1model to include new hash algorithms in the architecture.

We wonder if we could first add the new hash functions to `bmv2` here without also adding them to `v1model` in [p4c](https://github.com/p4lang/p4c)? Because that would enable us to use them in a hacky way at the least. We'll try to push them to `v1model` later. This could make our life much easier, but it's at the risk of not being approved by the P4 Language Design Working Group.

Does this plan sound good?

qobilidop:
Close because the hash functions we need are already added to `bmv2`.

PR #1132, opened 2022-08-14
Add the xor32 hash function

qobilidop:
This PR adds the xor32 hash function, similar to #1081. The background was stated in #1131.


antoninbas (approved):
LGTM, I triggered CI jobs

PR #1134, opened 2022-08-22
Add the Toeplitz hash function

qobilidop:
We would like to add the Toeplitz hash function to BMv2 because we are using BMv2 to model a software NIC, and the Toeplitz hash function is widely used for RSS (Receive-Side Scaling) in NIC. This is another follow-up of #1131. 

The current implementation roughly follows the one in DPDK: 
https://github.com/DPDK/dpdk/blob/72206323a5dd3182b13f61b25a64abdddfee595c/lib/hash/rte_thash.h#L165-L191

The following references are also consulted:
- https://doc.dpdk.org/guides/prog_guide/toeplitz_hash_lib.html
- https://docs.microsoft.com/en-us/windows-hardware/drivers/network/rss-hashing-functions


qobilidop:
I've left several TODOs on things I cannot decide myself alone. I'd like to get some suggestions before moving on to finalize the implementation.

src/bm_sim/calculations.cpp
```diff
@@ -392,6 +392,40 @@ struct xor_hash {
...
+      for (uint32_t input_word = input[j]; input_word; input_word &= (input_word - 1)) {
+        int i = ffs(input_word);
+        uint32_t key_word_l = rss_key[j % rss_key_len] << (31 - i);
+        uint32_t key_word_r = rss_key[(j + 1) % rss_key_len] >> (i + 1);
+        final_xor_value ^= (key_word_l | key_word_r);
+      }
+    }
+
+    // TODO: Decide whether to require len % 4 == 0,
+    // or to handle remaining bytes in some way.
```
antoninbas, line 411: yes, it should ideally accept inputs of any length. I am not familiar with this hash function but maybe 0-padding is the right way to deal with this.

src/bm_sim/calculations.cpp
```diff
@@ -392,6 +392,40 @@ struct xor_hash {
...
+
+    return final_xor_value;
+  }
+
+  // TODO: I'm not sure what would be the best API to configure the RSS key.
```
antoninbas, line 416: check out how arbitrary polynomials are configured for crc_custom, you can do the same for this

src/bm_sim/calculations.cpp
```diff
@@ -392,6 +392,40 @@ struct xor_hash {
...
+  void update_rss_key(const uint32_t *new_rss_key, size_t new_rss_key_len) {
+    rss_key_len = new_rss_key_len;
+    rss_key = new uint32_t(rss_key_len);
+    std::memcpy(rss_key, new_rss_key, rss_key_len * sizeof(uint32_t));
```
antoninbas, line 420: you have a memory leak when the key is updated. Please don't use `new` at all. Just use a `std::vector` for the key.

src/bm_sim/calculations.cpp
```diff
@@ -392,6 +392,40 @@ struct xor_hash {
...
+  }
+
+ private:
+  // TODO: Decide whether to use uint32_t* or unint8_t* for rss_key.
+  uint32_t *rss_key;
+  size_t rss_key_len;
```
antoninbas, line 426: based on what I see above, looks like a `std::vector<uint32_t>` would work fine for your key, unless the key length doesn't need to be a multiple of 4 bytes.

src/bm_sim/calculations.cpp
qobilidop: Done. I went with 0-padding.

src/bm_sim/calculations.cpp
qobilidop: Done.

src/bm_sim/calculations.cpp
qobilidop: Done.

src/bm_sim/calculations.cpp
qobilidop: Done.

qobilidop:
I think the next step would be to add an interface for the runtime CLI to configure the RSS key. I'll dig into this on my end, but if @antoninbas you could point me to some relevant code, that would be very helpful.

antoninbas:
> I think the next step would be to add an interface for the runtime CLI to configure the RSS key. I'll dig into this on my end, but if @antoninbas you could point me to some relevant code, that would be very helpful.

I think I already mentioned `crc_custom` to you, which can be configured at runtime:
* https://github.com/p4lang/behavioral-model/blob/a2e920781f72631159cc4a1bd591787c8b44c8ac/thrift_src/standard.thrift#L703-L713
* https://github.com/p4lang/behavioral-model/blob/a2e920781f72631159cc4a1bd591787c8b44c8ac/include/bm/bm_sim/runtime_interface.h#L399-L407

src/bm_sim/calculations.cpp
```diff
@@ -360,36 +360,110 @@ struct csum16 {
   }
 };
 
+// Concatenate the next n bytes into the specified type.
+// Pad zeros at tail when given bytes are shorter than the type.
+// For example, concat<uint32_t>(&{0xab, 0xcd}, 2) == 0xabcd0000.
+template <typename T>
+T concat(const char *buf, size_t n) {
```
antoninbas, line 367: I don't think `concat` is a great name for this function. It's not doing any concatenation. Maybe `emitAsInteger`?

src/bm_sim/calculations.cpp
```diff
@@ -360,36 +360,110 @@ struct csum16 {
...
+    auto raw_input_words = reinterpret_cast<const uint32_t *>(buf);
+    const size_t n_words = len / 4;
+    const size_t n_bytes_remain = len % 4;
+    for (size_t i_word = 0; i_word < n_words; i_word++) {
+      input_words.push_back(ntohl(raw_input_words[i_word]));
+    }
+    if (n_bytes_remain) {
+      input_words.push_back(
+        concat<uint32_t>(buf + (len - n_bytes_remain), n_bytes_remain));
+    }
```
antoninbas, line 423: it doesn't seem that this needs to be protected by the mutex

include/bm/bm_sim/calculations.h
```diff
@@ -387,6 +387,32 @@ class CustomCrcMgr {
...
+enum class RssErrorCode {
+  SUCCESS = 0,
+  INVALID_CALCULATION_NAME,
+  WRONG_TYPE_CALCULATION,
+  INVALID_KEY,
+};
+
+namespace detail {
+
+using rss_key_t = std::vector<uint8_t>;
```
antoninbas, line 399: could you remind me why this is not `std::vector<uint32_t>`, given that you require `len % 4 != 0`?

qobilidop: My main concern is the ambiguity of endianness. I think it puts less mental burden on users this way. Otherwise, users might need to worry about the endianness of `uint32_t` when trying to set the RSS key.

src/bm_sim/calculations.cpp
qobilidop: I see. I'll remove it then. I didn't truly understand why the mutex was used in `custom_crc` and simply followed their examples here.

src/bm_sim/calculations.cpp
```diff
@@ -360,36 +360,110 @@ struct csum16 {
...
+
+    // Compute hash.
+    uint32_t final_xor_value = 0;
+    const auto key_len = host_rss_key.size();
+    for (size_t i_word = 0; i_word < input_words.size(); i_word++) {
+      // Mod key_len so we could handle longer inputs.
+      const auto& base_key_l = host_rss_key[i_word % key_len];
+      const auto& base_key_r = host_rss_key[(i_word + 1) % key_len];
```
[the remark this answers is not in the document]
qobilidop: I thought there might be a small performance benefit using a const reference instead of a new variable.

src/bm_sim/calculations.cpp
qobilidop: Sure, will rename it.

src/bm_sim/calculations.cpp
antoninbas: to be clear, you need a mutex since another thread can come it and mutate the rss_key
but this specific section of the code (building the `input_words`) is expensive (vector memory allocation) and doesn't need to be protected by the mutex
the mutex is only needed to access rss_key

include/bm/bm_sim/calculations.h
antoninbas: sounds good

src/bm_sim/calculations.cpp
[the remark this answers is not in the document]
antoninbas: doubtful that there is any difference. If anything I'd say it's the opposite, since a reference would basically be a pointer (64-bit address).

src/bm_sim/calculations.cpp
qobilidop: I see. Thanks for your clarification!

src/bm_sim/calculations.cpp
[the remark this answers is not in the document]
qobilidop: That's good to know. Thanks for the knowledge! I'll change them to normal variables then.

include/bm/bm_sim/calculations.h
```diff
@@ -387,6 +387,32 @@ class CustomCrcMgr {
...
+
+std::ostream &operator<<(std::ostream &out, const rss_key_t &rss_key);
+
+}  // namespace detail
+
+class RssMgr {
```
antoninbas, line 405: is this going to be used for anything else? because as a standalone object, I find the name a bit confusing. If it will be specific to the Toeplitz hash function, then the name `ToeplitzRssMgr` would be better. Same for other names (e.g., `set_rss_key` -> `set_toeplitz_rss_key`).

tools/runtime_CLI.py
```diff
@@ -2602,6 +2605,11 @@ def do_set_crc32_parameters(self, line):
     def complete_set_crc32_parameters(self, text, line, start_index, end_index):
         return self._complete_crc(text, 32)
 
+    def do_set_rss_key(self, line):
+        "Change the RSS key for the Toeplitz hash: set_rss_key <name> <rss key in hex string>"
+        name, rss_key = line.split()
+        rss_key = bytearray.fromhex(rss_key)
```
antoninbas, line 2611: I'd suggest doing some extra client-side validation here on the key (e.g., make sure the length is a multiple of 4). That can help provide more helpful error messages.

src/bm_sim/calculations.cpp
```diff
@@ -360,36 +360,110 @@ struct csum16 {
...
+
+  uint32_t operator()(const char *buf, size_t len) const {
+    // Convert input bytes (in network byte order) to host words.
+    // Pad the last one with zeros if it's shorter than a word.
+    std::vector<uint32_t> input_words;
+    auto raw_input_words = reinterpret_cast<const uint32_t *>(buf);
+    const size_t n_words = len / 4;
+    const size_t n_bytes_remain = len % 4;
+    for (size_t i_word = 0; i_word < n_words; i_word++) {
+      input_words.push_back(ntohl(raw_input_words[i_word]));
```
antoninbas, line 416: out-of-curiosity, does the dpdk implementation also require an `ntohl` conversion?

src/bm_sim/calculations.cpp
```diff
@@ -360,36 +360,110 @@ struct csum16 {
   }
 };
 
+// Concatenate the next n bytes and emit as the specified integer type.
```
antoninbas, line 363:
```suggestion
// Emit the next n bytes as the specified integer type.
```

src/bm_sim/calculations.cpp
qobilidop: In DPDK, the input is already in 4-bytes chunks, so they don't need the step to convert from raw bytes. See:

https://github.com/DPDK/dpdk/blob/72206323a5dd3182b13f61b25a64abdddfee595c/lib/hash/rte_thash.h#L177

qobilidop: Also quoting the DPDK doc from https://doc.dpdk.org/guides/prog_guide/toeplitz_hash_lib.html:

> Both of above mentioned \_softrss\_ functions expect the tuple to be in “host” byte order and a multiple of 4 bytes in length.

tools/runtime_CLI.py
qobilidop: Done.

src/bm_sim/calculations.cpp
qobilidop: Done.

include/bm/bm_sim/calculations.h
qobilidop: Done.

src/bm_sim/calculations.cpp
```diff
@@ -507,6 +582,37 @@ template class CustomCrcMgr<uint16_t>;
...
+}
+
+}  // namespace detail
+
+ToeplitzRssErrorCode
+ToeplitzRssMgr::update_key(NamedCalculation *calculation,
+                   const rss_key_t &key) {
+  // TODO(qobilidop): Make the logger work.
+  // Logger::get()->info("Updating RSS key of {} to: {}",
+  //                     calculation->get_name(), key);
```
qobilidop, line 602: I got stuck on this. If I uncomment the logger code, it would result in an error saying:
```console
../../third_party/spdlog/bm/spdlog/sinks/./../details/./format.h:2567:6: error: no match for 'operator<<' (operand types are 'std::__cxx11::basic_ostringstream<char>' and 'const std::vector<unsigned char>')
 2567 |   os << value;
      |   ~~~^~~~~~~~
```

But `operator<<` is overloaded right above. I don't know what's wrong.

include/bm/bm_sim/calculations.h
```diff
@@ -387,6 +387,32 @@ class CustomCrcMgr {
...
+enum class ToeplitzRssErrorCode {
+  SUCCESS = 0,
+  INVALID_CALCULATION_NAME,
+  WRONG_TYPE_CALCULATION,
+  INVALID_KEY,
+};
+
+namespace detail {
+
+using rss_key_t = std::vector<uint8_t>;
```
antoninbas, line 399: s/rss_key_t/toeplitz_rss_key_t

thrift_src/standard.thrift
```diff
@@ -231,6 +231,18 @@ exception InvalidCrcOperation {
  1:CrcErrorCode code
 }
 
+typedef list<i8> BmRssKey
+
+enum RssErrorCode {
```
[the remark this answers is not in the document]
qobilidop: Here we can't do this renaming because it would result in a name collision with `ToeplitzRssErrorCode` defined in `calculations.h`. [For `custom_crc`, `CustomCrcErrorCode` and `CrcErrorCode` are used to avoid this collision in `bm_set_crc32_custom_parameters` in `Standard_servers.cpp`](https://github.com/p4lang/behavioral-model/blob/a2e920781f72631159cc4a1bd591787c8b44c8ac/src/bm_runtime/Standard_server.cpp#L1160-L1174). I followed that example to use the shorthand `RssErrorCode` here. But if you have a better name in mind, please let me know.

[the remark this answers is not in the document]
antoninbas: let's use the `ToeplitzRssHash` prefix in one place and the `ToeplitzRss` prefix in the other

thrift_src/standard.thrift
```diff
@@ -231,6 +231,18 @@ exception InvalidCrcOperation {
  1:CrcErrorCode code
 }
 
+typedef list<i8> BmToeplitzRssKey
+
+enum ToeplitzRssHashErrorCode {
```
antoninbas, line 236: for consistency, everything in the Thrift service should use the `ToeplitzRssHash` / `toeplitz_rss_hash` prefix

include/bm/bm_sim/calculations.h
```diff
@@ -387,6 +387,32 @@ class CustomCrcMgr {
...
+
+using toeplitz_rss_key_t = std::vector<uint8_t>;
+
+std::ostream &operator<<(std::ostream &out, const toeplitz_rss_key_t &rss_key);
+
+}  // namespace detail
+
+class ToeplitzRssMgr {
+ public:
+  using toeplitz_rss_key_t = detail::toeplitz_rss_key_t;
```
antoninbas, line 407: since this is in the scope of `ToeplitzRssMgr`, you could change it to `using rss_key_t = detail::toeplitz_rss_key_t;` if you want

src/bm_sim/calculations.cpp
```diff
@@ -360,36 +360,110 @@ struct csum16 {
...
-    for (size_t offset = 0; offset < len; offset++) {
-      t = (t << 8) + (static_cast<T>(buf[byte + offset]) & mask);
+    final_xor_value ^= emitAsInteger<T>(buf + byte, len);
+
+    return final_xor_value;
+  }
+};
+
+struct toeplitz {
+  using toeplitz_rss_key_t = ToeplitzRssMgr::toeplitz_rss_key_t;
```
antoninbas, line 402: ditto

thrift_src/standard.thrift
antoninbas: although I'm thinking more about this, and I don't see why we use "rss" everywhere. RSS is just one use case for a hash function, and in bmv2 there is nothing tying this hash function to RSS.

So IMO, we should use `toeplitz_hash_` / `ToeplitzHash` for the Thrift service, and `toeplitz_` / `Toeplitz` for the bm_sim core library.

qobilidop: Sure, I'll adopt this naming scheme.

src/bm_sim/calculations.cpp
qobilidop: I ended up using `ByteContainer` to replace `std::vector<uint8_t>`, which already has a `to_hex()` method defined.

antoninbas (approved):
LGTM

PR #1135, opened 2022-08-31
Add lifecycle management for PRs and issues

antoninbas:
PRs and issues will be marked as stale after 6 months, and closed
automatically after another 6 months.

Signed-off-by: Antonin Bas <abas@vmware.com>


rst0git (approved):
LGTM

We might need to wait some time to find out if the workflow works as expected, but a similar lifecycle management workflow is used in Antrea.

Issue #1150, opened 2022-11-02
Mininet integration: Encoding error in Python3 hashlib when opening JSON file


LaumiH:
Hi,

I came across a bug when using the Mininet integration, and I already have a possible fix. 

## Error
In the [example](https://github.com/p4lang/behavioral-model#integrating-with-mininet), the `./runtime_CLI` is executed without arguments, which works fine. However, when adding the JSON description of P4 program using `./runtime_CLI --json <file.json>`, then the command fails with
```
Running ['../p4-behavioral-model/tools/runtime_CLI.py', '--json', '<file.json>'']
Traceback (most recent call last):
  File "../p4-behavioral-model/tools/runtime_CLI.py", line 2640, in <module>
    main()
  File "../p4-behavioral-model/tools/runtime_CLI.py", line 2634, in main
    load_json_config(standard_client, args.json)
  File "../p4-behavioral-model/tools/runtime_CLI.py", line 2622, in load_json_config
    load_json_str(utils.get_json_config(
  File "/home/sdn/p4-behavioral-model/tools/bmpy_utils.py", line 67, in get_json_config
    check_JSON_md5(standard_client, json_path)
  File "/home/sdn/p4-behavioral-model/tools/bmpy_utils.py", line 37, in check_JSON_md5
    m.update(L)
TypeError: Unicode-objects must be encoded before hashing
```

## Reproduce
To reproduce, follow the example using 
- `cd mininet`
- `sudo python3 1sw_demo.py --behavioral-exe ../targets/simple_router/simple_router --json ../targets/simple_router/simple_router.json`

and then execute

- `cd targets/simple_router`
- `./runtime_CLI --json ../targets/simple_router/simple_router.json < commands.txt`.

I might want to add that I am on Ubunut 20.04 using Python 3.8.10, and I have cloned the repo just yesterday.

## Fix
I have traced the error down to `hashlib` requiring an encoded file as input for its `hashlib.md5().update()` method. The fix is trivial, however, I have not tested whether it has any side effects. [In this line](https://github.com/p4lang/behavioral-model/blob/main/tools/bmpy_utils.py#L34), open the file with mode `'rb'`, and that's it.

LaumiH:
Ah, and this error came with Version 1.15.0, the former one works fine (same Ubuntu and Python3 version). I suppose it has something to do with the Python3 changes?

antoninbas:
If you have a patch, please do consider submitting a PR

[...]

PR #1151, opened 2022-11-03
Fix encoding for JSON file when using hashlib

LaumiH:
Signed-off-by: Laura Henning <laura-marie.henning@stud.tu-darmstadt.de>

This PR contributes the tiny change required to solve #1150.


antoninbas (approved):
LGTM

LaumiH:
Apparently I messed up quite a bit, because I changed the wrong open function :D I am so sorry. There are two places where a file is opened, and I changed the wrong one. How do we proceed? Can someone ammend the commit, or should I issue another PR?

Issue #1157, opened 2022-11-19
Weird behavior if drop port is set to 0


tiantianz:
I implemented a very simple p4 program to run on BMv2, and test drop port and cpu port with it.
I set the cpu port to 42 and drop port to 0, installed table entry to route packets to the cpu port, but couldn't receive packets out of the cpu port and the logs say packet was dropped at the end of egress pipeline. (but if I set the drop port to any other number I can successfully receive packets out of the cpu port). 

Is this the intended behavior or a bug of BMv2? Since the documentation doesn't say that drop port cannot be set to 0.

Below are the BMv2 stdout logs:

Calling target program-options parser
Server listening on [2002:afb:afb::]:10015
[23:45:24.886] [bmv2] [D] [thread 3548] Set default default entry for table 'MinimalIngress.egress_port_table': NoAction - 
[23:45:24.892] [bmv2] [D] [thread 3548] simple_switch target has been notified of a config swap
[23:45:24.898] [bmv2] [D] [thread 3548] Entry 0 added to table 'MinimalIngress.egress_port_table'
[23:45:24.898] [bmv2] [D] [thread 3548] Dumping entry 0
Match key:
* in_port             : TERNARY   0001 &&& 01ff
Priority: 2147483646
Action entry: MinimalIngress.set_egress_port - 2a,

[23:45:24.899] [bmv2] [I] [thread 3545] Received packet of length 4 on port 1: 31323334
[23:45:24.900] [bmv2] [D] [thread 3554] [0.0] [cxt 0] Processing packet received on port 1
[23:45:24.900] [bmv2] [D] [thread 3554] [0.0] [cxt 0] Parser 'parser': start
[23:45:24.900] [bmv2] [D] [thread 3554] [0.0] [cxt 0] Parser 'parser' entering state 'start'
[23:45:24.901] [bmv2] [D] [thread 3554] [0.0] [cxt 0] Parser state 'start' has no switch, going to default next state
[23:45:24.901] [bmv2] [T] [thread 3554] [0.0] [cxt 0] Bytes parsed: 0
[23:45:24.901] [bmv2] [D] [thread 3554] [0.0] [cxt 0] Parser 'parser': end
[23:45:24.901] [bmv2] [D] [thread 3554] [0.0] [cxt 0] Pipeline 'ingress': start
[23:45:24.901] [bmv2] [T] [thread 3554] [0.0] [cxt 0] Applying table 'MinimalIngress.egress_port_table'
[...]

PR #1159, opened 2022-11-20
simple_switch: fix egress_spec initialization

rst0git:
When simple_switch_grpc is used with drop port set to 0, it causes unintended behavior of packets being dropped at the end of egress pipeline. This is because by default egress_spec is initialized to 0 and this value happens to match the specified drop port.

Fixes: #1157

### DIFF
--- a/targets/simple_switch/simple_switch.cpp
+++ b/targets/simple_switch/simple_switch.cpp
@@ -670,7 +670,9 @@ SimpleSwitch::egress_thread(size_t worker_id) {
     phv->get_field("standard_metadata.egress_port").set(port);
 
     Field &f_egress_spec = phv->get_field("standard_metadata.egress_spec");
-    f_egress_spec.set(0);
+    // When egress_spec == drop_port the packet will be dropped, thus
+    // here we initialize egress_spec to a value different from drop_port.
+    f_egress_spec.set(drop_port + 1);
 
     phv->get_field("standard_metadata.packet_length").set(
         packet->get_register(RegisterAccess::PACKET_LENGTH_REG_IDX));


jafingerhut (approved):
LGTM

antoninbas (approved):
LGTM

smolkaj:
Thanks for the fix, @rst0git!

targets/simple_switch/simple_switch.cpp
```diff
@@ -670,7 +670,9 @@ SimpleSwitch::egress_thread(size_t worker_id) {
     phv->get_field("standard_metadata.egress_port").set(port);
 
     Field &f_egress_spec = phv->get_field("standard_metadata.egress_spec");
-    f_egress_spec.set(0);
+    // When egress_spec == drop_port the packet will be dropped, thus
+    // here we initialize egress_spec to a value different from drop_port.
+    f_egress_spec.set(drop_port + 1);
```
smolkaj, line 675: Would we need a modulo here to ensure this is a legal port value?
Just a guess, I don't know if that would be needed.

jafingerhut: I believe egress_spec during egress processing, and after egress processing, is _not_ used as a port value at all.  It is simply compared to see if it is equal to, or not equal to, the drop port, at the end of egress.

If you want to be slightly safer, and more backwards-compatible with earlier behavior, you could instead use an assignment like this:
```
f_egress_spec.set((drop_port == 0) ? 1 : 0);
```

antoninbas: the current implementation should be good enough
only requirement is that the initial egress_spec value be different from the chosen drop port

PR #1182, opened 2023-02-15
base_test: fix typedef-name after 'class' error

rst0git:
This pull request fixes the following error, which appears when building BMv2 on Ubuntu 22.04:

```
In file included from base_test.cpp:28:
base_test.h:35:7: error: using typedef-name 'grpc::Channel' after 'class'
   35 | class Channel;
      |       ^~~~~~~
In file included from /usr/include/grpcpp/grpcpp.h:52,
                 from base_test.cpp:21:
/usr/include/grpcpp/channel.h:26:30: note: 'grpc::Channel' has a previous declaration here
   26 | typedef ::grpc_impl::Channel Channel;
      |                              ^~~~~~~
```

This error appears because a typedef [`Channel`](https://github.com/grpc/grpc/commit/40210d3b8a92a285d174322ee9f1e823d619cec3#diff-fb7527f41b6b9035fa3478ad27c113ce9fe63c6dcf4ca62b5e3559cb8eab8d57R26) has been introduced in newer versions of gRPC.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,3 +64,21 @@ jobs:
       run: docker build -t bm-no-pi --build-arg IMAGE_TYPE=test -f Dockerfile.noPI .
     - name: Run tests
       run: docker run --rm -w /behavioral-model bm-no-pi /bin/bash -c "make check -j$(nproc)"
+
+  test-ubuntu22:
+    needs: check-changes
+    if: ${{ needs.check-changes.outputs.has_changes == 'yes' }}
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: 'recursive'
+    - name: Install dependencies
+      run: sudo ./install_deps_ubuntu_22.04.sh
+    - name: Build BMv2
+      run: |
+        ./autogen.sh
+        ./configure --with-pdfixed --with-pi --with-stress-tests --enable-debugger --enable-Werror
+        make -j$(nproc)
+    - name: Run tests
+      run: make check -j$(nproc)
--- a/install_deps_ubuntu_22.04.sh
+++ b/install_deps_ubuntu_22.04.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+set -e
+
+apt-get update
+apt-get install -y curl gnupg
+echo 'deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_22.04/ /' | tee /etc/apt/sources.list.d/home:p4lang.list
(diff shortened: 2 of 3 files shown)


install_deps_ubuntu_22.04.sh
```diff
@@ -0,0 +1,44 @@
...
+    libtool \
+    pkg-config \
+    protobuf-compiler \
+    protobuf-compiler-grpc \
+    python3-all \
+    python3-six \
+    python3-thrift \
+    thrift-compiler \
+    libreadline-dev \
+    p4lang-pi
```
antoninbas, line 42: could you remind me if this package is automatically updated when the main branch of the p4lang/PI repo is updated?
sometimes, some changes to bmv2 require the latest p4lang/PI code

.github/workflows/test.yml
```diff
@@ -64,3 +64,21 @@ jobs:
...
+        submodules: 'recursive'
+    - name: Install dependencies
+      run: sudo ./install_deps_ubuntu_22.04.sh
+    - name: Build BMv2
+      run: |
+        ./autogen.sh
+        ./configure --with-pdfixed --with-pi --with-stress-tests --enable-debugger --enable-Werror
+        make -j$(nproc)
+    - name: Run tests
+      run: sudo make check -j$(nproc)
```
antoninbas, line 84: this is surprising, which tests require superuser privileges?

install_deps_ubuntu_22.04.sh
rst0git: I test and update this package manually as it could occasionally break due to changes in the PI dependencies.
This is to ensure that when users run `apt-get upgrade` the updated version would work as expected.

The build scripts the packages are available at https://github.com/rst0git/p4lang-packages. I'm using these packages in my development environment and intend to continue maintain them for the foreseeable future.

I am happy to apply any changes from the main branch of the p4lang/PI repository and update the package as necessary.

rst0git: We are also using these packages in the CI environment for p4c [^1] and to build the tutorials VM [^2].

[^1]: https://github.com/p4lang/p4c/blob/daa55c88/tools/ci-build.sh#L109
[^2]: https://github.com/p4lang/tutorials/blob/master/vm-ubuntu-20.04/root-release-bootstrap.sh#L43

.github/workflows/test.yml
rst0git: Thank you for pointing this out! I've removed `sudo` as the tests don't require superuser privileges.

install_deps_ubuntu_22.04.sh
antoninbas: it's a bit different, as p4lang/behavioral-model and p4lang/PI are strongly correlated (PI defines some C functions which are implemented in bmv2). But we can see how it goes. It's not like breaking updates are very frequent these days.

antoninbas (approved):

PR #1185, opened 2023-03-14
Add a missing #include <type_traits>

alexfh:
This is necessary to prevent compilation failures with recent Clang and libc++.


smolkaj (approved):
Thanks for the fix.

smolkaj:
@antoninbas could you take a look and merge if this looks good to you? Thanks

antoninbas (approved):

PR #1191, opened 2023-04-26
Patch for FreeBSD, which defines _assert() in standard headers

outpaddling:


include/bm/bm_sim/_assert.h
```diff
@@ -25,14 +25,14 @@
 
 namespace bm {
 
-[[ noreturn ]] void _assert(const char* expr, const char* file, int line);
+[[ noreturn ]] void bm_assert(const char* expr, const char* file, int line);
```
antoninbas, line 28: my c++ is a bit rusty, but how can there be a name conflict, given that `_assert` is declared in the `bm` namespace? Even if you include a header which declares the `_assert` name, it will be in the global namespace.

outpaddling: It's a macro in the header:

#define   _assert(e)      assert(e)

antoninbas: Thanks, that explains it.
Please rename bm_assert to _bm_assert, and sign the CLA.

antoninbas (approved):
Thanks for your contribution

PR #1204, opened 2023-07-28
Fix simple_switch.md: mark_to_drop overrides multicast.

smolkaj:
I empirically observed that mark_to_drop overrides multicast on BMv2.

@jafingerhut @antoninbas @verios-google

### DIFF
--- a/docs/simple_switch.md
+++ b/docs/simple_switch.md
@@ -159,7 +159,8 @@ starts egress processing. The clock is the same as for
 pipeline, but should not be written to.
 - `mcast_grp`: needed for the multicast feature. This field needs to be written
 in the ingress pipeline when you wish the packet to be multicast. A value of 0
-means no multicast. This value must be one of a valid multicast group configured
+means no multicast, and alling `mark_to_drop` will set the value to 0.
+A nonzero value should be that of a valid multicast group configured
 through bmv2 runtime interfaces. See the "after-ingress pseudocode" for
 relative priority of this vs. other possible packet operations at
 end of ingress.
@@ -227,7 +228,8 @@ if (digest to generate) {   // because your code called generate_digest
 }
 if (resubmit was called) {
     start ingress processing over again for the original packet
-} else if (mcast_grp != 0) {  // because your code assigned a value to mcast_grp
+} else if (mcast_grp != 0) {  // because your code assigned a value to mcast_grp and 
+                              // did not call mark_to_drop afterwards
     multicast the packet to the output port(s) configured for group mcast_grp
 } else if (egress_spec == DROP_PORT) {  // e.g. because your code called drop/mark_to_drop
     Drop packet.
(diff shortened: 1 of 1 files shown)


docs/simple_switch.md
```diff
@@ -311,11 +316,6 @@ if (resubmit was called) {
     for the mcast_grp value.  Enqueue each one in the appropriate
     packet buffer queue.  The instance_type of each will be
     PKT_INSTANCE_TYPE_REPLICATION.
-} else if (egress_spec == DROP_PORT) {
-    // This condition will be true if your code called the
-    // mark_to_drop (P4_16) or drop (P4_14) primitive action during
-    // ingress processing.
-    Drop packet.
```
rst0git, line 318: Hi Steffen, I noticed that the ingress thread in [simple_switch.cpp](https://github.com/p4lang/behavioral-model/blob/main/targets/simple_switch/simple_switch.cpp#L619C1-L635C6) implements these conditions as follows:
```C++
    // MULTICAST
    if (mgid != 0) {
      BMLOG_DEBUG_PKT(*packet, "Multicast requested for packet");
      auto &f_instance_type = phv->get_field("standard_metadata.instance_type");
      f_instance_type.set(PKT_INSTANCE_TYPE_REPLICATION);
      multicast(packet.get(), mgid);
      // when doing multicast, we discard the original packet
      continue;
    }

    port_t egress_port = egress_spec;

    if (egress_port == drop_port) {  // drop packet
      BMLOG_DEBUG_PKT(*packet, "Dropping packet at the end of ingress");
      continue;
    }
```
However, the [multicast example](https://github.com/p4lang/tutorials/blob/01ed2c499d88420cff382c5dc9444dbba19067c8/exercises/multicast/solution/multicast.p4#L113) in the tutorials repository demonstrates that `drop()` could be used in the egress pipeline to prune multicast packets.

> I empirically observed that mark_to_drop overrides multicast on BMv2.

Does the observed behaviour occur when mark_to_drop is used in the ingress pipeline?

jafingerhut:
@smolkaj Given the C++ code in `ingress_thread`, I am surprised by your statement.  Do you have a sample v1model P4 program, set of table entries, and an input packet that demonstrates this?

jafingerhut:
I just tried out a test case where for one test, for a packet in ingress I do these operations in this order:
```
        // order of operations #1
        stdmeta.mcast_grp = mcast_grp;
        mark_to_drop(stdmeta);
```
In this case, the packet is dropped at the end of ingress.

Then I tried this order of operations:
```
        // order of operations #2
        mark_to_drop(stdmeta);
        stdmeta.mcast_grp = mcast_grp;
```
In this case, the packet is not dropped at the end of ingress, but is multicast replicated according to the specified mcast_grp.

Fun, eh?

Why?

Because mark_to_drop(stdmeta) is effectively implemented as:
```
    stdmeta.egress_spec = DROP_PORT;
    stdmeta.mcast_grp = 0;     /// NOTE THIS LINE ESPECIALLY !!!
```
See https://github.com/p4lang/behavioral-model/blob/main/targets/simple_switch/primitives.cpp#L176-L203

So I think the pseudocode for what happens at the end of ingress is correct, but you need to be aware of the full behavior of mark_to_drop(stdmeta) in order to explain what is going on.

_Should_ it be this way?  I did not write the original BMv2 code, but the older drop() method also did this, and that goes back to 2016-Jan-18 when the source file primitives.cpp linked above was first added to the p4lang/behavioral-model repository.

jafingerhut:
Based on my previous message, if we wanted to clarify the documentation, we should make explicit exactly what `mark_to_drop` is doing.

docs/simple_switch.md
smolkaj: Yes, I was referring to mark_to_drop in the ingress pipeline.

smolkaj:
Thanks for figuring out what is going on here, @jafingerhut and @rst0git!

> Based on my previous message, if we wanted to clarify the documentation, we should make explicit exactly what mark_to_drop is doing.

I went back to v1model.p4, and actually, this seems to be documented already: http://google3/third_party/p4lang_p4c/p4include/v1model.p4;l=412-413;rcl=425419938

So perhaps there is nothing to do here, except for me reading the documentation more carefully?

> Should it be this way? I did not write the original BMv2 code, but the older drop() method also did this, and that goes back to 2016-Jan-18 when the source file primitives.cpp linked above was first added to the p4lang/behavioral-model repository.

It seems like reasonable behavior to me. I have yet to find out what our switches do, I am curious if they made the same choice there.

jafingerhut:
> It seems like reasonable behavior to me. I have yet to find out what our switches do, I am curious if they made the same choice there.

Good to know you realize it is a choice, one that can be made independently for every switch ASIC in existence.  And some can make it baked into gates, others with some level of configurability, and others much more programmable.  It is a continuum, but probably most switch ASICs are closer to the "some level of configurability, if you know who to ask or where in the documentation to look".

antoninbas:
Are we closing this?

smolkaj:
Sorry for dropping the ball on this. Taking another look at `simple_switch.md` to see if we should clarify this anywhere...

smolkaj:
I'll send a PR shortly.

smolkaj:
How is this?

docs/simple_switch.md
```diff
@@ -300,12 +302,13 @@ if (resubmit was called) {
...
     resubmit operation called.  The resubmitted packet will have
     instance_type equal to PKT_INSTANCE_TYPE_RESUBMIT.
-} else if (mcast_grp != 0) {
+}  else if (mcast_grp != 0) {
     // This condition will be true if your code made an assignment to
-    // standard_metadata.mcast_grp during ingress processing.  There
-    // are no special primitive actions built in to simple_switch for
-    // you to call to do this -- use a normal P4_16 assignment
-    // statement, or P4_14 modify_field() primitive action.
+    // `standard_metadata.mcast_grp`` during ingress processing AND did
```
jafingerhut, line 307: double backquote should probably be single backquote?  And another similar occurrence after modify_field() below.  Also seems like spurious addition of a space before "else if (mcast_grp != 0)" above, unless you think that is an improvement on the whitespace in the code.

smolkaj: Indeed, seems like I was tired :) Should be fixed.

jafingerhut (approved):
LGTM

PR #1209, opened 2023-08-18
Give error when entry is added to table with no key

### LINKED ISSUE #1208: Behavior of tables with empty keys
Following the discussions in https://github.com/p4lang/p4-spec/issues/1120 , newer versions of P416 define the following behavior for tables with empty keys.

> If a table has no key property, or if the value of its key property is the empty tuple, i.e. key = {}, then it contains no look-up table, just a default action—i.e., the associated lookup table is always the empty map.

I have the following table (in a v1model ingress pipeline)
```
table disable_xyz_table {
    key = {}
    actions = {
      @proto_id(1) disable_xyz_action;
    }
  }
```

And in BMv2, using P4Runtime I install an entry with empty set of keys and the action being `disable_xyz_action`  (Note that `is_default_action` is NOT set). 

BMv2's log shows that entry successfully gets installed and the trace shows that the entry gets hit:

```
[17:37:10.785] [bmv2] [T] [thread 5262] [0.0] [cxt 0] Applying table '...disable_xyz_table'
[17:37:10.785] [bmv2] [D] [thread 5262] [0.0] [cxt 0] Looking up key:

[17:37:10.785] [bmv2] [D] [thread 5262] [0.0] [cxt 0] Table '...disable_xyz_table': hit with handle 0
[17:37:10.786] [bmv2] [D] [thread 5262] [0.0] [cxt 0] Dumping entry 0
Match key:
Action entry: ...disable_xyz_action - 

[17:37:10.786] [bmv2] [D] [thread 5262] [0.0] [cxt 0] Action entry is ...disable_xyz_action - 
[17:37:10.786] [bmv2] [T] [thread 5262] [0.0] [cxt 0] Action ...disable_xyz_action
```

Am I missing something here or the BMv2 and P416 are not in agreement in this regard?

antoninbas:
In order to be conformant with recent P4_16 versions (v1.2.4). This breaks backwards-compatibility, but we believe it is unlikely that many programs are relying on this behavior.

Fixes #1208


jafingerhut (approved):
LGTM

PR #1213, opened 2023-09-03
Install a more recent version of nnpy

jafingerhut:


jafingerhut:
I have tested with these changes to install many open source P4 dev tools, including PI, p4c, ptf, mininet, etc. and with these changes, there are some p4lang/p4c tests executed via `make check` that pass with these changes, but fail without them.

rst0git (approved):
LGTM

antoninbas (approved):

PR #1221, opened 2023-10-26
Rewrite build steps for simple_switch_grpc

DolarJoe:
Expand the build steps with specific repos to avoid confusion and add a build step to actually generate/install simple_switch_grpc binary.

### DIFF
--- a/targets/simple_switch_grpc/README.md
+++ b/targets/simple_switch_grpc/README.md
@@ -17,8 +17,8 @@ Github or the p4-dev mailing list.
  2. Configure, build and install PI:
     ```
     git clone https://github.com/p4lang/PI.git
-    git submodule update --init --recursive
     cd PI
+    git submodule update --init --recursive
     ./autogen.sh
     ./configure --with-proto --without-internal-rpc --without-cli --without-bmv2 [--with-sysrepo]
     make
--- a/targets/simple_switch_grpc/README.md
+++ b/targets/simple_switch_grpc/README.md
@@ -33,14 +33,6 @@ Github or the p4-dev mailing list.
     make
     [sudo] make install  # if desired
     ```
- 4. Build the simple_switch_grpc in `behavioral_model/targets/simple_switch_grpc`
-    ```
-    cd ./targets/simple_switch_grpc
-    ./autogen.sh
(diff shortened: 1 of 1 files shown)


targets/simple_switch_grpc/README.md
```diff
@@ -12,22 +12,35 @@ Github or the p4-dev mailing list.
...
  1. Follow instructions in the [PI
     README](https://github.com/p4lang/PI#dependencies) to install required
-    dependencies for the `--with-profo` configure flag - as well as for the
+    dependencies for the `--with-proto` configure flag - as well as for the
     `--with-sysrepo` configure flag if desired.
  2. Configure, build and install PI:
     ```
+    git clone https://github.com/p4lang/PI.git
+    git submodule update --init --recursive
+    cd PI
```
jafingerhut, line 21: Shouldn't the 'cd PI' command be _before_ the 'git submodule update ...' command?

targets/simple_switch_grpc/README.md
```diff
@@ -12,22 +12,35 @@ Github or the p4-dev mailing list.
...
     [sudo] make install  # if desired
     ```
+ 4. Build the simple_switch_grpc in `behavioral_model/targets/simple_switch_grpc`
+    ```
+    cd ./targets/simple_switch_grpc
+    ./autogen.sh
+    ./configure [--with-thrift] [--with-sysrepo]
+    make
+    make install
+    ```
```
jafingerhut, line 43: I publish, and test every month, a script that builds several open source P4 development tools here: https://github.com/jafingerhut/p4-guide/blob/master/bin/README-install-troubleshooting.md

It does the steps at 3. above, and those steps cause `simple_switch_grpc` to be built and installed in /usr/local/bin (if you run `sudo make install`).

Thus I do not understand why you are proposing adding these steps at number 4.  They seem unnecessary to me.

antoninbas: Andy is correct. It used to be necessary to build simple_switch_grpc separately, but it has not been the case for about a year (since https://github.com/p4lang/behavioral-model/pull/1105 was merged).

jafingerhut (approved):
LGTM

antoninbas (approved):

PR #1238, opened 2024-04-05
Update googletest code in third_party

nibblerenush:
Update googletest code to v1.12.1 version from [here](https://github.com/google/googletest/releases/tag/release-1.12.1).


antoninbas (commented):
https://github.com/google/googletest/releases/tag/v1.14.0 mentions that GoogleTest requires at least C++14. This project only requires C++11 and builds with `-std=c+11`. While it seems that CI tests are passing fine, IMO we should still follow one of these 2 approaches:
1) Update to [v1.12.1](https://github.com/google/googletest/releases/tag/release-1.12.1) instead, as it is the last version to support C++11, but should still be much more recent than what we have today
2) Update this project (bmv2) to require C++14. Obviously, this is more involved than approach 1. It would require updating documentation, build scripts, and before making that change, we shoudl probably cut one last bmv2 release with "C++11 support".

nibblerenush:
> 1. Update to [v1.12.1](https://github.com/google/googletest/releases/tag/release-1.12.1) instead, as it is the last version to support C++11, but should still be much more recent than what we have today

I chose this method because I am new to open source.

antoninbas (approved):
LGTM

PR #1240, opened 2024-04-10
Added initial PNA support into BMv2.

rupesh-chiluka-marvell:
To add PNA support into the BMv2. Leveraging the _psa_switch_ target as the foundation, the _pna_nic_ target has been developed.

Basic L2 Forward test has been carried out.
Hardcoded the output port value in transmit function.


rupesh-chiluka-marvell:
Yes. My bad. nanomsg-1.0.0 should not be added. They are built using build scripts.

PR #1244, opened 2024-04-16
Fix #1243.

### LINKED ISSUE #1243: Race conditions when passing in a BMv2 JSON config
After using the Tsan tool on our tests with  a simple grpc service with a [BMV2 JSON config passed in with the command line arguments](https://github.com/p4lang/behavioral-model/issues/1239), I noticed that it resulted in some flaky tests with race conditions. The error follows as below:

```
ThreadSanitizer: data race [p4lang_behavioral_model/src/bm_sim/context.cpp:867:13] in bm::Context::do_swap()
```

I was wondering if this is the expected behavior?

@smolkaj @jonathan-dilorenzo

matthewtlam:
Fix race conditions when passing a BMv2 JSON Config into the command-line arguments when initializing SimpleSwitch.

### DIFF
--- a/src/bm_sim/switch.cpp
+++ b/src/bm_sim/switch.cpp
@@ -75,14 +75,12 @@ SwitchWContexts::receive(port_t port_num, const char *buffer, int len) {
 
 void
 SwitchWContexts::start_and_return() {
-  {
-    std::unique_lock<std::mutex> config_lock(config_mutex);
-    if (!config_loaded && !enable_swap) {
-      Logger::get()->error(
-          "The switch was started with no P4 and config swap is disabled");
-    }
-    config_loaded_cv.wait(config_lock, [this]() { return config_loaded; });
+  std::unique_lock<std::mutex> config_lock(config_mutex);
+  if (!config_loaded && !enable_swap) {
+    Logger::get()->error(
+        "The switch was started with no P4 and config swap is disabled");
   }
+  config_loaded_cv.wait(config_lock, [this]() { return config_loaded; });
   start();  // DevMgr::start
   start_and_return_();
   // Starts any registered periodically-executing externs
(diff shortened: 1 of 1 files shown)


smolkaj (approved):
Thanks!

smolkaj (approved):

src/bm_sim/switch.cpp
```diff
@@ -75,14 +75,12 @@ SwitchWContexts::receive(port_t port_num, const char *buffer, int len) {
...
-    }
-    config_loaded_cv.wait(config_lock, [this]() { return config_loaded; });
+  std::unique_lock<std::mutex> config_lock(config_mutex);
+  if (!config_loaded && !enable_swap) {
+    Logger::get()->error(
+        "The switch was started with no P4 and config swap is disabled");
   }
+  config_loaded_cv.wait(config_lock, [this]() { return config_loaded; });
   start();  // DevMgr::start
   start_and_return_();
```
antoninbas, line 85: A comment here explaining why we are still holding a lock on the `config_mutex` would be nice for future reference.

matthewtlam: Comment would definitely be helpful. Added!

src/bm_sim/switch.cpp
```diff
@@ -80,6 +80,8 @@ SwitchWContexts::start_and_return() {
     Logger::get()->error(
         "The switch was started with no P4 and config swap is disabled");
   }
+  // The lock is held until the config is loaded to avoid race conditions with   
+  // starting the PIGrpcServerRunV2 (P4RT server).
```
antoninbas, line 84: That's not really the complete picture, and this part of the code really has no knowledge of P4Runtime.

I think something like this is more appropriate:

> We keep holding the lock as the start_and_return_ callback which is implemented by targets may access the P4 config for some validations.

antoninbas (approved):

Issue #1245, opened 2024-04-17
To Add PNA support into BMv2.


ghost:
Hello everyone,
 
Currently BMv2 doesn't support PNA. I want to add initial PNA support into BMv2. I have outlined few changes to make PNA integration possible:
•	Updates to context.h, packet.h, and addition of a new file for PNA (just like switch.h) in the _include/bm_sim_ directory.
•	Modification of bm_runtime.h in the _include/bm_runtime_ directory.
•	Creation of few separate server.cpp files to facilitate PNA support in the _src/bm_runtime_ directory.
•	Creation of a new cpp file for PNA (just like switch.cpp) in the _src/bm_sim_ directory.
•	A separate directory containing the core logic and externs of PNA in the _targets_ directory.
•	Adjustments to related makefiles to ensure the new changes are compiled and built correctly.
 
I have submitted a PR in BMv2:  https://github.com/p4lang/behavioral-model/pull/1240 for the initial PNA integration into BMv2.
 
Your feedback is much appreciated.
Need feedback on how to best split the PR.

qobilidop:
Ping @antoninbas @jafingerhut @thantry @drobinkent

qobilidop:
First, thank you @rupesh-chiluka-marvell for your effort on open sourcing this work!

I'm not the most qualified person to give suggestions here, but just want to share some high-level thoughts and hopefully stimulate more discussions that could lead to better and more detailed solutions.

Regarding PR splitting, I think a test-driven approach might work for this process. The idea is to split into a series of PRs with each making a single P4 test program work. For example, the first PR could introduce minimal changes to BMv2 and p4c to enable a minimal P4 PNA program (e.g. single state parser + empty control block) to compile with p4c and run with BMv2. Each subsequent PRs will make the test P4 program a little bit more complicated.

fruffy:
In general, I would clean up the original PRs before we can even start to make suggestions on how to break them down. This means removing nanomsg and thrift.

rupesh-chiluka-marvell:
Created a new PR with just skeleton of `pna_nic`: https://github.com/p4lang/behavioral-model/pull/1255

qobilidop:
@KrisNey-MSFT There is definitely more work to do, and some collaboration is much appreciated!

As to the P4 GSoC project, @rupesh-chiluka-marvell will write a final report and publish it in https://github.com/p4lang/gsoc soon. That could probably be a good starting point for further discussions.

[...]

PR #1246, opened 2024-04-22
 Added initial PNA support into BMv2. Removed built modules (nanomsg and thrift).

rupesh-chiluka-marvell:
Based on the discussions in the issue https://github.com/p4lang/behavioral-model/issues/1245#issuecomment-2063913408, removed nanomsg and thrift directories.


include/bm/bm_sim/nic.h
```diff
@@ -0,0 +1,1104 @@
...
+class Packet;
+
+// multiple inheritance in accordance with Google C++ guidelines:
+// "Multiple inheritance is allowed only when all superclasses, with the
+// possible exception of the first one, are pure interfaces. In order to ensure
+// that they remain pure interfaces, they must end with the Interface suffix."
+//
+//! Base class for a switch implemenattion where multi-context support is
+//! required.
+class NicWContexts : public DevMgr, public RuntimeInterface {
```
antoninbas, line 100: we should really avoid introducing this object as we are just duplicating a lot of code. If some changes are needed to `SwitchWContexts` to accommodate the PNA use case, we should first consider making these changes, instead of introducing `NicWContexts`.

Can you provide some justification for the introduction of this class?

rupesh-chiluka-marvell: Yes, code wise, both the _NicWContexts_ and _SwitchWContexts_ are the same.

Within the _targets/pna_nic_ directory, the _PnaNic_ class is required to inherit from the _Switch_ class. However, considering that a Nic and a Switch are different physical entities. I just felt odd for the _PnaNic_ to inherit directly from the _Switch_. So, I created a duplicate _Nic_ class. But yes, _Nic_ is a duplicate of _Switch_ class and _NicWContexts_ is a duplicate of _SwitchWContexts_ class. 

If the community agrees to eliminate these duplicates and allowing _PnaNic_ class to inherit from the Switch class, we will change accordingly.

Please provide feedback.

antoninbas: A NIC following the PNA architecture is not that different from a switch. The "switch" terminology is not totally out of place here. Duplicating the code would be the worst possible solution here, it creates maintenance issues.

ChrisDodd: Maybe move all/most of the code to a common base `Device` class that `Nic` and `Switch` both inherit from?

jafingerhut: Here are two possible approaches, and what I see as their tradeoffs:

(a) Keep the existing class name `Switch`, and make `Nic` a subclass of it.

Advantages:
+ avoids the code duplication that Antonin commented on.

Disadvantages:
+ It seems like it needs a comment noting that yes, we agree that a class hierarchy where all instances of `Nic` are also instances of `Switch` might not be ideal, it was a reasonable choice given the alternatives.  One might not even think of it as a poor choice of hierarchy if you think of things like Open Vswitch being implemented in NICs: Why is it called V_switch_?  Well, probably because it is switching packets from a set of interfaces, and can end up sending them to 0, 1, or more other interfaces, i.e. it is switching packets between interfaces.

(b) Introduce a new class `Device` that is just a rename of the current `Switch` class, and make a new `Switch` class that is a subclass of it, and a new `Nic` class that is also a subclass of `Device`.

Advantages:
+ completely avoids any mental confusion anyone might have about whether a NIC is a kind of Switch, or not.

Disadvantages:
+ _If_ this requires a lot of little changes throughout the code, one might consider that a disadvantage.  I have not evaluated for certain whether it would require that.

I do not have a strong preference for either (a) or (b), myself, but agree with Antonin's review that approach (c) copy and paste class `Switch` to make a new class `Nic` does seem like significant code duplication.

rupesh-chiluka-marvell: Both options seems reasonable.

option (a): making `Nic` subclass of `Switch`; is an easy approach. But, if we are going to follow option (a), It will be better to follow @antoninbas suggestion and make `PnaNic` class directly inherit `Switch` class.

Following option (b) as @ChrisDodd suggested: Introducing new class `Device`; would need to make few changes in the below mentioned files:
| Source | Number of times `Switch` or `SwitchWContexts` classes are mentioned |
|--------|--------|
| targets/simple_router/simple_router.cpp | 2 |
| targets/l2_switch/l2_switch.cpp | 1 |
| targets/simple_switch_grpc/switch_runner.cpp | 1 |
| targets/psa_switch/psa_switch.cpp | 1 |
| targets/simple_switch/runner.cpp | 1 | 
| include/bm/bm_sim/packet.h | 2 |
| include/bm/bm_sim/context.h | only comments |
| include/bm/bm_sim/lookup_structures.h | 2 |
| include/bm/bm_runtime/bm_runtime.h | 1 |
| include/Makefile.am | 1 |
| PI/src/common.h | 2 |
| PI/src/pi_mc_imp.cpp | 1 |
| PI/src/pi_meter_imp.cpp | 1 |
| PI/src/pi.cpp | 1 |
| src/bm_sim/Makefile.am | 1 |
| src/bm_runtime/SimplePreLAG_server.cpp | 3 |
| src/bm_runtime/server.cpp | 5 |
| src/bm_runtime/Standard_server.cpp | 4 |
| src/bm_runtime/SimplePre_server.cpp | 3 |
| include/bm/bm_sim/switch.h | class changes |
| src/bm_sim/switch.cpp | class changes |
| Need to change documentation files (.md files) | |
| Few more changes in tests | |

But, As @jafingerhut mentioned, following option (b) will avoid any mental confusion between `Switch` and `Nic` in the future.

If the community thinks the above number changes are reasonable and not too many, I will prefer going with option (b). Otherwise, @antoninbas suggestion is a better approach.

rupesh-chiluka-marvell: Or Just name changing _Switch_ class to _Device_ class would be fine. Right?

jafingerhut: I suspect there will be many more significant decisions than this to make in order to enable Nic to work in BMv2 than this decision, and it could easily be 3-4 days between rounds of this conversation.

What do you think of the option of going with Antonin Bas's suggestion for now, and maybe we can revisit later whether changing class name `Switch` to `Device` is preferable?

qobilidop: Agree. I'd prefer that we aim to get a reasonable implementation in ASAP, and iterate on improving it later.

rupesh-chiluka-marvell: Ok. 
I will start by splitting the PR. The first part will include a basic implementation of `pna_nic`. This initial version will serve as a skeleton and will support basic P4 PNA programs (without tables).

KrisNey-MSFT:
hi - does this implement the P4 runtime API by chance?

antoninbas:
> hi - does this implement the P4 runtime API by chance?

No, there was no P4Runtime server added to the target, but I hope someone can do it in the future. There is no reason to have 2 versions of the target like we do for simple_switch (one with P4Runtime and one without).

KrisNey-MSFT:
Would this be a lot of work? 😊

antoninbas:
> Would this be a lot of work?

Hard to tell. I would say not really, but it there may be some plugging required to make sure that P4Runtime / PI calls get turned into the right extern method calls. It's a bit different from simple_switch which has no "externs" (only built-in objects). IIRC, we had a plan on how to do that for psa_switch a while back, and it's not that hard.

KrisNey-MSFT:
Yes, we definitely need P4 runtime :)

PR #1252, opened 2024-05-28
Upgrade to C++17

OsamaRab3:


antoninbas:
You should test your changes locally before you push. You should at least make sure that things build.
Right now I don't believe this can build, because `AX_CXX_COMPILE_STDCXX_17` is not defined anywhere, unless you forgot to check in one file.

OsamaRab3:
> You should test your changes locally before you push. You should at least make sure that things build. 
Subject: Issues Encountered During Project Testing

I have encountered some issues during the testing phase. Specifically, there are a few tests that are failing, and I am unsure about the cause. Here are the details:
1. **QueueingPriRLTest.Pri**
   - **Location:** `test_queueing.cpp:302`
   - **Failure Reason:** Expected `(0.2) > ((priority_1 - c) / static_cast<double>(priority_0))` but got `0.2` vs `0.297959`.

2. **QueueingPriRLTest.PriRateLimiter**
   - **Location:** `test_queueing.cpp:326`
   - **Failure Reason:** Expected `(diff) < (std::max(priority_0, priority_1) * 0.2)` but got `156` vs `103.4`.

3. **MetersTest.trTCM**
   - **Location:** `test_meters.cpp:159`
   - **Failure Reason:** Expected equality of these values:
     - Expected: `{ 0, 2, 0, 0, 1, 1, 0, 1, 1, 1, 2, 2, 0 }`
     - Output: `{ 0, 0, 2, 0, 1, 1, 0, 1, 1, 1, 2, 2, 0 }`

4. **PeriodicExtern.MultipleTasks**
   - **Location:** `test_periodic_tasks.cpp:106`
   - **Failure Reason:** Expected `(j) >= ((kPeriods * 2) - 1)` but got `8` vs `9`.


I have reviewed the test code and the implementation, but I am still not sure how to resolve these issues. Could you please provide some guidance or insights on how to approach these failures?



image

antoninbas:
They are several references to C++ 11 (found with `git grep "c++11"`) which I think should be looked at

antoninbas:
@OsamaRab3 A blind `sed` to replace all instances of `c++11` with `c++17` is not what I meant

OsamaRab3:
> @OsamaRab3 A blind `sed` to replace all instances of `c++11` with `c++17` is not what I meant

I apologize for the misunderstanding. I thought that updating to C++17 required changes in the documentation, so I made those changes. I now realize that a blind `sed` replacement was not the right approach. 

Thank you for pointing that out, and I'll make sure to follow the correct process moving forward.

jafingerhut:
@OsamaRab3 A couple of days ago I saw a message where you had some failing tests.  Were you able to get those working later, and deleted that comment for that reason, perhaps?

OsamaRab3:
> @OsamaRab3 A couple of days ago I saw a message where you had some failing tests. Were you able to get those working later, and deleted that comment for that reason, perhaps?

Yes,I resolved the issues with the tests and deleted the comment because they are now working correctly.

m4/ax_cxx_compile_stdcxx_17.m4
antoninbas, line 1: This looks like the same file as before, where you would have replaced "11" with "17".
This is not right. Please use the latest available official version of `AX_CXX_COMPILE_STDCXX_17`

I think you will need:
* https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_17.html
* https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html

tools/macos/brews/thrift.rb
```diff
@@ -52,7 +52,7 @@ def install
     exclusions << "--without-python" if build.without? "python"
     exclusions << "--without-perl" if build.without? "perl"
 
-    ENV.cxx11 if MacOS.version >= :mavericks && ENV.compiler == :clang
+    ENV.cxx17 if MacOS.version >= :mavericks && ENV.compiler == :clang
```
antoninbas, line 55: do you know what this change implies? because I don't

antoninbas (commented):
I don't think my earlier comment has been addressed TBH

PR #1263, opened 2024-08-07
Added Basic PNA externs: Meter, Counter, Hash, etc 

rupesh-chiluka-marvell:
Part of Issue: https://github.com/p4lang/behavioral-model/issues/1245

### DIFF
--- a/targets/pna_nic/pnic_CLI.py
+++ b/targets/pna_nic/pnic_CLI.py
@@ -48,28 +48,7 @@ class PnaNicAPI(runtime_CLI.RuntimeAPI):
         print(self.pnic_client.get_time_since_epoch_us())
 
 def load_json_pna(json):
-    def get_json_key(key):
-        return json.get(key, [])
-
-    for j_meter in get_json_key("extern_instances"):
-        if j_meter["type"] != "Meter":
-            continue
-        meter_array = runtime_CLI.MeterArray(j_meter["name"], j_meter["id"])
-        attribute_values = j_meter.get("attribute_values", [])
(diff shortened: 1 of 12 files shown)


rupesh-chiluka-marvell:
@antoninbas @qobilidop. Please review the PR

targets/pna_nic/pnic_CLI.py
```diff
@@ -48,7 +48,28 @@ def do_get_time_since_epoch(self, line):
         print(self.pnic_client.get_time_since_epoch_us())
 
 def load_json_pna(json):
-    pass
+    def get_json_key(key):
+        return json.get(key, [])
+
+    for j_meter in get_json_key("extern_instances"):
```
antoninbas, line 54: I would remove the changes to this file from this PR. I am not sure why we are only adding support for meters, but if that's what we need, I would still do it in a follow-up PR.

antoninbas (commented):
Is there any extern in this PR that is specific to PNA or different from its PSA equivalent? I am asking so I know where I should focus my review efforts.

targets/pna_nic/pnic_CLI.py
rupesh-chiluka-marvell: In this PR, No.

rupesh-chiluka-marvell:
> Is there any extern in this PR that is specific to PNA or different from its PSA equivalent? I am asking so I know where I should focus my review efforts.

In this PR. No. These are general externs used in P4 programs.

antoninbas (approved):

PR #1273, opened 2024-10-30
Added Recirculation functionality in PNA_NIC

ghost:
Part of https://github.com/p4lang/behavioral-model/issues/1245 issue.

Need to add below declaration in _p4c/p4include/bmv2/pna.p4_:
```
const PortId_t PNA_PORT_RECIRCULATE = (PortId_t) 0xfffffffa;
```

### DIFF
--- a/targets/pna_nic/pna_nic.cpp
+++ b/targets/pna_nic/pna_nic.cpp
@@ -204,6 +204,24 @@ PnaNic::main_thread() {
 
     Deparser *deparser = this->get_deparser("main_deparser");
     deparser->deparse(packet.get());
+
+    // recirculation
+    if ((port_t) packet->get_egress_port() == PNA_PORT_RECIRCULATE) {
+      BMLOG_DEBUG_PKT(*packet, "Recirculating packet");
+
+      phv->reset();
+      phv->reset_header_stacks();
+      phv->reset_metadata();
+
+      phv->get_field("pna_main_parser_input_metadata.input_port")
+        .set(input_port);
+      phv->get_field("pna_main_parser_input_metadata.recirculated")
+        .set(1);
+      
+      input_buffer.push_front(std::move(packet));
+      continue;
+    }
+    
     output_buffer.push_front(std::move(packet));
   }
 }
--- a/targets/pna_nic/pna_nic.h
+++ b/targets/pna_nic/pna_nic.h
@@ -79,6 +79,7 @@ class PnaNic : public Switch {
 
  private:
   static packet_id_t packet_id;
+  static constexpr port_t PNA_PORT_RECIRCULATE = 0xfffffffa;
 
   enum PktInstanceType {
     FROM_NET_PORT,


jafingerhut (approved):
All of this looks correct as far as I can tell.  Future new P4 program test cases that exercise this new code with p4c compiling the program, then packets passing through pna_nic in the expected ways, are the best assurance of that.

rupesh-chiluka-marvell:
> All of this looks correct as far as I can tell. Future new P4 program test cases that exercise this new code with p4c compiling the program, then packets passing through pna_nic in the expected ways, are the best assurance of that.

Yes, Will include a test case

targets/pna_nic/pna_nic.h
```diff
@@ -79,6 +79,7 @@ class PnaNic : public Switch {
 
  private:
   static packet_id_t packet_id;
+  static constexpr port_t PNA_PORT_RECIRCULATE = 0xfffffffa;
```
antoninbas, line 82: There is no mention of a recirculation port in the PNA spec. @jafingerhut is this correct?
All I know is that this is the SDN port value defined by the P4Runtime spec as the recirculation port.
I guess targets are free to "choose" (or define) the value used for the recirculation port.

targets/pna_nic/pna_nic.cpp
```diff
@@ -204,6 +204,21 @@ PnaNic::main_thread() {
...
+    if ((port_t) packet->get_egress_port() == PNA_PORT_RECIRCULATE) {
+      BMLOG_DEBUG_PKT(*packet, "Recirculating packet");
+
+      phv->reset();
+      phv->reset_header_stacks();
+      phv->reset_metadata();
+
+      phv->get_field("pna_main_parser_input_metadata.recirculated")
+        .set(1);
+      input_buffer.push_front(std::move(packet));
```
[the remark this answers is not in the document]
jafingerhut: I believe it should be unchanged from the original.  Unfortunately some things like this are not yet spelled out in as precise a form in the PNA architecture as they are in the PSA architecture, but I believe "input_port is unchanged after recirculate operation" is a reasonable starting place (and maybe finishing place, too).

targets/pna_nic/pna_nic.h
jafingerhut: Yes, there is a recirculate() operation in PNA, and if that is done, then the packet should recirculate when it completes the current execution of the main control.  There is NO recirculation port like there is in PSA or v1model architectures.

jafingerhut: Note: As an implementation technique, it is probably a reasonable idea to have a C++ boolean or 1-bit flag somewhere associated with the packet in the PNA implementation that is initialized to 0/false whenever a packet begins execution of the main control, and it is set to 1/true when the recirculate() extern function is called.

antoninbas: I couldn't find a `recirculate()` extern function in [pna.p4](https://github.com/p4lang/pna/blob/main/pna.p4)? Is there a reference to it somehere.
If PNA specifies such an extern function, then this is what this PR should introduce (not a special port value). Implementation of the extern function can use a special port value if appropriate (the boolean flag approach you suggested may be a bit more difficult within the constraints of bmv2).

targets/pna_nic/pna_nic.cpp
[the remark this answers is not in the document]
antoninbas: Thanks. This is not what the current PR does, but it can be done easily. See PSA implementation: https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/targets/psa_switch/psa_switch.cpp#L618-L619

targets/pna_nic/pna_nic.h
rupesh-chiluka-marvell: Yes. I didn't find any `recirculate()` extern function in [pna.p4](https://github.com/p4lang/pna/blob/main/pna.p4). So, I followed the PSA implementation.

If introducing a new `recirculate()` function is better approach, we can implement that.

targets/pna_nic/pna_nic.cpp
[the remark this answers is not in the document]
rupesh-chiluka-marvell: Ok. Will update the line.

Just a doubt: Why are we setting `ingress_port` field to `PSA_PORT_RECIRCULATE` ? The P4 programs will find whether the packet is recirculated on not using `recirculated_flag` field.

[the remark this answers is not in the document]
antoninbas: According to @jafingerhut, the port should be set to the original ingress port, not `PSA_PORT_RECIRCULATE`
But we should still set the port to something. With your current implementation, the port will always be set to 0. At least that's what will happen I think based on the `reset_metadata` call (I didn't test).

targets/pna_nic/pna_nic.h
rupesh-chiluka-marvell: Hi @antoninbas, Is the current approach Ok? Or should I change it recirculate() extern approach?

targets/pna_nic/pna_nic.cpp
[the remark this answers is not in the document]
jafingerhut: One approach is to wait until there is an official PNA definition for how recirculation is done, and behaves, and implement other features that are already well-defined.

targets/pna_nic/pna_nic.h
jafingerhut: One approach is to wait until there is an official PNA definition for how recirculation is done, and behaves, and implement other features that are already well-defined.

rupesh-chiluka-marvell: Ok

PR #1276, opened 2024-11-07
fix #1264 putting packets into the right priority queue

### LINKED ISSUE #1264: set queue rate/depth commands with priority queues
Hi, I am working on simple_switch target with priority queues.

I am generating traffic to the queues, and trying to change the capacity and rate of each queue, to see the impact of it in the QoS and in the occupation of each one. But, in my tests, it seems that the _set queue rate/depth_ commands are changing the values of the inverse queue of the input!

For example, when I have 3 priority queues, and I generate traffic (UDP) to queue/priority 2, and I want to reduce the rate of that queue, I use the command _set_queue_rate 100 0 2_ (on simple_switch_CLI), I can't see any difference in the QoS or in the queue occupation. But if I use the _set_queue_rate 100 0 0_, then the QoS and occupation of queue 2 change! The same thing happens when I change the depth/capacity of the queues.

Then I repeated the test with 4 priority queues, and in this case I had to "change queue 1" to see the difference in queue 2 occupation.

In the past I did a little change in _simple_switch.cpp_ code, to return the occupation of the specific priority queue of each packet pass in the enq_qdepth metadata, but I think it is not the reason of the problem.

Can anybody repeat these tests to confirm?

rizkyramadhana26:
As discussed in #1264 and p4lang/project-ideas#6 , the priority in ```set_queue_rate``` and ```set_queue_depth``` is not consistent with the real priority queue in which a packet is located. I propose a change to put a packet in queue with qid ```priority``` instead of ```nb_queues_per_port - 1 - priority```. This way, ```set_queue_rate``` and ```set_queue_depth``` will affect the right queue and queue with priority 0 has the highest priority.

I reproduce this behavior in a separate repository https://github.com/rizkyramadhana26/reproduce-bmv2-issue-1264

### DIFF
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -469,10 +469,24 @@ TODO: reset_state
 TODO: serialize_state
 TODO: set_crc16_parameters
 TODO: set_crc32_parameters
-TODO: set_queue_depth [simple_switch_CLI only]
-TODO: set_queue_rate [simple_switch_CLI only]
 TODO: shell
 ```
+
+### set_queue_depth [simple_switch_CLI only]
+```
+RuntimeCmd: help set_queue_depth
+Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port> [<priority>]]
+```
+
+This command will set queue's maximum depth for a priority queue that is identified by egress_port and priority. If priority is not given, this command will set queue's maximum depth for all queues in the given egress_port. If both priority and egress_port are not given, this command will set queue's maximum depth for all queues. The priority mentioned here is the same as qid in the queueing_metadata.
+
+### set_queue_rate [simple_switch_CLI only]
+```
+RuntimeCmd: help set_queue_rate
+Set rate of one / all egress queue(s): set_queue_rate <rate_pps> [<egress_port> [<priority>]]
+```
+
+This command will set how many packets are processed for each second in every queues that are identified by egress_port and priority. If priority is not given, this command will set queue's rate for all queues in the given egress_port. If both priority and egress_port are not given, this command will set queue's rate for all queues. The priority mentioned here is the same as qid in the queueing_metadata.
 
 ### show_actions
 
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -478,7 +478,12 @@ RuntimeCmd: help set_queue_depth
 Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port> [<priority>]]
 ```
 
-This command sets queue's maximum depth to `nb_pkts` in units of number of packets. The affected priority queue is identified by `egress_port` and `priority`. If `priority` is not given, this command sets queue's maximum depth for all queues in the given `egress_port`. If both `priority` and `egress_port` are not given, this command sets queue's maximum depth for all queues. The `priority` mentioned here is the same as qid in the queueing_metadata.
+This command sets the maximum depth of queue(s) to `nb_pkts` in units of 
+number of packets. The affected priority queue is identified by `egress_port` 
+and `priority`. If `priority` is not given, this command sets the maximum depth 
+for all queues in the given `egress_port`. If both `priority` and 
+`egress_port` are not given, this command sets the maximum depth for all queues. 
+The `priority` mentioned here is the same as qid in the queueing_metadata.
 
 ### set_queue_rate [simple_switch_CLI only]
 ```
@@ -486,7 +491,12 @@ RuntimeCmd: help set_queue_rate
 Set rate of one / all egress queue(s): set_queue_rate <rate_pps> [<egress_port> [<priority>]]
 ```
 
-This command sets the rate at which a queue is consumed to `rate_pps`, in units of packet per second. The affected priority queue is identified by `egress_port` and `priority`. If `priority` is not given, this command sets queue's rate for all queues in the given `egress_port`. If both `priority` and `egress_port` are not given, this command sets queue's rate for all queues. The `priority` mentioned here is the same as qid in the queueing_metadata.
+This command sets the rate at which queue(s) is(are) consumed to `rate_pps`, 
+in units of packet per second. The affected priority queue is identified by 
+`egress_port` and `priority`. If `priority` is not given, this command sets 
+the rate for all queues in the given `egress_port`. If both `priority` and 
+`egress_port` are not given, this command sets the rate for all queues. 
+The `priority` mentioned here is the same as qid in the queueing_metadata.
 
 ### show_actions
 
--- a/docs/runtime_CLI.md
+++ b/docs/runtime_CLI.md
@@ -478,7 +478,7 @@ RuntimeCmd: help set_queue_depth
 Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port> [<priority>]]
 ```
 
-This command will set queue's maximum depth for a priority queue that is identified by egress_port and priority. If priority is not given, this command will set queue's maximum depth for all queues in the given egress_port. If both priority and egress_port are not given, this command will set queue's maximum depth for all queues. The priority mentioned here is the same as qid in the queueing_metadata.
+This command sets queue's maximum depth to `nb_pkts` in units of number of packets. The affected priority queue is identified by `egress_port` and `priority`. If `priority` is not given, this command sets queue's maximum depth for all queues in the given `egress_port`. If both `priority` and `egress_port` are not given, this command sets queue's maximum depth for all queues. The `priority` mentioned here is the same as qid in the queueing_metadata.
 
 ### set_queue_rate [simple_switch_CLI only]
 ```
@@ -486,7 +486,7 @@ RuntimeCmd: help set_queue_rate
 Set rate of one / all egress queue(s): set_queue_rate <rate_pps> [<egress_port> [<priority>]]
 ```
 
-This command will set how many packets are processed for each second in every queues that are identified by egress_port and priority. If priority is not given, this command will set queue's rate for all queues in the given egress_port. If both priority and egress_port are not given, this command will set queue's rate for all queues. The priority mentioned here is the same as qid in the queueing_metadata.
+This command sets the rate at which a queue is consumed to `rate_pps`, in units of packet per second. The affected priority queue is identified by `egress_port` and `priority`. If `priority` is not given, this command sets queue's rate for all queues in the given `egress_port`. If both `priority` and `egress_port` are not given, this command sets queue's rate for all queues. The `priority` mentioned here is the same as qid in the queueing_metadata.
 
 ### show_actions
 
--- a/docs/simple_switch.md
+++ b/docs/simple_switch.md
@@ -223,7 +223,12 @@ queue.
 - `deq_qdepth`: the depth of queue when the packet was dequeued, in units of number of packets (not the total size of packets).
 - `qid`: when there are multiple queues servicing each egress port (e.g. when
 priority queueing is enabled), each queue is assigned a fixed unique id, which
-is written to this field. Otherwise, this field is set to 0. If priority queueing is enabled, the qid also describe the priority level of each queue. Starting with 0 that has the lowest priority until `number_of_priority_queues - 1` that has the highest priority. The number of priority queues for each port can be defined by adding --priority-queues when running ```simple_switch```.
+is written to this field. Otherwise, this field is set to 0. If 
+priority queueing is enabled, the qid also describes the priority level 
+of each queue. Starting with 0 that has the lowest priority until 
+`number_of_priority_queues - 1` that has the highest priority. The 
+number of priority queues for each port can be defined by adding 
+`--priority-queues` when running `simple_switch`.
 
 TBD: `qid` is not currently part of type `standard_metadata_t` in v1model.
 Perhaps it should be added?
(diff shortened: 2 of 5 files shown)


docs/runtime_CLI.md
```diff
@@ -469,11 +469,25 @@ TODO: reset_state
...
 TODO: shell
 ```
 
+### set_queue_depth [simple_switch_CLI only]
+```
+RuntimeCmd: help set_queue_depth
+Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port> [<priority>]]
+```
+
+This command will set queue's maximum depth for a priority queue that is identified by egress_port and priority. If priority is not given, this command will set queue's maximum depth for all queues in the given egress_port. If both priority and egress_port are not given, this command will set queue's maximum depth for all queues. The priority mentioned here is the same as qid in the queueing_metadata.
```
jafingerhut, line 481: Please be a little bit more explicit about the units of <nb_pkts> in the text, e.g. something like "will set the queue's maximum depth, in units of number of packets, for a priority queue ..."

docs/runtime_CLI.md
```diff
@@ -469,11 +469,25 @@ TODO: reset_state
...
+
+### set_queue_rate [simple_switch_CLI only]
+```
+RuntimeCmd: help set_queue_rate
+Set rate of one / all egress queue(s): set_queue_rate <rate_pps> [<egress_port> [<priority>]]
+```
+
+This command will set how many packets are processed for each second in every queues that are identified by egress_port and priority. If priority is not given, this command will set queue's rate for all queues in the given egress_port. If both priority and egress_port are not given, this command will set queue's rate for all queues. The priority mentioned here is the same as qid in the queueing_metadata.
```
jafingerhut, line 489: Again, please be a little bit more explicit in the text about the units of the setting.

docs/runtime_CLI.md
rizkyramadhana26: I add more description in the following commit, thanks for the input!

docs/runtime_CLI.md
rizkyramadhana26: I also add more description about the unit in the following commit, thanks for the input!

antoninbas (changes requested):
As pointed out by @jafingerhut in https://github.com/p4lang/behavioral-model/issues/1264#issuecomment-2455168459, it is not clear that this is a superior approach. Therefore< I don't see the point of breaking backwards compatibility.

I am happy to approve an addition to the documentation though.

jafingerhut:
> As pointed out by @jafingerhut in [#1264 (comment)](https://github.com/p4lang/behavioral-model/issues/1264#issuecomment-2455168459), it is not clear that this is a superior approach. Therefore< I don't see the point of breaking backwards compatibility.
> 
> I am happy to approve an addition to the documentation though.

Antonin, are you saying that you have a preference for the implementation choice in which the priority value Q-1 is the highest priority queue number, and 0 is the lowest priority queue number?

I have no problem with that choice, as long as it is consistently implemented in both the use of the `priority` metadata field in P4 source code, and the simple_switch_CLI commands for configuring the queues, and the documentation.

rizkyramadhana26:
Yes, I also don't mind if queue Q-1 has the highest priority. Currently, the problem is:
1. Let say there are 4 queue for every egress port, with qid 0,1,2,3
2. If a user sends a packet to queue with priority 3, it goes to qid 0 (4-1-3)
3. However, if a user set the rate/depth for queue with priority 3, it affects qid 3
4. This might creates confusion, as the same priority refers to different queues in these actions.

jafingerhut:
> Yes, I also don't mind if queue Q-1 has the highest priority. Currently, the problem is:
> 
> 1. Let say there are 4 queue for every egress port, with qid 0,1,2,3
> 2. If a user sends a packet to queue with priority 3, it goes to qid 0 (4-1-3)
> 3. However, if a user set the rate/depth for queue with priority 3, it affects qid 3
> 4. This might creates confusion, as the same priority refers to different queues in these actions.

There are code changes that can be made so that when a user sets the rate/depth for queue with priority 3, the code calculates Q-1-3 to get 0, and makes the configuration change on queue 0 instead.  Just like the calculation used currently after ingress processing is done, that calculates the queue number from the `priority` metadata field.

jafingerhut:
Yet another possibility, perhaps a simpler one, if the desire is to end up with a system where Q-1 is strict highest priority and 0 is the lowest priority, is to modify the loop here:

+ https://github.com/p4lang/behavioral-model/blob/main/include/bm/bm_sim/queueing.h#L581

so that it iterates through the queues in the opposite order from what it does now.

jafingerhut:
> So I think we should keep the current behavior where setting the priority for 0 for a given packet means that packet will be processed with the highest priority. If running CLI command `set_queue_rate <rate_pps> <egress_port> 0` does not increase the rate for the correct queue (i.e., the queue where my example packet is enqueued), then I agree it needs to be fixed. It should be fixed here
> 
> https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/targets/simple_switch/simple_switch.cpp#L333
> 
> 
> and there
> https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/targets/simple_switch/simple_switch.cpp#L352

I believe that is what this PR will do, i.e. packet metadata `priority` equal to 0 will put the packet into queue 0 in the data structure, which is currently being scheduled with highest priority, according to the code analysis and experiments that have been done.  (I have not done those experiments myself.)

antoninbas:
> I believe that is what this PR will do, i.e. packet metadata priority equal to 0 will put the packet into queue 0 in the data structure, which is currently being scheduled with highest priority, according to the code analysis and experiments that have been done. (I have not done those experiments myself.)

I think I described it poorly and inaccurately.
The datapath behavior should stay the same, and it seems that according to that behavior, setting the priority to 0 makes the packet low priority:
https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/targets/simple_switch/simple_switch.h#L42

The runtime API implementation should be fixed so that it is consistent with the datapath behavior.

jafingerhut:
I haven't run an experiment with multiple simultaneously non-empty priority queues for one output port to confirm, but I do not see how that comment that queue 0 is the lowest priority can be consistent with this loop, which iterates through queues from 0 to increasing queue numbers, and picks the first non-empty one that isn't being stopped by its rate limit configuration:

+ https://github.com/p4lang/behavioral-model/blob/main/include/bm/bm_sim/queueing.h#L581

antoninbas:
The priority value is inverted in simple_switch when enqueueing:

https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/targets/simple_switch/simple_switch.cpp#L422-L424

So assuming 8 queues per port:
- P4 program sets priority to 0 (lowest priority) in ingress
- simple_switch enqueues packet in queue 7 (lowest priority queue, served last)
- `queueing_metadata.qid` is set to 0 (we invert a second time IIRC) for egress processing

The problem is that the CLI commands will not invert. So using `priority=0` in a command will actually configure queue 0, and not queue 7.

jafingerhut:
So note that one way to get the behavior you want, Antonin, that I think is less confusing for reading and understanding the code, is to:

(a) change the `for` loop here https://github.com/p4lang/behavioral-model/blob/main/include/bm/bm_sim/queueing.h#L581 so that it iterates through the queues from the maximum queue number down to queue number 0, and picking the first one that is non-empty and not stopped by its rate limit.

(b) Change the logic here https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/targets/simple_switch/simple_switch.cpp#L423 so that the formula `nb_queues_per_port - 1 - priority` is replaced with `priority`.

(c) leave the simple_switch_CLI queue configuration commands unchanged.

Then consistently everywhere in the C++ code implementation we would be using the convention that queue Q-1 is the highest priority, queue 0 is the lowest priority, and no remapping or formulas are used anywhere to "invert" them.

What do you think?

antoninbas:
@jafingerhut that is fine by me as long as the code documentation / comments in queueing.h are also updated correctly

jafingerhut:
@rizkyramadhana26 Do my suggested changes in this comment make sense to you?  https://github.com/p4lang/behavioral-model/pull/1276#issuecomment-2463309248

Basically I think from the PR you have right now, the changes required would be:

(a) reverse the order that queues are iterated over in https://github.com/p4lang/behavioral-model/blob/main/include/bm/bm_sim/queueing.h#L581 so that instead of 0 up to Q-1, it goes Q-1 down to 0.

(b) change comments above that function to say that higher queue numbers have higher priority.

(c) Change the documentation you have added to be consistent with this.

rizkyramadhana26:
I have added the suggested change. In short:

1. packets that are sent to priority 8 (highest priority) are enqueued in qid 8
2. set_queue_rate and set_queue_depth to priority 8 also affect qid 8
3. queues are being dequeued from highest priority values (nb_priorities-1) to 0

I also propose some changes to comments in queueing.h. Thanks for the input!

rizkyramadhana26:
I am checking the failed test

rizkyramadhana26:
https://github.com/p4lang/behavioral-model/blob/28b736caabca7a4c9ff23e3b947354bd55191372/tests/test_queueing.cpp#L295-L302

Should we also change this unit test? Since queue 1 has the highest priority now, I believe that most of the element should be P1 (instead of P0). In short, this test case will be
1. We have to receive at least capacity P0 element
2. Most elements should be P1

include/bm/bm_sim/queueing.h
```diff
@@ -578,7 +578,7 @@ class QueueingLogicPriRL {
       } else {
         auto now = clock::now();
         auto next = clock::time_point::max();
-        for (pri = 0; pri < nb_priorities; pri++) {
+        for (pri = nb_priorities - 1; 0 <= pri; pri--) {
```
jafingerhut, line 581: If `size_t` is an unsigned type, then note that `0 <= pri` is _always_ true, and this becomes an infinite loop unless something prevents that in the loop body.  It might be best to change the declaration of `pri` to `int pri;` ?

jafingerhut: According to a warning during compilation in one of the CI tests, `0 <= pri` is always true, so `size_t` is definitely an unsigned type and that declaration of `pri` should be changed to a signed type.

antoninbas: this looks cool:
```cpp
for (pri = nb_priorities; pri --> 0 ;)
```
source: https://stackoverflow.com/a/3626468

it will avoid writing a cast...

antoninbas (commented):
I would expect that with this approach, some unit tests will need to be updated https://github.com/p4lang/behavioral-model/blob/main/tests/test_queueing.cpp

rizkyramadhana26:
I didn't realize the infinite loop, thank you! I also propose a change to the unit test and all tests are passed now.

jafingerhut (approved):
Looks good to me.

docs/runtime_CLI.md
```diff
@@ -469,11 +469,25 @@ TODO: reset_state
...
 TODO: shell
 ```
 
+### set_queue_depth [simple_switch_CLI only]
+```
+RuntimeCmd: help set_queue_depth
+Set depth of one / all egress queue(s): set_queue_depth <nb_pkts> [<egress_port> [<priority>]]
+```
+
+This command sets queue's maximum depth to `nb_pkts` in units of number of packets. The affected priority queue is identified by `egress_port` and `priority`. If `priority` is not given, this command sets queue's maximum depth for all queues in the given `egress_port`. If both `priority` and `egress_port` are not given, this command sets queue's maximum depth for all queues. The `priority` mentioned here is the same as qid in the queueing_metadata.
```
antoninbas, line 481: s/queue/the queue

(applies to multiple locations)

docs/simple_switch.md
```diff
@@ -223,7 +223,8 @@ queue.
 - `deq_qdepth`: the depth of queue when the packet was dequeued, in units of number of packets (not the total size of packets).
 - `qid`: when there are multiple queues servicing each egress port (e.g. when
 priority queueing is enabled), each queue is assigned a fixed unique id, which
-is written to this field. Otherwise, this field is set to 0.
+is written to this field. Otherwise, this field is set to 0. If priority queueing is enabled, the qid also describe the priority level of each queue. Starting with 0 that has the lowest priority until `number_of_priority_queues - 1` that has the highest priority. The number of priority queues for each port can be defined by adding --priority-queues when running ```simple_switch```.
```
antoninbas, line 226: s/the qid also describe/the qid also describes

antoninbas, line 226: * please wrap the line at a reasonable character limit (~100)

 * use backticks for the `--priority-queues` flag

* why use triple backticks for `simple_switch`, but a single one for every other inline code snippet?

antoninbas (commented):
It would be good if `simple_switch.md` could be updated in a future PR to mention `intrinsic_metadata.priority`, as this is how the P4 program chooses the priority queue and as of now there is no mention of it.

rizkyramadhana26:
I made some changes to the docs, thank you for the input! Should we also add description about `intrinsic_metadata.priority` in this PR? Or we can leave it for future PR?

antoninbas:
> I made some changes to the docs, thank you for the input! Should we also add description about `intrinsic_metadata.priority` in this PR? Or we can leave it for future PR?

Yes, a follow up PR would be great

PR #1277, opened 2024-11-08
Update docs for if support in action bodies

jafingerhut:


jafingerhut:
Closing this PR, as it seems to include some changes that revert recently made improvements in the docs.  I will try creating a different PR without this problem.

PR #1279, opened 2024-11-10
Always call packet->reset_exit() at beginning of Pipeline::apply()

jafingerhut:
The only place the code checks for is_marked_for_exit() is inside of Pipeline::apply(), so by resetting exit flag at beginning of Pipeline::apply, no users of Pipeline class need to worry about the exit flag at all.  It is handled completely locally within Pipeline::apply(), and could even be a local variable of Pipeline::appply() if the primitive made that straightforward (which it does not).

### DIFF
--- a/src/bm_sim/pipeline.cpp
+++ b/src/bm_sim/pipeline.cpp
@@ -28,6 +28,7 @@ namespace bm {
 
 void
 Pipeline::apply(Packet *pkt) {
+  pkt->reset_exit();
   BMELOG(pipeline_start, *pkt, *this);
   // TODO(antonin)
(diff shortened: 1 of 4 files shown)


antoninbas:
Only concern for me is that the last part of the `mark_for_exit` comment is no longer valid: https://github.com/p4lang/behavioral-model/blob/c0ff3aba77a78ed08dfcb8634f0875fa172e18ab/include/bm/bm_sim/packet.h#L256-L259

Now I have no idea why I thought that was the desired behavior. But if someone calls `exit` in a parser, the new behavior doesn't match the old one.

PR #1282, opened 2024-12-24
Adding PNA IPSec Accelerator extern

rupesh-chiluka-marvell:
Part of https://github.com/p4lang/behavioral-model/issues/1245 Issue.


jafingerhut:
A couple of CI tests are failing with your PR's changes.  Have you investigated the cause?

rupesh-chiluka-marvell:
> A couple of CI tests are failing with your PR's changes. Have you investigated the cause?

Yes.
I think the dependency should be changed OpenSSL 1.1 to OpenSSL 3.

rupesh-chiluka-marvell:
> A couple of CI tests are failing with your PR's changes. Have you investigated the cause?

The docker environments in the failed CI tests use OpenSSL version 1.1.
However, since the 1.1 version is deprecated, I have used OpenSSL 3 functionalities for the above changes.

Can we add dependency of the latest OpenSSL version similarly to `ci/install-nanomsg.sh`

jafingerhut:
> > A couple of CI tests are failing with your PR's changes. Have you investigated the cause?
> 
> The docker environments in the failed CI tests use OpenSSL version 1.1. However, since the 1.1 version is deprecated, I have used OpenSSL 3 functionalities for the above changes.
> 
> Can we add dependency of the latest OpenSSL version similarly to `ci/install-nanomsg.sh`

As a test on this PR, you can modify whatever files you want in this PR that install the OpenSSL version you want, yes?  Please try that, and see if the tests pass in CI.

Then we can decide whether we want some kind of option or environment variable or something that selects the version of OpenSSL to be installed by the scripts, or perhaps we just have it be the newer version of OpenSSL always, if with that newer version everything else works when that updated behavioral-model code is used to run tests with p4c.

antoninbas:
The fact that the base Docker image is still Ubuntu 20.04 is going to have to be addressed eventually. Maybe it should be done sooner rather than later. I updated all dependencies (including Ubuntu, from 16.04 to 20.04) a couple of years ago: https://github.com/p4lang/third-party/pull/30. It wasn't that big of a change, though I am sure it took some trial and error, and a few cycles of trying to build the image locally. Maybe someone could take a stab at it now and update all dependencies again, including Ubuntu to 24.04.

jafingerhut:
> The fact that the base Docker image is still Ubuntu 20.04 is going to have to be addressed eventually. Maybe it should be done sooner rather than later. I updated all dependencies (including Ubuntu, from 16.04 to 20.04) a couple of years ago: [p4lang/third-party#30](https://github.com/p4lang/third-party/pull/30). It wasn't that big of a change, though I am sure it took some trial and error, and a few cycles of trying to build the image locally. Maybe someone could take a stab at it now and update all dependencies again, including Ubuntu to 24.04.

Antonin, for my own clarification: I believe that since this is an Ubuntu 20.04 Docker image, with Ubuntu 24.04 as the base OS that the container is running on, it should not suddenly start failing on April 1, 2025 when Github stops supporting Ubuntu 20.04 as the base OS for CI tests.

Agreed that the OS running inside the container should be updated, probably to 24.04, soon.  Just not necessarily with the April 1 deadline.

jafingerhut:
I created this issue to track the task that Antonin describes, copying his description of the work he did in the past to update to Ubuntu 20.04: https://github.com/p4lang/behavioral-model/issues/1296

rupesh-chiluka-marvell:
After updating the base image to Ubuntu 24.04, the CI tests for this PR will pass. 
However, if this PR gets merged, users on Ubuntu versions before 24 without OpenSSL 3 will encounter compilation errors and must install OpenSSL 3 separately.

jafingerhut:
> After updating the base image to Ubuntu 24.04, the CI tests for this PR will pass. However, if this PR gets merged, users on Ubuntu versions before 24 without OpenSSL 3 will encounter compilation errors and must install OpenSSL 3 separately.

As long as we test it going back to Ubuntu 22.04 as also working, and Ubuntu 20.04 if you really want to (free long term support ends in April 2025 soon), I don't see a problem with requiring OpensSSL 3.

PR #1283, opened 2024-12-24
Print the multicast group id along with the number of packets replicated.

angelazhang8:


jafingerhut (approved):
LGTM.  @smolkaj You might want to check this to see if the updated BMv2 log output would require adjustments to anything you do that check the logs, and/or pass this PR on to others to double-check.

jonathan-dilorenzo:
Should be fine for us. We're trying to use this extra information to extract information from the logs more effectively. Would appreciate if you could merge it when you get a chance @jafingerhut, since I don't have the relevant authorizations. But approved on my and @smolkaj's (who's out until Feb) behalf.

jonathan-dilorenzo (approved):
Looks great! Thanks Angela!

PR #1287, opened 2025-01-13
Add extra checks on next tables

jafingerhut:


jafingerhut:
Add a bit more error checking on the contents of BMv2 JSON file contents.  I have successfully run all p4c tests with these changes to bmv2, so I know that none of those hundreds of test programs violate these additional checks that cause exceptions to be thrown.

include/bm/bm_sim/P4Objects.h
```diff
@@ -59,6 +59,8 @@ class Value;
 
 namespace bm {
 
+using std::string;
```
antoninbas, line 62: I'd personally recommend against doing that, especially in a header file, even if namespaced

include/bm/bm_sim/P4Objects.h
```diff
@@ -378,6 +380,10 @@ class P4Objects {
   void init_meter_arrays(const Json::Value &root, InitState *);
   void init_register_arrays(const Json::Value &root);
   void init_actions(const Json::Value &root);
+  void check_next_nodes(const Json::Value &cfg_next_nodes,
+                        const Json::Value &cfg_actions,
+                        const string table_name,
```
antoninbas, line 385: no reason to pass the string by value here I think, so ideally this would be `const std::string &table_name`

it would be more consistent with other methods as well

include/bm/bm_sim/P4Objects.h
antoninbas: no worries, you just need to fully qualify `std::string` (instead of `string`). You have 2 occurrences in the patch (`table_name` parameter in `check_next_nodes` function).

docs/JSON_format.md
```diff
@@ -625,7 +625,7 @@ attributes for these objects are:
   - `actions`: the list of actions (order does not matter) supported by this
   table
   - `next_tables`: maps each action to a next table name. Alternatively, maps
-  special string `__HIT__` and `__MISS__` to a next table name.
+  special string `__HIT__` and `__MISS__` to a next table name. See Note 3 below.
```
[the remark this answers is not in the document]
jafingerhut: Short answer: There is no use case I know of where it makes sense to mix the two kinds of keys.

Gory details:
p4c uses __HIT__ and/or miss as keys in next_tables object if and only if you use `table.apply().hit` or `table.apply().miss` in a way that different code is executed next based upon the value of that expression.  In those cases, the action executed by the table is never used to determine the next node.

In all other cases, p4c uses only the action names (or ids) as keys.  The common case is that all of the next nodes are the same for every action, but if the P4 source code uses `switch (table.apply().action_run) ...` then the next nodes can differ for each action.

src/bm_sim/P4Objects.cpp
```diff
@@ -1575,6 +1575,65 @@ std::vector<MatchKeyParam> parse_match_key(
...
+  //     backend for tables that use none of .hit, .miss, and
+  //     .action_run, and always execute the same code next regardless
+  //     of hit, miss, or which action the table executed.  In that
+  //     case, every action will have the same next node to execute
+  //     regardless of the action.
+  int num_next_nodes = 0;
+  for (const auto &node : cfg_next_nodes) {
+    (void) node;
+    num_next_nodes++;
+  }
```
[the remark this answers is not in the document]
jafingerhut: No reason other than my weak C++-fu.  I will update this, and the ditto comment, to use the size() method.

src/bm_sim/P4Objects.cpp
```diff
@@ -1831,19 +1893,24 @@ P4Objects::init_pipelines(const Json::Value &cfg_root,
...
-      if (cfg_next_nodes.isMember("__HIT__"))
-        table->set_next_node_hit(get_next_node(cfg_next_nodes["__HIT__"]));
-      if (cfg_next_nodes.isMember("__MISS__"))
-        table->set_next_node_miss(get_next_node(cfg_next_nodes["__MISS__"]));
+      if (next_is_hit_miss) {
+          table->set_next_node_hit(get_next_node(cfg_next_nodes["__HIT__"]));
+          table->set_next_node_miss(get_next_node(cfg_next_nodes["__MISS__"]));
+      }
 
       if (cfg_table.isMember("base_default_next")) {
```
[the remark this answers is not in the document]
jafingerhut: I would be happy to document it, if I could tell what it was for :-)

I see it in the JSON files generated by p4c in table definitions, pretty much most of the tables, but I _also_ see in most tables [1] that the `next_tables` key defines a next node for _every_ action name of the table.  It isn't clear to me when the value of `base_default_next` is ever used by BMv2 code.  Even if it is sometimes used, it seems to me that it might be for a feature that P4 the language does not need.

[1] the ones that _don't_ use `table.apply().hit` or `table.apply().miss` features, which are most tables in the test suite (and typical P4 programs in general only use that feature occasionally).

docs/JSON_format.md
```diff
@@ -634,14 +634,14 @@ attributes for these objects are:
     - `action_id`: the id of the default action
     - `action_const`: an optional boolean value which is `true` iff the control
     plane is not allowed to change the default action function. Default value is
-    `false`. It can only be set to `true` for `simple` tables.
+    `false`. It can only be set to `true` for `simple` tables.  See Note 2 below.
     - `action_data`: an optional JSON array where each entry is the hexstring
```
antoninbas, line 638: is this attribute (`action_data`) always set by p4c nowadays?

include/bm/bm_sim/P4Objects.h
jafingerhut: removed in commit 4

include/bm/bm_sim/P4Objects.h
jafingerhut: updated in commit 4

src/bm_sim/P4Objects.cpp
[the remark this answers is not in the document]
jafingerhut: changed in commit 4

src/bm_sim/P4Objects.cpp
```diff
@@ -1575,6 +1575,65 @@ std::vector<MatchKeyParam> parse_match_key(
...
+          << " but either it does not have both of them,"
+          << " or it has other keys that should not be there.",
+          cfg_next_nodes);
+    }
+  } else {
+    *next_is_hit_miss = false;
+    int num_actions = 0;
+    for (const auto &cfg_action : cfg_actions) {
+      (void) cfg_action;
+      num_actions++;
```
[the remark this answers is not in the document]
jafingerhut: changed in commit 4

docs/JSON_format.md
jafingerhut: I committed some additional checks that threw exceptions if `action_id` or `action_data` were ever absent inside of a `default_entry` object value, and they were always found for all p4c tests, so it seems very likely that they are guaranteed to be present.  `default_entry` is always present for tables with type `simple`, too.

src/bm_sim/P4Objects.cpp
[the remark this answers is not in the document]
jafingerhut: `default_entry` is _not_ present for tables with type `indirect` (corresponding to implementation=action_profile) and `indirect_ws` (corresponding to implementation=action_selector).  `default_entry` seems to always be present for tables with type `simple`.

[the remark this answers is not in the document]
jafingerhut: Note that I am not aware of any strong reason, other than similarity to Tofino perhaps, that we could not go ahead and implement full support for default_action in tables that have action profiles or action selectors.

docs/JSON_format.md
jafingerhut: The change to throw an error if `default_entry` is not present in a table with type `simple` led to updating more test cases than I expected.  I can back out those changes if they seem like busy-work.

antoninbas: I don't think we should change these old test cases, it just creates noise. However, I think you should mention it in the Node 2 below.

jafingerhut: Just so I am clear on your meaning, your recommendation is:
(a) document that `default_entry` is included in all simple tables in today's p4c
(b) remove the check that throws an exception if `default_entry` is not present in a simple table
(c) remove all changes to old tests
Yes?

jafingerhut: I have removed the code that threw an error if `default_entry` was absent in a table definition, restoring it back to optional, as it was before this commit.  I also removed all the commits in the old unit test cases that were needed to make those old unit tests pass with such a change.

jafingerhut: > I don't think we should change these old test cases, it just creates noise. However, I think you should mention it in the Node 2 below.

I have added to Note 2 in the JSON docs a mention that p4c always include key `default_entry` for tables with `type` equal to `simple`, and always omits `default_entry` for other types of tables.

src/bm_sim/P4Objects.cpp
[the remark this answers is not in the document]
jafingerhut: Here is the code in the p4c BMv2 back end that specifically checks for, and omits, any default_action definition in the P4_16 source code for tables with action profiles or action selectors: https://github.com/p4lang/p4c/commit/96ec4d56c9acb126db14632dbb00221ccbfef825#diff-e1b9c4c9f99e44821b652694b5a72f9a58784dd8e05572850c0a08b7f7c1c87fR1215-R1221

The commit comment by Mihai was "bmv2 does not support default_action for all tables", so apparently he believed that was true at that time.  For all I know, perhaps it was true for bmv2 in May 2016.

I am happy to try experimenting with a modification to p4c that removes that restriction, and test bmv2 to see if it seems to behave correctly with default_action definitions on tables with action profiles and action selectors.  I agree that if bmv2 does not support them as it is written today, I suspect it would be a very small change to enable it to work for default_action's on such tables.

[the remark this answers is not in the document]
antoninbas: > The commit comment by Mihai was "bmv2 does not support default_action for all tables", so apparently he believed that was true at that time. For all I know, perhaps it was true for bmv2 in May 2016.

Yes that would have been true then given that #551 is from 2018

docs/JSON_format.md
```diff
@@ -625,23 +625,25 @@ attributes for these objects are:
...
+    - `action_id`: the id of the default action.  This is required if
+	`default_entry` is present.
     - `action_const`: an optional boolean value which is `true` iff the control
     plane is not allowed to change the default action function. Default value is
-    `false`. It can only be set to `true` for `simple` tables.
+    `false`. It can only be set to `true` for `simple` tables.  See Note 2 below.
     - `action_data`: an optional JSON array where each entry is the hexstring
     value for an action argument. The size of the array needs to match the
     number of parameters expected by the action function with id `action_id`.
+	This is required if `default_entry` is present.
```
antoninbas, line 642: I don't think this is correct.
The request was more about including in Note 2 that:
> p4c always include the `action_data` attribute in the `default_entry` object.

src/bm_sim/P4Objects.cpp
```diff
@@ -1888,6 +1947,11 @@ P4Objects::init_pipelines(const Json::Value &cfg_root,
 
           table->set_default_default_entry(action, std::move(adata),
                                            is_action_entry_const);
+        } else {
+          throw json_exception(
+              EFormat() << "'default_entry' of table '" << table_name
+              << "' should have key 'action_data'",
+              cfg_default_entry);
```
antoninbas, line 1954: see comment above. I am not sure we need to enforce this new restriction. But I don't feel very strongly about it, so if you think this is more meaningful this way, we can add it. I assume it won't break the test JSON files included in this repo?

src/bm_sim/P4Objects.cpp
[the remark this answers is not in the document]
jafingerhut: If you are OK reviewing this PR as it is right now, I am happy to make a new issue to track the following things, perhaps modified in later PRs:
(a) consider modifying p4c bmv2 back end so that it supports default_action in P4 source code for tables with action profiles or action selectors.
(b) Document the key `base_default_next`, including mentioning what it was for originally, and any known cases where its value affects the behavior of bmv2.

docs/JSON_format.md
antoninbas: If we do add this new restriction, we should update the phrasing above ("an optional JSON array").

src/bm_sim/P4Objects.cpp
jafingerhut: All behavioral-model CI tests are passing, as you can see on Github page for this PR.

I have also run all p4c tests with behavioral-model with the changes on this PR, and this exception never occurs.

I am OK removing the check if you prefer.  The main reason I even include it is that it makes it more clear what behavioral-model can rely on going forward, and reduce the number of cases that we need to think about when modifying code related to this.

docs/JSON_format.md
jafingerhut: Updated in commit 9.

src/bm_sim/P4Objects.cpp
jafingerhut: Updated docs in commit 9 to make it clearer that `action_data` key is required inside of `default_entry`.

src/bm_sim/P4Objects.cpp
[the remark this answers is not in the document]
jafingerhut: p4c issue to track idea (a) above: https://github.com/p4lang/p4c/issues/5101
behavioral-model issue to track idea (b) above: https://github.com/p4lang/behavioral-model/issues/1289

antoninbas (commented):
LGTM

In theory because we are making the rules more restrictive for what constitutes a valid JSON (I'm thinking of `default_entry.action_data` being required), we should update the JSON format version: https://github.com/p4lang/behavioral-model/blob/87e6bf3bf7c870e004eb113dea107d5b8fcbacf4/src/bm_sim/P4Objects.cpp#L46-L49

In theory, the version would become 2.24, and both `max_minor_version` and `min_minor_version` would need to be set to 2.24. That would create a bit of an headache potentially, as you would need to update p4c as well and regenerate all the JSON files there to use version 2.24.

Now given that all JSON files nowadays are generated by p4c, and the updated requirements for the JSON format should match what's p4c has been doing for several years, updating the version in this way may not be worth the trouble...

jafingerhut:
I haven't reviewed the code in detail yet, but there is a min_minor_version and max_minor_version, which could be set to a range of allowed minor versions that the current program expects, and perhaps a different range for what the updated BMv2 from this PR would accept?  I can think later about specific suggestions there, after reviewing what the range checking code does in BMv2.

jafingerhut:
> LGTM
> 
> In theory because we are making the rules more restrictive for what constitutes a valid JSON (I'm thinking of `default_entry.action_data` being required), we should update the JSON format version:
> 
> https://github.com/p4lang/behavioral-model/blob/87e6bf3bf7c870e004eb113dea107d5b8fcbacf4/src/bm_sim/P4Objects.cpp#L46-L49
> 
> In theory, the version would become 2.24, and both `max_minor_version` and `min_minor_version` would need to be set to 2.24. That would create a bit of an headache potentially, as you would need to update p4c as well and regenerate all the JSON files there to use version 2.24.
> 
> Now given that all JSON files nowadays are generated by p4c, and the updated requirements for the JSON format should match what's p4c has been doing for several years, updating the version in this way may not be worth the trouble...

I have looked at the version checking code in BMv2 to remind myself what it does.

It aborts if the major version in the JSON file is different than what BMv2 expects.  The major version has been 2 for quite some time, and we should not change that here.

BMv2 aborts if the JSON file's minor version is larger than max_minor_version compiled into BMv2.  p4c has produced minor version 23 for some time now, and BMv2 has has max_minor_version 23 for a while now.  Reference: https://github.com/p4lang/behavioral-model/blob/main/src/bm_sim/P4Objects.cpp#L544-L559

If we change BMv2 max_minor_version to 24 in this PR, then BMv2 with this PR's changes will continue to accept JSON files with minor version 23.  It will also start accepting JSON files with minor version 24, which p4c never produces, and given one approach to implementing loops in BMv2 that I have thought of recently, I have no near-future plans to change p4c to need a new JSON schema.

Thus I have added another commit today to this PR that changes max_minor_version to 24.  Today's p4c with no changes should continue interoperating with this new BMv2 just fine.

antoninbas:
@jafingerhut so you think setting `min_minor_version` to 24 is not needed?
Technically, a valid 2.23 JSON input (with a missing `default_entry.action_data` attribute) is no longer a valid 2.24 JSON input, so after merging this PR, bmv2 requires a JSON input that is >= 2.24.
But:
- we probably mostly (only?) care about p4c at this stage
- rejecting a JSON input because of an unsupported version number may not be better than rejecting a JSON input because it has a missing `default_entry.action_data` table attribute. In the second case, we actually get a better error message
- we would just be making life harder on ourselves

So after listing these, I tend to agree that setting `min_minor_version` to 24 is  probably not desired :)

antoninbas (commented):
Please update the VERSION number at the top of `docs/JSON_format.md`: https://github.com/p4lang/behavioral-model/blob/87e6bf3bf7c870e004eb113dea107d5b8fcbacf4/docs/JSON_format.md?plain=1#L13

Otherwise, LGTM

jafingerhut:
> Please update the VERSION number at the top of `docs/JSON_format.md`:

Done.

antoninbas (approved):

PR #1290, opened 2025-01-19
Allow indirect tables to have default actions declared const in P4

jafingerhut:


jafingerhut:
@antoninbas I have started a PR on p4c to enable the p4c bmv2 backend to create indirect tables with a `default_entry` key, here: https://github.com/p4lang/p4c/pull/5104

In testing it, I found that if in the P4 source code the `default_action` property of an indirect table is declared `const`, then with the changes on the PR linked above, a JSON file is created where the `action_const` key of `default_entry` exists, and has the value `true`.  For such a JSON file, BMv2 gives an error:

```
Table '<name>' does not have type 'simple' and therefore setting 'action_const' to true is meaningless
```

If an indirect table can support a non-const default_action, I don't see why it could not also support a const default_action -- just support the default_action and don't let the control plane change it later.

Do you have any recollection why this is checked for, and produces the above error?

jafingerhut:
In looking at some of the behaviral-model code, I am suspecting that perhaps its current indirect table implementations _do not_ implement default_action behavior.

By default_action behavior, I mean:

+ do a lookup of the original key
+ if there is a match, then yes the profiles, or groups, are important for indirect tables
+ but if there is a miss, there should be a separate action with whatever action parameters it needs that can be configured independently of all profile actions, or action selector groups/members.

jafingerhut:
And in hacking a bit further, _if_ we wanted p4c, behavioral-model, and the P4Runtime API to all support indirect tables with a source-code-configurable default_action, and a runtime configurable default_action, it seems to require changes in all of these:

+ p4c bmv2 backend
+ behavioral-model
+ P4Runtime specification - The section named "Default Entry" in the spec says: "In this P4Runtime release, we have decided to restrict the default entry for indirect tables — tables with an ActionProfile or ActionSelector implementation property — to a constant NoAction action entry, with the hope that it would simplify the implementation of the P4Runtime service."
+ probably PI implementation of P4Runtime server behavior

If it were just the first two that required small changes, I think I would be interested, but it is becoming a bit of "going down the rabbit hole" for questionable gain.

PR #1293, opened 2025-02-18
Fix #1291 Do not assert when --dump-packet-data option is enabled ...

### LINKED ISSUE #1291: Suggested enhancement: drop any 0-length packets that are output by the deparser, with a message in the log
"Sending" a 0-length packet on an Ethernet interface is not something I have ever heard of anyone wanting to do.  If someone wants some kind of "event packet", they just create a short packet with at least one header in it.

There are corner-case test programs for p4c that p4testgen generates input packets and expected output packets, where this can happen.

While recently testing one of these programs using p4testgen, as part of p4c's CI test suite, on an Ubuntu 24.04 system, I encountered a BMv2 crash due to an assert statement firing while running such a test.  https://github.com/p4lang/p4c/issues/5130

For a while now, when p4testgen generates such a test, it replaces the 0-length output packet with a 4-byte packet having contents `02000000` in hexadecimal, because many Ubuntu Linux systems apparently have a behavior in BMv2 and/or its pcap library that cause it to send a 0-length packet as that particular 4-byte packet.

If it is true that this was a pcap library bug, and that it has been fixed in Ubuntu 24.04, these p4c tests will begin failing once p4c CI has been enabled for Ubuntu 24.04.

The recommendation I am giving here is NOT high priority.  It is a corner-case check for something that I don't expect to happen in real production systems, but seems useful for enabling some of these funkier p4testgen cases to continue working (with a small corresponding change in p4testgen): modify BMv2 so that if a 0-length packet with 0 unparsed bytes is output by the deparser, drop it.  No 0-length packet should ever be sent to an output port, nor recirculated.  If logging is enabled, there should be a message in the log that such a packet was encountered and it is being dropped, as an aid to those debugging the behavior.

jafingerhut:
... when sending a length 0 packet.

Granted this is an unusual situation, but there are test programs for p4c and p4testgen where p4testgen creates tests where the expected output packet is 0 bytes long, and if you run them while enabling --dump-packet-data N option for behavioral-model, simple_switch crashes with an assert.

### DIFF
--- a/src/bm_sim/dev_mgr.cpp
+++ b/src/bm_sim/dev_mgr.cpp
@@ -282,10 +282,11 @@ DevMgr::clear_port_stats(port_t port_num) {
 std::string
 DevMgr::sample_packet_data(const char *buffer, int len) {
   size_t amount = std::min(dump_packet_data, static_cast<size_t>(len));
-  std::ostringstream ret;
-  if (amount > 0) {
-      utils::dump_hexstring(ret, &buffer[0], &buffer[amount]);
+  if (amount == 0) {
+      return "";
   }
+  std::ostringstream ret;
+  utils::dump_hexstring(ret, &buffer[0], &buffer[amount]);
   return ret.str();
 }
 
--- a/src/bm_sim/dev_mgr.cpp
+++ b/src/bm_sim/dev_mgr.cpp
@@ -282,7 +282,6 @@ DevMgr::clear_port_stats(port_t port_num) {
 std::string
 DevMgr::sample_packet_data(const char *buffer, int len) {
   size_t amount = std::min(dump_packet_data, static_cast<size_t>(len));
-  assert(amount > 0);
   std::ostringstream ret;
   utils::dump_hexstring(ret, &buffer[0], &buffer[amount]);
   return ret.str();
--- a/src/bm_sim/dev_mgr.cpp
+++ b/src/bm_sim/dev_mgr.cpp
@@ -283,7 +283,9 @@ std::string
 DevMgr::sample_packet_data(const char *buffer, int len) {
   size_t amount = std::min(dump_packet_data, static_cast<size_t>(len));
   std::ostringstream ret;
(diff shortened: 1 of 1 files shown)


jafingerhut:
This comment shows a list of 6 p4c / p4testgen test cases that fail if you enable `--dump-packet-data N` on `simple_switch` while running those tests: https://github.com/p4lang/p4c/pull/5132#issuecomment-2666738172

antoninbas (approved):
LGTM, I assume `dump_hexstring` still behaves correctly in this case?

It may still be worth having an early return is `amount` is 0, but no strong preference from me

jafingerhut:
> LGTM, I assume `dump_hexstring` still behaves correctly in this case?
> 
> It may still be worth having an early return is `amount` is 0, but no strong preference from me

commit 2 only calls dump_hexstr if the packet length is greater than 0, to be extra-cautious.  I am currently assuming that the final return is guaranteed to return an empty string in that case, but I will run a test soon to verify.

src/bm_sim/dev_mgr.cpp
```diff
@@ -282,9 +282,10 @@ DevMgr::clear_port_stats(port_t port_num) {
 std::string
 DevMgr::sample_packet_data(const char *buffer, int len) {
   size_t amount = std::min(dump_packet_data, static_cast<size_t>(len));
-  assert(amount > 0);
   std::ostringstream ret;
-  utils::dump_hexstring(ret, &buffer[0], &buffer[amount]);
+  if (amount > 0) {
```
antoninbas, line 286: IMO, it is usually more readable to have an early return:
```
if (amount == 0) {
        return "";
}
std::ostringstream ret;
...
```
it may be optimized away, but there is no need to instantiate `std::ostringstream ret` if amount is 0.

jafingerhut: Thanks as always for the explicit review comments, that even a C++ dilettante like me can follow :-)

PR #1297, opened 2025-03-11
Convert bf_lpm_trie.c into C++ style

Hoooao:
Currently, BMv2 uses some legacy code written in C with [bf_lpm_trie.c](https://github.com/p4lang/behavioral-model/blob/d12eefc7bc19fb4da615b1b45c1235899f2e4fb1/src/bf_lpm_trie/bf_lpm_trie.c) being one of the instances.

~Currently, one test fails: `TableBadInputKey/1.NonByteAligned`~


antoninbas:
When this PR is ready for review, we will need a comparison of both implementations when running the `test_LPM_match_1` stress test program (https://github.com/p4lang/behavioral-model/tree/main/tests/stress_tests).

Hoooao:
Comparison:


**C++ impl** 
```
Imported 1000 packets of size 512 (512000 bytes)
Processing 1000000 packets...
Done.
Test took 9359 ms.
Switch was processing 106849 packets per second.
PASS test_LPM_match_1 (exit status: 0)
```

**C impl**
```
Imported 1000 packets of size 512 (512000 bytes)
Processing 1000000 packets...
Done.
Test took 10487 ms.
Switch was processing 95356 packets per second.
PASS test_LPM_match_1 (exit status: 0)
```

src/bf_lpm_trie/bf_lpm_trie/bf_lpm_trie.h
antoninbas, line 1: can we just remove `src/bf_lpm_trie/` and use `src/bm_sim/lpm_trie.h` / `src/bm_sim/lpm_trie.cpp` for the implementation?

src/bf_lpm_trie/bf_lpm_trie/bf_lpm_trie.h
```diff
@@ -22,40 +22,124 @@
...
+               ? (key < other.key)
+               : (prefix_length > other.prefix_length);
+  }
+};
 
-bool bf_lpm_trie_lookup(const bf_lpm_trie_t *trie, const char *key,
-			value_t *pvalue);
+class PrefixesVec {
+public:
+  void insertPrefix(uint8_t prefix_length, byte_t key, value_t value);
```
antoninbas, line 70: I think the project uses snake case for all method and variable names

src/bf_lpm_trie/bf_lpm_trie/bf_lpm_trie.h
Hoooao: Updated.


```
Imported 1000 packets of size 512 (512000 bytes)
Processing 1000000 packets...
Done.
Test took 15130 ms.
Switch was processing 66093 packets per second.
PASS test_LPM_match_1 (exit status: 0)
```

qobilidop:
In case you didn't know, the DCO thing is further explained [here](https://github.com/p4lang/p4c/blob/main/CONTRIBUTING.md#contributing-license). It shall be fairly straightforward to fix it and pass the CI.

src/bf_lpm_trie/bf_lpm_trie/bf_lpm_trie.h
antoninbas: why the large drop in performance?

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,130 @@
...
+// Hao: value_t is a c type, equivalent to unsigned long, but need to make
+// sure
+typedef std::uintptr_t value_t;
+typedef unsigned char byte_t;
+
+static_assert(sizeof(value_t) == sizeof(std::uintptr_t), "Invalid type sizes");
+
+class Node;
+
+struct Branch {
```
antoninbas, line 45: I am not sure all the classes need to be included in this header?
Don't you just need the `class Node` forward declaration (for the pointer in `LPMTrie`)?

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,130 @@
...
+  byte_t v;
+  std::unique_ptr<Node> next;
+};
+
+class BranchesVec {
```
antoninbas, line 50: given that in C++, you have `std::vector` available to you, there is really no point in defining this class, which just wraps a `std::vector<Branch>` object.

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,130 @@
...
+  value_t value;
+  // replace prefix_cmp, order from long to short
+  bool operator<(const Prefix &other) const {
+    return (prefix_length == other.prefix_length)
+               ? (key < other.key)
+               : (prefix_length > other.prefix_length);
+  }
+};
+
+class PrefixesVec {
```
antoninbas, line 73: same comment as for `BranchesVec`: this class doesn't seem useful and just increases code complexity

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,226 @@
...
+ */
+
+/*
+ * Antonin Bas
+ *
+ */
+
+#include <lpm_trie.h>
+
+#define _unused(x) ((void)(x))
```
antoninbas, line 24: use https://github.com/p4lang/behavioral-model/blob/d12eefc7bc19fb4da615b1b45c1235899f2e4fb1/include/bm/bm_sim/_assert.h#L37

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,130 @@
...
+ public:
+  void insert_prefix(uint8_t prefix_length, byte_t key, value_t value);
+  Prefix *get_prefix(uint8_t prefix_length, byte_t key);
+  bool delete_prefix(uint8_t prefix_length, byte_t key);
+
+  inline bool is_empty() const { return prefixes.empty(); }
+  inline Prefix *back() { return prefixes.back().get(); }
+
+  std::vector<std::unique_ptr<Prefix>> prefixes;
```
antoninbas, line 82: what's the reason for the extra indirection here?

antoninbas (commented):
some more comments

src/bf_lpm_trie/bf_lpm_trie/bf_lpm_trie.h
Hoooao: Oh, forgot to charge while running, thus CPU frequency was low.. Here is the normal case. 
```
Imported 1000 packets of size 512 (512000 bytes)
Processing 1000000 packets...
Done.
Test took 9620 ms.
Switch was processing 103950 packets per second.
PASS test_LPM_match_1 (exit status: 0)
```

src/bm_sim/lpm_trie.h
Hoooao: Can you elaborate? Do you mean some classes should be separated to another header file?

src/bm_sim/lpm_trie.h
Hoooao: to make it consistent with the c impl which passed ptr around and allocated the prefix on heap. But seems unnecessary as only value is read. Will update it.

src/bm_sim/lpm_trie.h
Hoooao: Updated.

src/bm_sim/lpm_trie.h
Hoooao: Updated.

src/bm_sim/lpm_trie.h
antoninbas: I meant that there doesn't seem to be a reason for declaring these classes in a header file to begin with. We can move them to `lpm.trie.cpp`

src/bm_sim/lpm_trie.h
antoninbas: unless I am missing something, the C implementation was using a "vector" of  `prefix_t` and not a vector of `*prefix_t`: https://github.com/p4lang/behavioral-model/blob/d12eefc7bc19fb4da615b1b45c1235899f2e4fb1/src/bf_lpm_trie/bf_lpm_trie.c#L205

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+void Node::insert_prefix(uint8_t prefix_length, byte_t key, value_t value) {
+  auto it = search_prefix_with_key(prefix_length, key);
+
+  if (it != prefixes.end()) {
+    (*it).value = value;
+    return;
+  }
+  prefixes.insert(it, {prefix_length, key, value});
```
antoninbas, line 55: my C++ is a bit rusty, but it seems that `emplace_back` would be a better choice here?

https://en.cppreference.com/w/cpp/container/vector/emplace_back

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,207 @@
...
+                       [](const Branch &b, byte_t val) { return b.v < val; });
+
+  return (it != branches.end() && it->v == byte) ? it->next.get() : nullptr;
+}
+
+void Node::insert_prefix(uint8_t prefix_length, byte_t key, value_t value) {
+  auto it = search_prefix_with_key(prefix_length, key);
+
+  if (it != prefixes.end()) {
+    (*it).value = value;
```
antoninbas, line 52: why not `it->value = value`?

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,207 @@
...
+    return;
+  }
+  prefixes.insert(it, {prefix_length, key, value});
+}
+
+bool Node::get_prefix(uint8_t prefix_length, byte_t key, value_t *value) {
+  auto it = search_prefix_with_key(prefix_length, key);
+  if (it != prefixes.end()) {
+    *value = (*it).value;
```
antoninbas, line 61: ditto

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,106 @@
...
+#include <vector>
 
 namespace bm {
 
-static_assert(sizeof(value_t) == sizeof(uintptr_t),
-              "Invalid type sizes");
+// Hao: value_t is a c type, equivalent to unsigned long, but need to make
+// sure
+typedef std::uintptr_t value_t;
+typedef unsigned char byte_t;
```
antoninbas, line 39: the code base uses `using` exclusively, not `typedef`

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,106 @@
...
+
+static_assert(sizeof(value_t) == sizeof(std::uintptr_t), "Invalid type sizes");
```
antoninbas, line 41: I don't see how the static assert could ever be violated, given that you typedef `std::uintptr_t` to `value_t`...

src/bm_sim/lpm_trie.h
antoninbas, line 39: I'd prefer not to have these at the top-level. You can define the type aliases you need as part of the `LPMTrie` scope ideally.

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,106 @@
...
   /* Move constructor */
-  LPMTrie(LPMTrie&& other) noexcept
-  : key_width_bytes(other.key_width_bytes), trie(other.trie) {}
+  LPMTrie(LPMTrie &&other) noexcept : key_width_bytes(other.key_width_bytes) {
+    root.reset(other.root.release());
+  }
 
   ~LPMTrie() {
-    bf_lpm_trie_destroy(trie);
+    // Hao: may need to clearn up memory for trie
```
antoninbas, line 123: which memory would you need to explicitly clean up with the new implementation?

src/bm_sim/lpm_trie.h
```diff
@@ -21,32 +21,106 @@
...
+struct Branch {
+  byte_t v;
+  std::unique_ptr<Node> next;
+};
+
+struct Prefix {
+  uint8_t prefix_length;
+  byte_t key;
+  value_t value;
+  // replace prefix_cmp, order from long to short
```
antoninbas, line 54: the comment is confusing. Was `prefix_cmp` a function in the original C code?

src/bm_sim/lpm_trie.h
Hoooao: oh, forgot to remove this, in C code it was uintptr_t not std::uintptr_t

src/bm_sim/lpm_trie.h
Hoooao: Yeah, but what i meant was, as they are allocated on heap using alloc/realloc, their addressing would be more deterministic than addressing an element in std::vector, so I though storing the pointer would be better. 

Yet, rn, i looked through the code and there are no spots in the code that require such addressing for the prefix object, but merely the value reading. I have changed it to vector<Prefix>.

src/bm_sim/lpm_trie.h
Hoooao: legacy cmts, removed

src/bm_sim/lpm_trie.h
Hoooao: yes, removed the cmts

src/bm_sim/lpm_trie.h
Hoooao: actually, i'd restore it to `static_assert(sizeof(value_t) == sizeof(uintptr_t), "Invalid type sizes");` for now,  as there are other spots in some module that use the old uintptr_t. just in case somehow different platform have different length for the type..

src/bm_sim/lpm_trie.cpp
Hoooao: yes, but previously i didn't write the corresponding constructor.

src/bm_sim/lpm_trie.h
Hoooao: hmmm ok i see, i just usually prefer to put all decls to header file.

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
+ * Copyright 2021 VMware, Inc.
```
antoninbas, line 2: @jafingerhut do you know if it would be ok to just use `Copyright 2025 P4lang Authors` here?

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Antonin Bas
+ *
+ */
```
antoninbas, line 20: no need to include my name here

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
...
+  while (current_node) {
+    if (key_width == 0) {
+      return current_node->get_empty_prefix(value);
+    }
+
+    for (auto &prefix : current_node->get_prefixes()) {
+      byte = static_cast<byte_t>(key[key_idx]) >> (8 - prefix.prefix_length);
+      if (byte == prefix.key) {
+        *value = prefix.value;
+        return true;
```
antoninbas, line 279: this doesn't seem to match the previous code:

```c
    for (i = 0; i < current_node->prefixes.size; i++) {
      p = &current_node->prefixes.prefixes[i];
      byte = (unsigned) (unsigned char) *key >> (8 - p->prefix_length);
      if (p->key == byte) {
        found = true;
        *pvalue = p->value;
        break;
      }
    }
```

if there is a match, you have to keep looking for a longer match, you cannot stop immediately.

For example, if I have a `10.0.0.0/16` entry and a `10.0.0.0/24` entry in the table, and I lookup address `10.0.0.100`, I cannot stop when I match the first entry, I have to keep looking, as the second entry (`10.0.0.0/24`) is the correct one here.

If we don't have a unit test catching this issue, it's kind of a big deal IMO and you should add one as part of this PR. Or am I misunderstanding the new implementation?

src/bm_sim/lpm_trie.h
```diff
@@ -21,77 +21,79 @@
...
 
   /* Move constructor */
-  LPMTrie(LPMTrie&& other) noexcept
-  : key_width_bytes(other.key_width_bytes), trie(other.trie) {}
+  LPMTrie(LPMTrie &&other) noexcept;
 
-  ~LPMTrie() {
-    bf_lpm_trie_destroy(trie);
-  }
+  ~LPMTrie();
```
antoninbas, line 53:
```suggestion
  ~LPMTrie() = default;
```

and remove the manual implementation. My C++ is rusty but I think that's always better to use `default`?

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
...
+  byte_t byte;
+  size_t key_width = key_width_bytes;
+
+  int key_idx = 0;
+  while (current_node) {
+    if (key_width == 0) {
+      return current_node->get_empty_prefix(value);
+    }
+
+    for (auto &prefix : current_node->get_prefixes()) {
```
antoninbas, line 275: I don't know if it's right to expose the internal `prefixes` using a `Node::get_prefixes` method. Do you think the functionality of this `for` loop should be moved to a dedicated `Node` method, given that this is the only usage?

src/bm_sim/lpm_trie.cpp
jafingerhut: As at least a reasonable placeholder for what I think it might end up being, perhaps `Copyright 2025 Contributors to the P4 Project`.  If we end up with something different, I think it will get changed later.

src/bm_sim/lpm_trie.h
```diff
@@ -21,77 +21,79 @@
...
   void insert_prefix(const ByteContainer &prefix, int prefix_length,
-                     uintptr_t value) {
-    bf_lpm_trie_insert(trie, prefix.data(), prefix_length,
-                       static_cast<value_t>(value));
+                     value_t value) {
+    std::string prefix_str(prefix.data(), prefix.size());
+    insert(prefix_str, prefix_length, value);
   }
 
+  bool remove(const std::string &prefix, int prefix_length);
```
matthewtlam, line 68: Let's keep this name consistent with the other. 

NIT:  delete_prefix

src/bm_sim/lpm_trie.h
```diff
@@ -21,77 +21,79 @@
...
 
   /* Move assignment operator */
-  LPMTrie &operator=(LPMTrie &&other) noexcept {
-    key_width_bytes = other.key_width_bytes;
-    std::swap(trie, other.trie);
-    return *this;
-  }
+  LPMTrie &operator=(LPMTrie &&other) noexcept;
 
+  void insert(const std::string &prefix, int prefix_length, value_t value);
```
matthewtlam, line 61: Let's keep this name consistent

NIT: insert_prefix

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
...
+    assert(key_width_bytes <= 64);
+    root = std::make_unique<Node>();
+  }
+
+LPMTrie::LPMTrie(LPMTrie &&other) noexcept :
+    key_width_bytes(other.key_width_bytes) {
+  root.reset(other.root.release());
+}
+
+LPMTrie::~LPMTrie() {}
```
matthewtlam, line 185: No need for this if we set the destructor to default as Antonin mentioned

src/bm_sim/lpm_trie.h
matthewtlam: +1

src/bm_sim/lpm_trie.h
```diff
@@ -21,77 +21,79 @@
...
-                     uintptr_t value) {
-    bf_lpm_trie_insert(trie, prefix.data(), prefix_length,
-                       static_cast<value_t>(value));
+                     value_t value) {
+    std::string prefix_str(prefix.data(), prefix.size());
+    insert(prefix_str, prefix_length, value);
   }
 
+  bool remove(const std::string &prefix, int prefix_length);
   bool delete_prefix(const ByteContainer &prefix, int prefix_length) {
```
matthewtlam, line 69: Let's move the implementation details of insert_prefix, delete_prefix, has_prefix, retrieve_value in the .cc file

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
...
+ *
+ */
+
+#include <bm/bm_sim/_assert.h>
+#include <lpm_trie.h>
+
+namespace bm {
+
+struct Branch {
+  byte_t v;
```
matthewtlam, line 28: Naming the parameter the value v is not good practice.

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,294 @@
...
+    current_node = current_node->get_next_node(key[key_idx]);
+    key_width--;
+    key_idx++;
+  }
+
+  return false;
+}
+
+void LPMTrie::clear() {
+  root.reset(new Node());
```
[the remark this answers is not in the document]
Hoooao: I think it is fine as `root` is a`unique_ptr`.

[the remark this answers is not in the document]
matthewtlam: I did not see that root was a unique pointer. Thanks for pointing me in the right direction. Yup, it should be fine since root is unique.

src/bm_sim/lpm_trie.cpp
Hoooao: you are right. I misunderstood `found` thus early returned.

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+  bool is_empty() const { return prefixes.empty() && branches.empty(); }
+
+ private:
+  std::map<std::pair<uint8_t, byte_t>, value_t> prefixes;
+  std::unordered_map<byte_t, Branch> branches;
+  Node *parent;
+  byte_t child_id;
+
+  std::optional<value_t> search_prefix_with_key(uint8_t prefix_length,
```
antoninbas, line 66: why not also use `std::optional<value_t>` for `get_prefix`?

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+  // Branch
+  void add_branch(byte_t byte, std::unique_ptr<Node> nextNode);
+
+  bool delete_branch(byte_t byte);
+
+  bool is_empty() const { return prefixes.empty() && branches.empty(); }
+
+ private:
+  std::map<std::pair<uint8_t, byte_t>, value_t> prefixes;
```
antoninbas, line 61: this feels different from the earlier version I reviewed. Did you switch from a `vector` to a `map`? A sorted vector should be faster to search (and more compact) than a map, so why make this change?

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,278 @@
...
+  std::unordered_map<byte_t, Branch> branches;
```
antoninbas, line 62: even here, I am not convinced that `unordered_map` is a better choice than a `vector` but it's probably easier to justify

antoninbas:
IMO, it's better to stick to the vector-based implementation (which aligns with the previous C implementation) in this PR, then submit follow up PR with empirical evidence that switching to maps is not a performance regression.

src/bm_sim/lpm_trie.cpp
Hoooao: yes

tests/test_tables.cpp
```diff
@@ -2383,6 +2382,37 @@ class AdvancedLPMTest : public AdvancedTest {
     return table->add_entry(match_key, &action_fn, ActionData(), handle);
   }
 
+  MatchErrorCode add_entry_2(entry_handle_t *handle) {
```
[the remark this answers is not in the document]
Hoooao: wrapped them in a func

src/bm_sim/lpm_trie.h
Hoooao: since LPMTrie has a unique ptr to Node, it has to know the size of it in destructor, so i put the default impl to .cc file, and have to leave the decl here.

matthewtlam:
Please fix the build errors @Hoooao .
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/unique_ptr.h:79:16: error: invalid application of 'sizeof' to an incomplete type 'bm::Node'
        static_assert(sizeof(_Tp)>0,
                      ^~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/unique_ptr.h:292:4: note: in instantiation of member function 'std::default_delete<bm::Node>::operator()' requested here
          get_deleter()(std::move(__ptr));
          ^
./lpm_trie.h:46:3: note: in instantiation of member function 'std::unique_ptr<bm::Node, std::default_delete<bm::Node> >::~unique_ptr' requested here
  ~LPMTrie() = default;
  ^
./lpm_trie.h:35:7: note: forward declaration of 'bm::Node'
class Node;
      ^
1 error generated.
make[2]: *** [Makefile:587: lookup_structures.lo] Error 1
```

matthewtlam:
@antoninbas Looking for final review from you

src/bm_sim/lpm_trie.h
matthewtlam: We set default in the implementation file otherwise it would create the following build error 

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/unique_ptr.h:79:16: error: invalid application of 'sizeof' to an incomplete type 'bm::Node'
        static_assert(sizeof(_Tp)>0,
                      ^~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/unique_ptr.h:292:4: note: in instantiation of member function 'std::default_delete<bm::Node>::operator()' requested here
          get_deleter()(std::move(__ptr));
          ^
./lpm_trie.h:46:3: note: in instantiation of member function 'std::unique_ptr<bm::Node, std::default_delete<bm::Node> >::~unique_ptr' requested here
  ~LPMTrie() = default;
  ^
./lpm_trie.h:35:7: note: forward declaration of 'bm::Node'
class Node;
      ^
1 error generated.
make[2]: *** [Makefile:587: lookup_structures.lo] Error 1
```

antoninbas (approved):
LGTM, thanks for your contribution.

src/bm_sim/lpm_trie.cpp
```diff
@@ -0,0 +1,293 @@
...
+  std::vector<Prefix> prefixes;
+  Node *parent;
+  byte_t child_id;
+
+  std::vector<Prefix>::iterator search_prefix_with_key(uint8_t prefix_length,
+                                                       byte_t key);
+};
+
+Node *Node::get_next_node(byte_t byte) const {
+  auto it = std::lower_bound(
```
fruffy, line 81: @Hoooao This seems to fail in more recent versions of gcc:

https://github.com/p4lang/p4c/actions/runs/15490297907/job/43614080729?pr=5310#step:5:3194

fruffy: Unclear why the failure didn't occur here, maybe older CI?

PR #1299, opened 2025-03-19
Convert bmi_interface.c to bmi_interface.cpp

Vineet1101:


Vineet1101:
@antoninbas @jafingerhut  can you provide some guidance and suggestion on this task

Vineet1101:
`int bmi_interface_create(bmi_interface_t **bmi, const char *device) {
  bmi_interface_t *bmi_ = malloc(sizeof(bmi_interface_t));

  if(!bmi_) return -1;

  bmi_->pcap_input_dumper = NULL;
  bmi_->pcap_output_dumper = NULL;

  char errbuf[PCAP_ERRBUF_SIZE];
  bmi_->pcap = pcap_create(device, errbuf);

  if(!bmi_->pcap) {
    free(bmi_);
    return -1;
  }

  if(pcap_set_promisc(bmi_->pcap, 1) != 0) {
    pcap_close(bmi_->pcap);
    free(bmi_);
    return -1;
  }`  in this code it initializes the attributes using pointer so in c++ code what should i use RAII or a unique pointer 
@antoninbas

Vineet1101:
@antoninbas  should I also update the .h file to .hpp file and modify its code also. I have completed the conversion of .c file to .cpp file

PR #1300, opened 2025-04-07
git commit -m "Refactor: Split LpmTrie implementation into .h and .cp…

kris70lesgo:
🚀 Refactor: Convert LPM Trie from C to C++ and Split Header/Implementation
Summary:
This PR modernizes the existing LPM Trie implementation by converting it from C to idiomatic C++ and splitting it into separate implementation (.cpp) and declaration (.h) files.

Key Changes:
Converted C-style code to C++ using:

std::unique_ptr for memory management

std::vector, std::optional, std::string_view for cleaner logic

RAII principles and modern C++ containers

Renamed bf_lpm_trie.hpp to bf_lpm_trie.h

Created bf_lpm_trie.cpp and moved all method definitions there

Ensured namespace encapsulation and safer memory handling

### DIFF
--- a/include/bf_lpm_trie/bf_lpm_trie.h
+++ b/include/bf_lpm_trie/bf_lpm_trie.h
@@ -1,0 +1,47 @@
+#ifndef BF_LPM_TRIE_H
+#define BF_LPM_TRIE_H
+
+#include <cstdint>
+#include <vector>
+#include <memory>
+#include <optional>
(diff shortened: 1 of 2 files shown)


antoninbas:
Thanks for your PR, but we have another one that has been going through several rounds of review already: https://github.com/p4lang/behavioral-model/pull/1297

PR #1301, opened 2025-04-11
Add support for multiple applications of the same table in a pipeline 

### LINKED ISSUE #1286: Generalize BMv2 JSON file format to enable the same identical table to be applied multiple times in a single pipeline
I am considering making changes to BMv2 and p4c’s BMv2 back end that would allow compiling and executing P4 programs that apply the same table multiple times during the invocation of a single ingress or egress control.

Root cause for current restriction:

The BMv2 JSON data has a "tables" key, where for each table it has a "next_tables" key, with a value that is one of these two things:
+ {action-name-executed -> next-table-or-condition-to-execute}
+ {hit/miss -> next-table-or-conditional-to-execute}

Thus each named table in the P4 source code, if it is applied more than once in the same pipeline, the current BMv2 JSON schema can only represent it if all such table applies for a table t1 have the same "next_tables" value.

Motivation: Mainly personal dissatisfaction that this restriction has existed since BMv2 was originally developed, and having answered occasional questions about it over the years to newcomers.

Plan to generalize:

Create a new kind of object in a pipeline, perhaps named "table_applies", that is a list of nodes, each of which has the following fields:

+ "next_tables", exactly like "next_tables" is now in a "table", except that instead of referring to next-tables, it will be next-table-applies.
+ "table", the name of the table to apply when this "table_applies" node is executed.

Remove the "next_tables" field from "tables" objects.

Normally there will be exactly one "table_applies" node for each "table", but if a table is applied multiple times in the same pipeline, there will be one "table_applies" node for each such apply call, all pointing at the same "table" object.

"conditionals" nodes would also never point at "tables" in their "next_tables" fields, instead pointing at table-applies nodes.

There would be a new minor version number created in the BMv2 JSON file in order to take advantage of this new schema.

I think it would be best to continue supporting the current BMv2 JSON schema for at least a year or two, given how prevalent it is.  That will require checking the version number in a few places during processing of the BMv2 JSON file.

sudhanshu112233shukla:
# Description
This PR implements support for applying the same table multiple times in a pipeline while maintaining backward compatibility with existing JSON formats.

Fixes #1286

## Key Changes
- Added a new `table_apply.h` and `table_apply.cpp` to implement the TableApply class that represents a specific application of a table in the control flow
- Modified P4Objects to support table_applies with unique IDs and names
- Updated the JSON format to version 3.0, adding a new "table_applies" array to each pipeline
- Moved the "next_tables" field from table objects to table_apply objects
- Added methods to get and manage TableApply objects in P4Objects
- Added comprehensive unit tests to verify the new functionality
- Updated documentation in JSON_format.md to explain the new format

## Backward Compatibility
This implementation maintains full backward compatibility with existing JSON formats (version 2.x). The code checks the JSON version and only processes table_applies for version 3.0+. Legacy JSON files will continue to work without modification.

## Testing
- Added unit tests that verify both the new functionality and backward compatibility
- Created test_table_apply.cpp to test the TableApply class and its integration with P4Objects
- All existing tests continue to pass

## Documentation
- Updated JSON_format.md with details about the new "table_applies" objects and version 3.0 changes
- Added explanatory comments in the code to clarify the implementation


jafingerhut:
I have not looked at the code changes in detail yet.

It would be great if you were able to get all CI build & tests passing before we go through code review.  Feel free to ask questions about how to do that if you get stuck, but I would start by looking at the section "Some checks were not successful" on the PR's web page here: https://github.com/p4lang/behavioral-model/pull/1301, and clicking on the names of tests that are failing to see the logs.  Try to determine why the tests are failing, and in some cases it is helpful if you can reproduce the failures on your local development system, too.

sudhanshu112233shukla:
Sure sir, I’ll do it. I will follow your instructions and reach out if I face any difficulties. I will check the workflow as well.

jafingerhut:
It is not clear to me why the CI tests all show status "Exepcted - Waiting for status to be reported" when the last commit to this PR was yesterday.  It seems either they never started, or they started and are "hung" somehow.  It would be nice to see the CI test results for these proposed changes before reviewing them in detail.

@sudhanshu112233shukla One way to "kick" the CI process is to add another commit to the PR.  That usually causes CI tests to be started again.  In this particular case, I noticed that your PR removed the file Dockerfile.  That is not required for your changes, is it?  If not, that file should _not_ be removed by the PR.  Pushing another commit to this PR that restores the existence of that file in the repo would be one way to make CI tests run again.

Alternately, any simple inconsequential whitespace-only change in a README or a comment is another way.

fruffy:
> It is not clear to me why the CI tests all show status "Exepcted - Waiting for status to be reported" when the last commit to this PR was yesterday. It seems either they never started, or they started and are "hung" somehow. It would be nice to see the CI test results for these proposed changes before reviewing them in detail.
> 
> @sudhanshu112233shukla One way to "kick" the CI process is to add another commit to the PR. That usually causes CI tests to be started again. In this particular case, I noticed that your PR removed the file Dockerfile. That is not required for your changes, is it? If not, that file should _not_ be removed by the PR. Pushing another commit to this PR that restores the existence of that file in the repo would be one way to make CI tests run again.
> 
> Alternately, any simple inconsequential whitespace-only change in a README or a comment is another way.

I assume security feature since it is a first-time contributor editing CI workflow files. Let me check our settings but I do not think we can manually run them here...

PR #1304, opened 2025-05-26
Adding pcap mock test. finding and fixing bugs using it.

at-nojavan:
Added a libpcap mock for testing the BMI code in P4 behavioral-model by injecting packets using the actual BMI code instead of their nanomsg based injector (which is bypassing the BMI code).
It is not meant to be a full mock of libpcap, it only implements the functions required for this special purpose.

This test was used to find and fix 3 bugs.
Bugs fixed:

bmi_port.c:

In bmi_port_destroy_mgr function:
The previous code removed only half of the interfaces because it was iterating through an array while removing its members.

In _bmi_port_interface_remove function:
Bug fix to prevent a deadlock. Function uses "write" to write on a pipe and waits for the answer using "read" while it holds the lock "port_mgr->lock". Meanwhile, the "run_select" thread requires to get that lock once per iteration of the loop, causing a deadlock.

dev_mgr_bmi.cpp:

In BmiDevMgrImp class:
p_monitor should be stopped in destructor and before destruction of port_mgr to prevent pure virtual function call.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -324,6 +324,7 @@ AC_CONFIG_FILES([Makefile
 		targets/l2_switch/learn_client/Makefile
 		targets/simple_switch/Makefile
 		targets/simple_switch/tests/Makefile
+    	targets/simple_switch/tests/pcap_mock/Makefile
 		targets/simple_switch/tests/CLI_tests/Makefile
 		targets/psa_switch/Makefile
 		targets/psa_switch/tests/Makefile
--- a/src/BMI/bmi_port.c
+++ b/src/BMI/bmi_port.c
@@ -319,7 +319,9 @@ static int _bmi_port_interface_remove(bmi_port_mgr_t *port_mgr, int port_num) {
 
   char buf[1] = {'\x00'};
   write(port_mgr->socketpairfd[0], buf, sizeof(buf));
+  pthread_rwlock_unlock(&port_mgr->lock);
   read(port_mgr->socketpairfd[0], buf, sizeof(buf));
+  pthread_rwlock_wrlock(&port_mgr->lock);
 
   if (bmi_interface_destroy(port->bmi) != 0) return -1;
 
@@ -343,7 +345,7 @@ int bmi_port_destroy_mgr(bmi_port_mgr_t *port_mgr) {
   pthread_rwlock_wrlock(&port_mgr->lock);
   bmi_port_t *port;
   int i;
-  for (i = 0; i < port_mgr->port_count; i++) {
+  for (i = port_mgr->port_count - 1; i >= 0; i--) {
     port = &port_mgr->ports[i];
     _bmi_port_interface_remove(port_mgr, port->port_num);
   }
--- a/targets/simple_switch/tests/pcap_mock/Makefile.am
+++ b/targets/simple_switch/tests/pcap_mock/Makefile.am
@@ -1,0 +1,10 @@
+AM_CXXFLAGS = -g -pthread -std=c++20
+AM_LDFLAGS = -L.
+LDLIBS = -lpcap
+
+bin_PROGRAMS = test_pcap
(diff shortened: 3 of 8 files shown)


src/BMI/bmi_port.c
```diff
@@ -319,7 +319,9 @@ static int _bmi_port_interface_remove(bmi_port_mgr_t *port_mgr, int port_num) {
 
   char buf[1] = {'\x00'};
   write(port_mgr->socketpairfd[0], buf, sizeof(buf));
+  pthread_rwlock_unlock(&port_mgr->lock);
   read(port_mgr->socketpairfd[0], buf, sizeof(buf));
+  pthread_rwlock_wrlock(&port_mgr->lock);
```
antoninbas, line 324: I agree there is a possible deadlock and it needs to be fixed, however I don't see any justification that the change is correct, and no explanation as to the placement of the `unlock` instruction (why not right after the `FD_CLR` call?).
For example, any possible issue if there is a concurrent call to `bmi_port_interface_add` with the same `port_num`?

targets/simple_switch/tests/pcap_mock/Makefile.am
```diff
@@ -0,0 +1,10 @@
+AM_CXXFLAGS = -g -pthread -std=c++20
```
antoninbas, line 1: The codebase has not been updated to C++20 yet, so please stick to C++17, or submit a separate change to update the code base to C++20

targets/simple_switch/tests/test_packet_redirect_bmi.cpp
antoninbas, line 1: why not support make `test_packet_redirect.cpp` parametrized so it can run in both "dataplane" modes? This is supported by the google test library. It would avoid a lot of duplication as far as I can tell.

targets/simple_switch/tests/pcap_mock/pcap_mock.hpp
antoninbas, line 1: The rest of the codebase uses the `.h` extension for headers unless my memory is failing me

targets/simple_switch/tests/pcap_mock/mock_libpcap.cpp
```diff
@@ -0,0 +1,299 @@
...
+    pcap_objects_by_name[name] = object;
+  }
+  static void remove_pcap(pcap* object) {
+    std::lock_guard lock(pcap_objects_by_name_mutex);
+    pcap_objects_by_name.erase(object->name_);
+  }
+
+  std::string name_;
+
+  bool is_activated_ = false;
```
antoninbas, line 166: maybe add a comment to explain this field

targets/simple_switch/tests/pcap_mock/mock_libpcap.cpp
```diff
@@ -0,0 +1,299 @@
...
+    return no_error; 
+  }
+
+  int activate() { 
+    PCAP_ERROR_IF_ACTIVATED; 
+    is_activated_ = true; 
+    return no_error; 
+  }
+
+  int	sendpacket(const u_char* buf, int size) {
```
antoninbas, line 104: double whitespace

configure.ac
```diff
@@ -324,6 +324,7 @@ AC_CONFIG_FILES([Makefile
 		targets/l2_switch/learn_client/Makefile
 		targets/simple_switch/Makefile
 		targets/simple_switch/tests/Makefile
+    	targets/simple_switch/tests/pcap_mock/Makefile
```
antoninbas, line 327: inconsistent indentation

antoninbas (commented):
biggest concern is that I am not convinced about the correctness of the change in `_bmi_port_interface_remove` (but I agree that the current code suffers from a deadlock).

PR #1307, opened 2025-06-01
explicitly cast uint to int to avoid error

grg:
Second half of PR to perform cast and compare of casted value. Previous commit performed the cast, but was still comparing against the non-cast value :-(


grg:
@fruffy Ugh, I was doing some testing of my PR and missed checking in one line.

PR #1309, opened 2025-06-10
Fix include issue for std::lower_bound in g++15

Hoooao:
The change fixes https://github.com/p4lang/p4c/actions/runs/15490297907/job/43614080729?pr=5310#step:5:3194 where std::lower_bound with compare functor cannot be found. 

In newer g++, the header file that defines std::lower_bound with a compare factor is only ported with `<algorithm>`, which was not included in `lpm_trie.h`. Previous versions of g++ seem to have the header file included with `<vector>`, thus compiled.

### DIFF
--- a/src/bm_sim/lpm_trie.h
+++ b/src/bm_sim/lpm_trie.h
@@ -18,11 +18,12 @@
 
 #include <bm/bm_sim/bytecontainer.h>
 
-#include <algorithm> 
+#include <algorithm>
 #include <cassert>
 #include <cstdint>
 #include <memory>
 #include <string>
+#include <utility>
 #include <vector>
 
 namespace bm {


src/bm_sim/lpm_trie.h
```diff
@@ -22,7 +22,7 @@
 #include <cstdint>
 #include <memory>
 #include <string>
-#include <utility>  // for std::swap
+#include <algorithm>
```
antoninbas, line 25: keep them in lexicographic order

Hoooao: done

matthewtlam (approved):

fruffy:
> can we update compiler versions in CI?
> 
> @fruffy

At least for the P4 compiler we use the oldest supported Ubuntu version as the lower bound for supported compiler versions. In this case this Ubuntu 22.04 with GCC 11 and clang 14 iirc. We can do the same thing here.

Hoooao:
@fruffy Sure, opened a PR here https://github.com/p4lang/p4c/pull/5319

fruffy (approved):
> @fruffy Sure, opened a PR here [p4lang/p4c#5319](https://github.com/p4lang/p4c/pull/5319)

Thanks! Looks good.

PR #1313, opened 2025-06-29
include libpcap when checking for pcap symbol

grg:
Link the pcap library when checking for `pcap_set_immediate_mode` function.

If the pcap library is not loaded, then the test for `pcap_set_immediate_mode` will always fail, and so `WITH_PCAP_FIX` will never be defined. This results in no packets being captured on the interface as it is not placed in promiscuous mode.


grg:
Yes. Without this fix, we were having local tests fail when using the cmake built version of BMv2.

I'll rebase, sign, and re-push.

fruffy (approved):

PR #1319, opened 2025-07-13
Clarify EventLogger purpose and document structured output for packet…

Sarthak-Developer-Coder:
… tracing  - Updated the class and file comments to emphasize EventLogger as the canonical   mechanism for structured, machine-readable packet trace logging (not just text logs). - Added guidance on extensibility for output formats (e.g., Protobuf, JSON). - Provided a virtual set_output_format() method as a documented extension point. - Improved comments to direct future packet trace features to use/extend EventLogger.

This commit updates the `bm::EventLogger` interface and documentation in `include/bm/bm_sim/event_logger.h` to clarify the intended use as the central, structured event logging facility for BMv2. The revised comments now explicitly state that the `EventLogger` should be used for generating machine-readable, structured event logs—including but not limited to packet traces—rather than relying on unstructured console output.

The new documentation encourages future developers and contributors to extend `EventLogger` when implementing advanced packet tracing or analysis capabilities, such as exporting packet events in formats like Protobuf or JSON. A virtual `set_output_format()` method is introduced as a documented extension point for alternative output formats. 

These changes guide the BMv2 community to leverage and evolve the structured event log infrastructure, promoting robust, parseable, and extensible logging for debugging, analysis, and automated testing, rather than ad-hoc console log parsing. This aligns with current project goals and ongoing discussions for improved packet trace extraction and analysis.


jafingerhut:
There is one CI test named "Test / test (clang++-8)" that is failing, which you can find a link to on the main page for this PR: https://github.com/p4lang/behavioral-model/pull/1319

Please fix those up, or ask if you are not able to determine what needs changing for that.

antoninbas (commented):
As I pointed out in a comment, I don't believe this is the right approach. In any case, taking this PR in isolation, I don't really see a justification for the changes as they are not currently leveraged. I think it would be a good idea to provide the skeleton for an alternative EventLogger output format (e.g., Protobuf) as part of the PR.

Issue #1323, opened 2025-08-23
simple_switch_grpc built with cmake fails some p4c tests that pass when it is built with GNU Autotools


jafingerhut:
All test results described were obtained on an aarch64 Ubuntu 24.04 Linux system.

When I build open source P4 dev tools with the script at [1] it uses GNU Autotools to build BMv2.

When I do it with script [2], everything is the same except that BMv2 is built with cmake.  I use these command line options with cmake:
```
cmake -DWITH_PI=on -DWITH_THRIFT=on ..
```

Then I run p4c tests in the p4c/build directory using first this command run as a non-root user:
```
make -j3 check
```
About 500 tests fail because I am not running as root.  I re-run those tests as root using:
```
sudo PATH=${PATH} VIRTUAL_ENV=${VIRTUAL_ENV} ${P4_EXTRA_SUDO_OPTS} make -j3 recheck
```

There are 5 tests that fail regardless of whether I build simple_switch_grpc with [1] or [2].  I won't list those here.  They've been failing for years with [1], and are not news.

Below is a list of p4c tests that pass when run with simple_switch_grpc built with [1] but fail when run with simple_switch_grpc built with [2].  For all of these, the reason for failure is that the test controller code tries to configure a clone session, with this error message in the error status returned by simple_switch_grpc:
```
Error when creating clone session in target
```
Here is the list of p4c tests that fail for this reason:
```
        4070 - testgen-p4c-bmv2-ptf/issue1001-1-bmv2.p4 (Failed)
        4071 - testgen-p4c-bmv2-ptf/issue1001-bmv2.p4 (Failed)
        4107 - testgen-p4c-bmv2-ptf/issue1642-bmv2.p4 (Failed)
        4108 - testgen-p4c-bmv2-ptf/issue1653-bmv2.p4 (Failed)
        4109 - testgen-p4c-bmv2-ptf/issue1653-complex-bmv2.p4 (Failed)
        4110 - testgen-p4c-bmv2-ptf/issue1660-bmv2.p4 (Failed)
        4217 - testgen-p4c-bmv2-ptf/issue383-bmv2.p4 (Failed)
        4245 - testgen-p4c-bmv2-ptf/issue562-bmv2.p4 (Failed)
        4366 - testgen-p4c-bmv2-ptf/pins_wbb.p4 (Failed)
        4371 - testgen-p4c-bmv2-ptf/bmv2_clone_preserving_field_list_eg.p4 (Failed)
[...]

PR #1324, opened 2025-08-26
set WITH_SIMPLE_SWITCH when compiling with targets

grg:
Attempt to address issue #1323 but setting WITH_SIMPLE_SWITCH when compiling with targets enabled.


jafingerhut:
When I tried building with these changes on my local system, I see a compile-time error that looks the same as the one in the failure for CI Test / test-ubuntu22 here: https://github.com/p4lang/behavioral-model/actions/runs/17248602562/job/48944687094?pr=1324

PR #1326, opened 2025-08-29
Fix #1325 - Enable ENABLE_MODULES if dlopen is found

### LINKED ISSUE #1325: simple_switch_grpc built with cmake fails some p4c tests that pass when it is built with GNU Autotools
After PR #1324 was merged, I find only two p4c tests that fail when BMv2 is built with cmake, that pass when BMv2 is built with GNU Autotools.  They are these:
```
255 - bmv2/testdata/p4_16_samples/extern-funcs-bmv2.p4 (Failed)
549 - bmv2/bmv2_emit_externs (Failed)
```
Here is the output of the first failing test, in case it gives anyone extra clues about the reason for the failure.
```
$ cd p4c

# commands to build p4c from source code go here

$ cd build

$ ./bmv2/testdata/p4_16_samples/extern-funcs-bmv2.p4.test -v
unrecognised option '--load-modules'
Target parser returned an error
Unknown primitive action: extern_func
ERROR: Test failed
```
It appears that perhaps `simple_switch_grpc` is being run with the command line option `--load-modules`, but it is not being recognized by the command line options parsing code.

jafingerhut:


jafingerhut:
@grg Note that I am a neophyte when it comes to editing CMakeLists.txt files, and it probably shows from this PR.  Feel free to suggest better ways to achieve this goal, or to create a separate PR that does it in a better way if that is easier for you.

grg:
Hi Andy, I've put up an alternative proposal in #1327.

The original autoconf enable-modules support was effectively a tri-state config: `--enable-modules` required support, `--disable-modules` disabled support, and not specifying anything would cause the system to check for `dlopen()` and enable support if found.

The new PR introduces a second cmake option to mimic this behavior.

PR #1327, opened 2025-09-04
Fix #1325 - Enable ENABLE_MODULES if dlopen is found

### LINKED ISSUE #1325: simple_switch_grpc built with cmake fails some p4c tests that pass when it is built with GNU Autotools

grg:
Alternative to #1326 by @jafingerhut 

This PR attempts to better mimic the autoconf build behavior. There, there are three configs: `--enable-modules`, `--disable-modules` (or `--enable-modules=no`), and nothing specified:
 * `--enable-modules`: dynamic loading via dlopen is required and it's an error if dlopen is missing.
 * `--disable-modules`: no dynamic loading.
 * nothing specified: support dynamic loading if dlopen is present, otherwise don't support dynamic loading.

To support this behavior, I had to introduce a second cmake option. So we now have `ENABLE_MODULES` and `REQUIRE_MODULES`. `ENABLE_MODULES` specifies that we want to support dynamic loading _if `dlopen()` is present on the system_; it is turned on by default. `REQUIRE_MODULES` indicates that dynamic loading support is _required_ and will fail if this is turned on but `dlopen()` is missing from the system; this is off by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,6 +167,17 @@ if(WITH_PI)
   endif()
 endif()
 
+# Check for dlopen
+check_include_file(dlfcn.h HAVE_DLFCN_H)
+if(HAVE_DLFCN_H)
+  include(CheckSymbolExists)
+  check_symbol_exists(dlopen "dlfcn.h" HAVE_DLOPEN)
+  message(STATUS "Found dlopen.  Enabling ENABLE_MODULES")
+  set(ENABLE_MODULES TRUE)
+else()
+  message(STATUS "Did not find dlopen")
+endif()
+
 if(ENABLE_MODULES)
   # Check for dlopen
   check_include_file(dlfcn.h HAVE_DLFCN_H)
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
These changes look good to me.  I have also tested them on a local system with various combinations of REQUIRE_MODULES and ENABLE_MODULES on/off, as well as a small tweak that forced HAVE_DLOPEN to be off, and all combinations functioned as desired.

PR #1329, opened 2025-10-05
Selector Packet Fanout  -  Core implementation

Hoooao:
This PR is derived from #1316 .
The testcases will be in another PR.

Main updates after #1316 :
1. Added command-line option `--enable-pkt-fanout` to `configure`, which defines `PKT_FANOUT_ON`. Is not supported in cmake for now. 
2. Raise an exception when `selector_fanout` mode is used while the above option is not given.
3. Set the `grp_selector` for all specified selectors during switch `start_and_return_`, less intrusive.
4. Previously, replicated pkts are pushed to ingress/egress-buffers in the `ingress/egree_thread` after the original packet gets processed. Now the logic is colocated with packet replication, which is much cleaner.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,6 +63,15 @@ AS_IF([test "x$enable_debugger" = "xyes"], [
     ])
 ])
 
+# TODO: add this to cmake build as well
+pkt_fanout_enabled=no
+AC_ARG_ENABLE([pkt_fanout],
+    AS_HELP_STRING([--enable-pkt-fanout], [Enable packet fanout support]))
+AS_IF([test "x$enable_pkt_fanout" = "xyes"], [
+    pkt_fanout_enabled=yes
+    AC_DEFINE([PKT_FANOUT_ON], [], [Enable packet fanout support])
+])
+
 logging_macros_enabled=no
 AC_ARG_ENABLE([logging_macros],
     AS_HELP_STRING([--disable-logging-macros],
--- a/include/bm/bm_sim/fanout_pkt_mgr.h
+++ b/include/bm/bm_sim/fanout_pkt_mgr.h
@@ -1,4 +1,4 @@
-/* Copyright 2013-present Contributors to the P4 Project
+/* Copyright 2025-present Contributors to the P4 Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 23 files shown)


jafingerhut:
@jonathan-dilorenzo FYI it would be good for someone in Google who wants this feature to learn enough about the code base to give it a good review.  Unless you employ someone who knows this feature and how it works, or can contract with someone who does, any bugs discovered in it later are unlikely to be fixed by someone who is not interested in the feature.

configure.ac
```diff
@@ -63,6 +63,15 @@ AS_IF([test "x$enable_debugger" = "xyes"], [
     ])
 ])
 
+# TODO: add this to cmake build as well
```
matthewtlam: Do we still need this comment? I think you already added it to cmake build

include/bm/bm_sim/fanout_pkt_mgr.h
```diff
@@ -0,0 +1,112 @@
...
+        ap->set_group_selector(grp_selector);
+      }
+    }
+    inline void register_thread(std::thread::id thread_id,
+      const std::function<void(const bm::Packet *)> &buffer_push_fn) {
+      BMLOG_DEBUG("Registering thread {}", thread_id);
+      fanout_ctx_map.emplace(thread_id, FanoutCtx(buffer_push_fn));
+    }
+
+    // TODO(Hao): deduplicate packets fanout, optional
```
matthewtlam: Do we still need this comment?

include/bm/bm_sim/fanout_pkt_mgr.h
```diff
@@ -0,0 +1,112 @@
...
+
+    mbr_hdl_t get_from_hash(grp_hdl_t grp, hash_t h) const override;
+
+    void reset() override {}
+
+ private:
+    std::unordered_map<grp_hdl_t, std::vector<mbr_hdl_t>> groups;
+};
+
+class FanoutPktMgr {
```
matthewtlam: Let's add some documentation regarding the FanoutPktMgr. For example how it is used and use cases

targets/simple_switch/simple_switch.cpp
```diff
@@ -506,20 +538,25 @@ SimpleSwitch::ingress_thread() {
...
-      phv->get_field("standard_metadata.parser_error").set(
-          packet->get_error_code().get());
-    }
 
-    if (phv->has_field("standard_metadata.checksum_error")) {
-      phv->get_field("standard_metadata.checksum_error").set(
-           packet->get_checksum_error() ? 1 : 0);
-    }
+    // Check if the packet has an optional continue node for pkt fanout
+    // TODO(Hao): update the doc/simple_switch.md
```
matthewtlam: Do we still need this TODO?

matthewtlam (commented):
Leaving behind first set of comments. Will add more later today

PR #1330, opened 2025-10-07
Updating Parser to work with GTP-U 

trvon:
In one of my research experiments, recirculated GTP-U packets triggered the PacketTooShort error. I assumed this was because the parser’s bounds were still compared against Packet::get_ingress_length(), even after the INT shim/BeamSec insertion increased the buffer size. During the second-pass parsing, this stale length caused the deparser to emit only the original payload, resulting in the omission of telemetry headers.

To address this issue, I updated all parser bounds checks (VL extracts, fixed extracts, shift/advance operations) to rely on Packet::get_data_size(), which provides the current length of the live buffer. This change ensures that the parser logic accurately reflects the real payload and only raises an error when the buffer itself is insufficient in size. I also  added unit tests that simulate ingress_length < actual buffer.

### DIFF
--- a/src/bm_sim/parser.cpp
+++ b/src/bm_sim/parser.cpp
@@ -267,7 +267,8 @@ void extract_VL(Header *hdr,
   // get_nbytes_packet counts the VL field in the header as 0 bits
   auto bytes_to_extract = static_cast<size_t>(
       hdr->get_nbytes_packet() + nbits / 8);
-  if (pkt->get_data_size() - *bytes_parsed < bytes_to_extract)
+  if (*bytes_parsed > pkt->get_data_size() ||
+      pkt->get_data_size() - *bytes_parsed < bytes_to_extract)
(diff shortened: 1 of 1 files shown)


antoninbas (commented):
It's not clear to me what the issue is based on this PR description.
As far as I can tell, the `ingress_length` is set correctly (based on what was produced by the deparser) for recirculated packets: https://github.com/p4lang/behavioral-model/blob/68f4a978f465fd76e98fcdecb762981843fb7310/targets/simple_switch/simple_switch.cpp#L762

It would be helpful to clarify this, as well as provide a P4 program to reproduce the issue. Adding @jafingerhut in case he has some time to help reproduce the issue / review the patch.

trvon:
I do not have a clean way to share my p4 code, but I will reproduce the issue I saw with a smaller p4 program and ptf test.

PR #1339, opened 2026-02-02
Fix entry addition logic for keyless match tables

Shivampal157:
### Summary
Fixes entry addition logic for keyless match tables.

Previously, adding an entry to a table with no match key (get_size_key() == 0)
would always fail with NO_TABLE_KEY, even when the provided match_key was empty.

### What changed
- NO_TABLE_KEY is now returned *only when*:
  - the table has no match key, *and*
  - a non-empty match_key is provided
- Keyless tables with a single default action can successfully add entries again.

### Files changed
- src/bm_sim/match_units.cpp

### Motivation
This restores correct behavior for keyless tables such as get_lb_key,
which rely on a default action without match keys.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,38 @@
+{
+    "configurations": [
+        {
+            "name": "Win32",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "${workspaceFolder}/include",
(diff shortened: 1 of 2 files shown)


jafingerhut:
Closing this, as for reasons I mentioned in an earlier comment I believe this is a change we do not want to make.

PR #1357, opened 2026-03-20
feat: add configurable log rotation options (fixes #930)

### LINKED ISSUE #930: Support configuring log rotation for bmv2 switch logs
When I activate the logger in bmv2 software switches, the system only keeps  last 4 log files. But I want to keep more than 4 log files. 

Can someone kindly help me about,  How to configure (where is the configuration) the system to increase the number of log files.

Vineet1101:
Add --log-file-max-size and --log-file-max-files CLI options to allow users to configure the maximum log file size before rotation and the number of rotated backup files to keep.

Previously these values were hardcoded to 5MB and 3 files respectively. The new options default to the same values for backward compatibility.

fixes #930

### DIFF
--- a/include/bm/bm_sim/logger.h
+++ b/include/bm/bm_sim/logger.h
@@ -89,9 +89,13 @@ class Logger {
   static void set_logger_console();
 
   //! Log all messages to the given file. If \p force_flush is true, the logger
-  //! will flush to disk after every log message.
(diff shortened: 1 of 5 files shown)


Vineet1101:
@antoninbas can you please review it

antoninbas (approved):

PR #1358, opened 2026-03-20
Feature/configurable pre limits

Vineet1101:
## Changes

### Core Library (`bm_sim`)
- **[McSimplePre](cci:1://file:///home/p4/src/behavioral-model/include/bm/bm_sim/simple_pre.h:136:2-137:49)**: Converted `MGID_TABLE_SIZE`, `L1_MAX_ENTRIES`, and `L2_MAX_ENTRIES` to non-static instance members. Added a constructor to accept these values, using the original values as defaults.
- **[McSimplePreLAG](cci:1://file:///home/p4/src/behavioral-model/include/bm/bm_sim/simple_pre_lag.h:41:2-45:71)**: Updated to forward configuration parameters to the base [McSimplePre](cci:1://file:///home/p4/src/behavioral-model/include/bm/bm_sim/simple_pre.h:136:2-137:49) class.
- **Capacity Checks**: Updated internal logic in both classes to use the configurable instance members instead of the old static constants.

### Switch Targets
- **`simple_switch`**: Added 3 new command-line options and threaded them through the [SimpleSwitch](cci:1://file:///home/p4/src/behavioral-model/targets/simple_switch/simple_switch.h:138:2-138:46) constructor to the PRE.
- **`simple_switch_grpc`**: Added the same 3 command-line options and updated the [SimpleSwitchGrpcRunner](cci:1://file:///home/p4/src/behavioral-model/targets/simple_switch_grpc/switch_runner.cpp:496:0-517:1) to pass them to the underlying switch instance.

### New CLI Options
Both `simple_switch` and `simple_switch_grpc` now support:
- `--max-mc-groups <uint>`: Maximum number of multicast groups (default: 4096)
- `--max-l1-entries <uint>`: Maximum number of L1 multicast entries (default: 4096)
- `--max-l2-entries <uint>`: Maximum number of L2 multicast entries (default: 8192)

## Verification
- **Unit Tests**: Added 3 new test cases to [tests/test_pre.cpp](cci:7://file:///home/p4/src/behavioral-model/tests/test_pre.cpp:0:0-0:0):
    - `McSimplePre.ConfigurableLimits`: Verifies `TABLE_FULL` is correctly triggered when custom small limits are exceeded.
    - `McSimplePre.DefaultLimits`: Verifies that the default constructor still uses the original hardcoded limits.
    - `McSimplePreLAG.ConfigurableLimits`: Verifies the same behavior for the LAG-enabled PRE variant.
- All 9 PRE unit tests (existing + new) pass successfully.
- Build verified for both `simple_switch` and `simple_switch_grpc` targets.

## Backward Compatibility
The original hardcoded values are preserved as default constructor arguments and default CLI option values. No existing P4 programs or switch configurations are affected by these changes.

Fixes #1342

### DIFF
--- a/targets/simple_switch/main.cpp
+++ b/targets/simple_switch/main.cpp
@@ -22,6 +22,8 @@
 
 #include <bm/config.h>
 
+#include <climits>
+
 #include <bm/SimpleSwitch.h>
 #include <bm/bm_runtime/bm_runtime.h>
 #include <bm/bm_sim/options_parse.h>
@@ -95,6 +97,11 @@ main(int argc, char* argv[]) {
       mgid_table_size = SimpleSwitch::default_mgid_table_size;
     else if (rc != bm::TargetParserBasic::ReturnCode::SUCCESS)
       std::exit(1);
+    if (mgid_table_size == 0 || mgid_table_size > INT_MAX) {
+      std::cerr << "max-mc-groups must be between 1 and "
+                << INT_MAX << std::endl;
+      std::exit(1);
+    }
   }
 
   uint32_t l1_max_entries = 0xffffffff;
@@ -105,6 +112,11 @@ main(int argc, char* argv[]) {
       l1_max_entries = SimpleSwitch::default_l1_max_entries;
     else if (rc != bm::TargetParserBasic::ReturnCode::SUCCESS)
       std::exit(1);
+    if (l1_max_entries == 0 || l1_max_entries > INT_MAX) {
+      std::cerr << "max-l1-entries must be between 1 and "
+                << INT_MAX << std::endl;
+      std::exit(1);
+    }
   }
 
   uint32_t l2_max_entries = 0xffffffff;
@@ -115,6 +127,11 @@ main(int argc, char* argv[]) {
       l2_max_entries = SimpleSwitch::default_l2_max_entries;
     else if (rc != bm::TargetParserBasic::ReturnCode::SUCCESS)
       std::exit(1);
+    if (l2_max_entries == 0 || l2_max_entries > INT_MAX) {
+      std::cerr << "max-l2-entries must be between 1 and "
+                << INT_MAX << std::endl;
+      std::exit(1);
+    }
   }
 
   simple_switch = new SimpleSwitch(enable_swap_flag, drop_port,
--- a/targets/simple_switch/main.cpp
+++ b/targets/simple_switch/main.cpp
@@ -28,6 +28,7 @@
 #include <bm/bm_sim/target_parser.h>
 
 #include <climits>
+#include <string>
 
 #include "simple_switch.h"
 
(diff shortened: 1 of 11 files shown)


Vineet1101:
@jafingerhut can you review it

jafingerhut:
> @jafingerhut can you review it

I will.  First, I'd like to know if you have tested the changes yourself?  i.e. bulit the code locally and verified that with the new command line options it actually enables enforcing the limits specified on the command line?

I do not see any range-checking on the values provided on the command line.  Should there be?

e.g. negative values don't make any sense, but maybe the function you are using to parse the option values only support non-negative values?

What if the user specifies a value like 4294967295 (2^32-1) on the command line.  Should that give an error message as not supportable?  I do not know what good maximum values to support are yet.

Vineet1101:
@jafingerhut  I have added explicit bounds checking for max-mc-groups, max-l1-entries, and max-l2-entries in main.cpp
 for both simple_switch and simple_switch_grpc.

The values are now strictly validated to be between 1 and INT_MAX (2,147,483,647). If a user provides 0, a negative number (which parses as > INT_MAX), or any other value exceeding INT_MAX, the executable will catch it, print a clear error to stderr, and exit gracefully. 

I've verified this behavior manually and ensured all existing unit tests continue to pass.

targets/simple_switch/main.cpp
```diff
@@ -49,6 +51,15 @@ main(int argc, char* argv[]) {
   simple_switch_parser.add_uint_option(
       "priority-queues",
       "Number of priority queues (default is 1)");
+  simple_switch_parser.add_uint_option(
+      "max-mc-groups",
+      "Maximum number of multicast groups (default is 4096)");
```
jafingerhut: Sorry for more nitpick questions, but if we can print in these messages the value of McSimplePre::DEFAULT_MGID_TABLE_SIZE and other similar constants, then we avoid repeating the value of the constant in two places like it is now.  e.g. if someone updates the default value of the constant in the future, and we use the value of the constant in printing this message, they do not need to remember to update the value in two places.

targets/simple_switch_grpc/main.cpp
```diff
@@ -81,6 +82,15 @@ main(int argc, char* argv[]) {
   simple_switch_parser.add_uint_option(
     "priority-queues",
     "Number of priority queues (default is 1)");
+  simple_switch_parser.add_uint_option(
+      "max-mc-groups",
+      "Maximum number of multicast groups (default is 4096)");
```
jafingerhut: Same comment here as above, about using the value of the declared constant instead of the literal 4096 here.

targets/simple_switch/main.cpp
Vineet1101: yup, you are right. I have replaced hardcoded values with variables

jafingerhut (approved):
LGTM

PR #1361, opened 2026-03-29
Switch to pynng instead of nnpy. 

fruffy:
nnpy has been deprecated four years ago and shows signs of bit-rotting. Installing it on newer distributions seems to cause issues. This is a draft to replace nnpy with pynng.

The reason nanomsg is very useful is because it allows us to model network switch behavior without requiring veths. And without veths we do not need sudo. When working with agents, a tight feedback loop can be useful, which includes running packet tests. sudo requirements can get in the way of that.

Related PRs: #https://github.com/p4lang/ptf/pull/234 https://github.com/p4lang/p4c/pull/5576


jafingerhut (approved):
LGTM

jafingerhut:
Should we be installing nng [1] instead of the current nanomsg, too?

[1] https://github.com/nanomsg/nng

fruffy:
> Should we be installing nng [1] instead of the current nanomsg, too?
> 
> [1] https://github.com/nanomsg/nng

It's a pretty large change and the Python dep is meant to be backwards compatible. Can do it in a separate PR.

PR #1364, opened 2026-04-06
Replace nanomsg with nng

jafingerhut:


jafingerhut:
This is a draft PR that I may never complete.  I started it hoping that the replacement of the nanomsg library with nng would be a drop in replacment, but from what I recall in the way I left this, it appeared that there are differences in the APIs of the two libraries that would require some changes to behavioral-model code in order to use the newer library, and I did not have the time/patience to work through that.  Leaving this open for now in case someone else is interested in picking up that code updating task for themselves, but certainly OK to close this PR if it feels too much like clutter.

PR #1367, opened 2026-04-08
Fix Thrift support for simple_switch_grpc (#1366)

### LINKED ISSUE #1366: simple_switch_grpc built without Thrift support even with -DWITH_THRIFT=ON
## Problem

When building with `-DWITH_THRIFT=on` using cmake, the resulting `simple_switch_grpc` binary shows the warning:

```
You used the '--thrift-port' command-line option, but this target was compiled without Thrift support.
```

This happens even though:
- cmake is run with `-DWITH_THRIFT=on`
- Thrift libraries are installed on the system

## Root Cause

In `targets/simple_switch_grpc/CMakeLists.txt`, the `WITH_THRIFT` compile definition is not passed to the simple_switch_grpc target.

The Thrift code in `switch_runner.cpp` is guarded by `#ifdef WITH_THRIFT`:

```cpp
#ifdef WITH_THRIFT
  bm_runtime::start_server(simple_switch.get(), thrift_port);
  // Thrift service setup
#else
  if (parser.option_was_provided("thrift-port")) {
    bm::Logger::get()->warn("compiled without Thrift support...");
  }
#endif
```

While `-DTHRIFT_ON` is added at the top-level CMakeLists.txt (not to be confused with `-DWITH_THRIFT`), this definition is not propagated to the simple_switch_grpc target's compilation.

## Fix

Add `add_definitions(-DWITH_THRIFT)` in the simple_switch_grpc CMakeLists.txt when `WITH_THRIFT` is enabled:

```cmake
if(WITH_THRIFT)
  add_definitions(-DWITH_THRIFT)
  target_link_libraries(simple_switch_grpc PRIVATE
          thrift
          simpleswitch_thrift
  )
endif()
```

This ensures:
1. The preprocessor macro is defined during compilation, so Thrift code is included in the binary
2. Both the Thrift shared library (`libthrift.so`) and the generated static library (`libsimpleswitch_thrift.a`) are linked

## Testing

- Build behavioral-model with cmake flags: `-DWITH_PI=on -DWITH_THRIFT=on`
- Build the simple_switch_grpc target
- Run simple_switch_grpc with `--thrift-port 9091`
- Verify no warning about missing Thrift support appears
- Check port is listening: `netstat -an | grep 9091`

davideandres95:
Fixes #1366, ensuring:
1. The preprocessor macro is defined during compilation, so Thrift code is included in the binary
2. Both the Thrift shared library (`libthrift.so`) and the generated static library (`libsimpleswitch_thrift.a`) are linked

### DIFF
--- a/targets/simple_switch_grpc/CMakeLists.txt
+++ b/targets/simple_switch_grpc/CMakeLists.txt
@@ -59,6 +59,14 @@ target_link_libraries(simple_switch_grpc PRIVATE
   ${GRPCPP_LIBRARIES}
 )
 
+if(WITH_THRIFT)
+  add_definitions(-DWITH_THRIFT)
+  target_link_libraries(simple_switch_grpc PRIVATE
+         thrift
+         simpleswitch_thrift
+  )
+endif()
+
 # We follow this tutorial to link with grpc++_reflection:
 # https://github.com/grpc/grpc/blob/master/doc/server_reflection_tutorial.md
 #target_link_options(simple_switch_grpc PRIVATE


targets/simple_switch_grpc/CMakeLists.txt
```diff
@@ -59,6 +59,14 @@ target_link_libraries(simple_switch_grpc PRIVATE
   ${GRPCPP_LIBRARIES}
 )
 
+if(WITH_THRIFT)
```
jafingerhut, line 62: There is an earlier block of code `if(WITH_THRIFT)` earlier in this file that does some of this.  Perhaps add the new parts there, instead of a seaprate section here?

jafingerhut: I am editing this comment about enabling WITH_THRIFT by default.  This is already done in the top level CMakeLists.txt file of this repo.  Sorry for the noise.

jafingerhut:
Please fix DCO signing problem.  On the page for this PR https://github.com/p4lang/behavioral-model/pull/1367 look in the section "Some checks were not successful" for "DCO", and click on that link for instructions on how to modify this PR so that it is signed properly.  If that does not work for some reason, creating a new PR where you are careful to use `git commit -s` to make all of your commits will also work.

jafingerhut (approved):
I tested a build with and without these changes, and they do fix the stated problem.  Thanks!

PR #1371, opened 2026-04-17
bugfix: "make distclean"

chreekat:
Before this change, third_party shows up in two different Makefiles, and distclean tries to clean it twice, which failed on the second try.

I hit this while building with debian packaging tools, which runs a distclean at some point.

The fix is to remove the duplicate build of third_party, though that does mean one must be careful with build order.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,7 +20,9 @@ if COND_PI
     MAYBE_PI = PI services
 endif
 
-# simple_switch depends on libbmpi so PI needs to appear first
+# The targets directory (simple_switch, etc.) depends on libbmpi (from PI) and
+# gtest (from third_party), so those must appear before MAYBE_TARGETS in this
+# list.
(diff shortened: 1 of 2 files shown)


chreekat:
I suppose another fix would be to make the distclean for third_party idempotent, now that I think about it. Still, having third_party triggered twice seems unnecessary.

chreekat:
Well, that's an interesting error. Looks new, but I don't see how it's related to my change. I sort of manually traced through the Docker images and it does look like make should be there. Not sure what's wrong. Any idea?

fruffy:
Are you able to replay this locally? 

I think we should not depend on docker for the test runs. Opened a PR to clean this up: 
https://github.com/p4lang/behavioral-model/pull/1373

fruffy (approved):

chreekat:
Yeah, I specifically ran into this while building packages with p4lang-packages. I guess that's not an easy repro. :)

PR #1373, opened 2026-04-19
Refresh behavioral-model CI

fruffy:
- Do not depend on the docker image for testing.
- Add support for Ubuntu24 builds, fix new compilation errors.
- Add support for ccache for faster builds.
- Simplify build matrix. 

Once we deprecate autotools in favour of CMake, the build matrix should become smaller again.


PI/src/pi_tables_imp.cpp
```diff
@@ -648,23 +648,23 @@ bm::MatchTableAbstract::EntryCommon get_match_entry(
           0, t_name, handle, &entry);
       if (error_code != bm::MatchErrorCode::SUCCESS)
         throw bm_exception(error_code);
-      return std::move(entry);  // avoid copy
+      return entry;
```
grg, line 651: What problems was the `std::move()` causing? Asking for my own education, not because I'm saying it should be present.

targets/simple_switch_grpc/tests/example.cpp
```diff
@@ -98,7 +98,8 @@ test() {
     auto status = pi_stub_->SetForwardingPipelineConfig(
         &context, request, &rep);
     assert(status.ok());
-    config->release_p4info();
+    auto *released_p4info = config->release_p4info();
+    assert(released_p4info == &p4info);
```
grg, line 102: Should this and the other assert calls that were added actually be invoking the ASSERT_EQ gtest macro? Would require including either base_test.h or gtest.h.

grg (approved):
Looks good in general. One query about `assert()` vs gtest `ASSERT/ASSERT_EQ` in tests -- happy for you to move forward either way.

PI/src/pi_tables_imp.cpp
fruffy: Afaik this was a case for RVO (https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization). The std::move might actually prevent the optimization.

targets/simple_switch_grpc/tests/example.cpp
fruffy: The entire class uses `assert`. Might probably make sense to do this alltogether in a separate PR. 

These lines were just added because `release_p4info()` was marked `[[nodiscard]]` and not checked.

fruffy: https://github.com/p4lang/behavioral-model/issues/1374

targets/simple_switch_grpc/configure.ac
```diff
@@ -6,6 +6,30 @@ AC_INIT([simple-switch-grpc], [0.1], [antonin@barefootnetworks.com])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 AC_CONFIG_HEADERS([config.h])
 
+AH_TOP([
+/* Copyright 2013-present Barefoot Networks, Inc.
```
chreekat, line 10: Is this still accurate?

jafingerhut: The copyright notice on generated Makefiles, you mean?  It seems like a reasonable copyright notice for the file configure.ac itself, as it was first added to this repo when Antonin Bas was working for Barefoot Networks.

I am not yet aware of whether a copyright notice in generated Makefiles is very significant, but if it is, generating the same one as from the `.ac` file it was generated from sounds like a reasonable option.

jafingerhut: FYI, here is one article, not legally binding nor necessarily the most recent advice from the Linux Foundation, but good food for thought, on updating the copyright year on every commit to a file that is in a new calendar year (TLDR it recommends using the year of first adding the file, and then never change it): https://liferay.dev/b/how-and-why-to-properly-write-copyright-statements-in-your-code

chreekat: That all makes sense. Nothing worth spending more time on, I think. :)

PR #1375, opened 2026-04-22
Part 1 of adding REUSE-lintable copyright and license info to files

jafingerhut:


jafingerhut:
See this comment on a recent similar p4-spec repository PR that explains the rationale for REUSE, and some of its mechanics: https://github.com/p4lang/p4-spec/pull/1395#issuecomment-4286250882

jafingerhut:
This is ready for review.  I realize it is not of a size convenient for review, so here is how these changes were created, if you trust me.  I was pretty careful.

Every file in third_party I went through and checked who the copyright holder was, and what license was in the file, if any.  I kept those licenses as they were.  They were all things compatible with Apache-2.0, such as MIT, BSD-2-Clause, BSD-3-Clause, and one custom license that was similar to BSD-2-Clause, but not the same, so I created a custom license file for it in LICENSES/ directory.

For all files that already had an Apache-2.0 license explicitly mentioned in the file, I was sure to keep those files as Apache-2.0.  This was most of the C/C++ source files in the repo.  I maintained their copyright holders.

The same goes for the remaining few C/C++ source files that did not have an explicit Apache-2.0 license.  I used the Apache-2.0 license for them.  There were only a few of them, and they did not mention any other license explicitly in their comments, and they were typically added in the same commits as other files with an Apache-2.0 license, and/or by authors working for Barefoot Networks at the time of adding the file.

For infrastructure files, e.g. yml, sh, CMakeLists.txt etc. I used Apache-2.0, and maintained the copyright holder and year in the file if there was one, otherwise I used the year the file was first added to the repo as the copyright year, and the name of the first commit Author as the copyright holder, unless I knew they were working for Barefoot Networks at the time, in which case I used Barefoot Networks.

No proprietary/confidential files were found.

There are a handful of files with suffix `.m4` for Automake that are GPL, but they have an explicit exception in their licnese that allows the output generated from them to be used for non-GPL'd projects.

fruffy (approved):

PR #1376, opened 2026-04-23
update googletest to 1.17.0/add googlemock

grg:
Update the gtest files in third_party and add gmock files. Both come from v1.17.0 of googletest.


grg:
Updated gtest to 1.17.0 (current release version).

Added googlemock from the googletest repo as I would like to be able to use `EXPECT_THAT` and `::testing::HasSubstr` to match against some warning strings I expect the model to produce.
```
TEST(MyTest, StringContainsSubstring) {
    std::string result = "Hello, World!";
    EXPECT_THAT(result, ::testing::HasSubstr("World"));
}
```

grg:
Eventually it would be nice to handle googletest the same way we do in p4c, but that's hard to do while we're still supporting autotools 🙁

fruffy:
> Eventually it would be nice to handle googletest the same way we do in p4c, but that's hard to do while we're still supporting autotools 🙁

Soon I hope :)

fruffy (approved):

Issue #1370, opened 2026-04-15
Reads of fields from invalid headers: additional options for return value


grg:
The P4 language specification (v1.2.5, §8.25) states:

> The result of reading a value in any of the situations below is that some unspecified value will be used for that field.
> * reading a field from a header that is currently invalid.
> * ...

The behavioral model currently returns that last written value when reading a field from an invalid header.

I propose adding two additional modes when reading from invalid headers:

1. Return 0: All reads of fields in invalid headers return 0. (Likely mirrors some hardware implementations.)
2. Return the last written value and **print a warning**.

I suggest that the choice of mode is a compile-time setting (e.g., via a cmake parameter).

grg:
@fruffy / @jafingerhut Any objections to me adding these additional modes for invalid field reads?

jafingerhut:
This sounds like a useful choice to have, and given that #2 in your list behaves the same way for packet processing behavior as the current code does, only generating additional warnings, that seems as backwards compatible as anyone would need.

fruffy:
Yes, sounds reasonable. It makes sense to keep 2 the default since it puts emphasis that this is undefined behavior.

grg:
I've opened a PR (#1377). I ended up making the warning/zero-return configurable via command-line parameters instead of compile-time defines. This seemed like the better approach for testing 🙂

jafingerhut:
@grg Do you think this issue should be closed now?  Or perhaps there are additional features that PR #1377 does not implement that you think are worth doing?

PR #1377, opened 2026-04-24
option to warn/return 0 on invalid field read

grg:
Add configurable options when reading an invalid field (currently defined as reading a field in a header that is not valid):
* Report a warning.
* Return zero.

Each of these behaviors can be enabled independently. The default behavior is to maintain the existing behavior (no warning, return the last written value, even if the header is invalid.)

Addresses #1370.

### DIFF
--- a/include/bm/bm_sim/data.h
+++ b/include/bm/bm_sim/data.h
@@ -460,8 +460,7 @@ class Data {
   //! Could have named it get_value(), but wanted to avoid accidental
   //! modification of the value when someone intended to call the const
   //! version of get_value()
-  Bignum &get_nc_value()
-  {
+  Bignum &get_nc_value() {
     return value;
   }
 
--- a/include/bm/bm_sim/data.h
+++ b/include/bm/bm_sim/data.h
@@ -446,8 +446,31 @@ class Data {
   Data &operator=(Data &&other) = default;
 
  protected:
-  Bignum value{0};
   bool arith{true};
+
+  //! Get the internal value as a constant Bignum
+  const Bignum &get_value() const {
+    return value;
+  }
+
+  //! Get the internal value as a Bignum that can be modified.
+  //! (nc = non-const)
+  //! Could have named it get_value(), but wanted to avoid accidental
+  //! modification of the value when someone intended to call the const
+  //! version of get_value()
+  Bignum &get_nc_value()
+  {
+    return value;
+  }
+
+  //! Set the internal value from a Bignum. This is used
+  //! by derived classes that need to modify the value.
+  void set_value(const Bignum &v) {
+    value = v;
+  }
+
+ private:
+  Bignum value{0};
 };
 
 }  // namespace bm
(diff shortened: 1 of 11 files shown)


grg:
Would appreciate feedback in a couple of areas in particular:
1. The `Data::value` is now a private member instead of a protected one. Most accesses are now done via a _virtual_ accessor, so most reads pay a vtable lookup penalty. Is there a better way to do this? (Considered a PIMPL approach, but seemed uglier.)
2. Suggestions for making the new test more compact.

grg:
Looks like I've got some fixes to make to the old build system 🙁

jafingerhut:
On the performance question, hopefully this weekend some time I will get some time to try out locally building bmv2 with these changes, and running them through the p4c test suite, and measure performance with and without these changes on the same tests.  I'll bet $10 that it will be difficult to measure any performance difference.

Note to self: Ensure that bmv2 with and without these changes are built with any debug/logging options turned off.

jafingerhut:
Minor question/suggestion: If this is implementing warning/force-0-value on reading a field in a header that is currently invalid, it might be better to call the options and internal variables tracking the values of the options something like "warn_on_accessing_field_in_invalid_header" or something similar.

Reason: I can imagine someone perhaps in the future wanting to go even further than what you are doing here, and tracking whether a field has actually ever been assigned to, and if not, issuing a warning if it is read before being initialized (even when the header the field is in is currently valid).

jafingerhut:
I have no suggestions for modifying the unit tests you have written.  They have the benefit of being clear and straightforward.  No need that I can see to make them more compact.

grg:
> On the performance question, hopefully this weekend some time I will get some time to try out locally building bmv2 with these changes, and running them through the p4c test suite, and measure performance with and without these changes on the same tests. I'll bet $10 that it will be difficult to measure any performance difference.
> 
> Note to self: Ensure that bmv2 with and without these changes are built with any debug/logging options turned off.

I also don't expect a huge impact. I'm mostly highlighting it for completeness 🙂 
I don't think we've ever advertised bmv2 as a performance-optimized implementation. Since it's a reference implementation, I think it's better to lean towards features that aid in verification and debugability over performance.

grg:
> Minor question/suggestion: If this is implementing warning/force-0-value on reading a field in a header that is currently invalid, it might be better to call the options and internal variables tracking the values of the options something like "warn_on_accessing_field_in_invalid_header" or something similar.
> 
> Reason: I can imagine someone perhaps in the future wanting to go even further than what you are doing here, and tracking whether a field has actually ever been assigned to, and if not, issuing a warning if it is read before being initialized (even when the header the field is in is currently valid).

I included a FIXME comment as a suggestion for future enhancement here: https://github.com/p4lang/behavioral-model/pull/1377/changes#diff-abec37663881efd42bbc504a119d447e9df2898b11ac6c35b944103cf3a3a34dR156-R157

But I will rename the variables to better reflect their current behavior.

jafingerhut (approved):
LGTM.  @fruffy let me know if you want to review this before we merge it, and if so, whether you want me to run a comparative before vs. after performance test to see what the effect of this, but personally my best guess is the performance difference should be difficult to measure.

jafingerhut:
Going aheaed and merging this in.  Thanks for the enhancement!

Issue #1360, opened 2026-03-28
Remove boost dependencies.


fruffy:
We can likely clean up some dependencies on boost since we have upgraded to C++17. Two easy low-hanging fruits are boost::filesystem and boost::optional. 

There is also (as suggested by GPT5.3): 

- Easy win: Boost::thread  
  - You’re using Boost shared mutex/locks in core paths (include/bm/bm_sim/*.h, src/bm_sim/*.cpp), e.g. boost::shared_mutex, boost::shared_lock, boost::unique_lock.  
  - In C++17 this can be replaced with std::shared_mutex, std::shared_lock, std::unique_lock (<shared_mutex>, <mutex>).  
  - That should let you remove Boost::thread from CMake (src/bm_sim/CMakeLists.txt:80) and related autotools checks (configure.ac:251).
- Medium effort: Boost hash helpers (boost::hash_combine)  
  - Used in a few places (include/bm/bm_sim/P4Objects.h:442, src/bm_sim/debugger.cpp:174, include/bm/bm_sim/field_lists.h:125, etc.).  
  - C++17 has no std::hash_combine, so you’d replace with a small local helper (common combine formula) and drop <boost/functional/hash.hpp> includes.

The last remaining piece is boost::program_options. I have great experience with CLI11: https://github.com/CLIUtils/CLI11

It's a standalone dependency and can be vendored, which makes it easier to work with. 

@matthewtlam Could this be in scope for the GSoC modernization task you have in mind? Otherwise I kick this off myself.

c8ef:
Hi @fruffy!

I'm one of the applicants for the GSoC project on BMv2. While my proposal outlines four potential approaches to codebase modernization, I hadn't considered modernizing the Boost dependencies - though I believe it fits well within the project's scope. Since I have availability now, I'm wondering whether we should wait until the project is officially announced (April 30) or if we can get started right away.

fruffy:
> Hi [@fruffy](https://github.com/fruffy)!
> 
> I'm one of the applicants for the GSoC project on BMv2. While my proposal outlines four potential approaches to codebase modernization, I hadn't considered modernizing the Boost dependencies - though I believe it fits well within the project's scope. Since I have availability now, I'm wondering whether we should wait until the project is officially announced (April 30) or if we can get started right away.

We can wait until we officially announce the project. @matthewtlam @qobilidop any thoughts?

qobilidop:
[...]

PR #1383, opened 2026-05-07
Migrate from `{boost => std}::{shared_mutex, shared_lock, unique_lock}`

c8ef:
Part of #1360.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -13,8 +13,6 @@ sudo apt-get install -y \
     g++ \
     libboost-dev \
     libboost-program-options-dev \
-    libboost-system-dev \
-    libboost-test-dev \
     libboost-thread-dev \
     libevent-dev \
     libffi-dev \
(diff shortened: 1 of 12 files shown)


c8ef:
cc @matthewtlam @jafingerhut for review.

install_deps.sh
```diff
@@ -13,8 +13,6 @@ sudo apt-get install -y \
     g++ \
     libboost-dev \
     libboost-program-options-dev \
-    libboost-system-dev \
-    libboost-test-dev \
     libboost-thread-dev \
```
c8ef, line 16: Note that we still need `libboost-thread-dev` since `p4lang/PI` depends on `boost::thread`. Will clean this up in a follow-up.

matthewtlam: @jafingerhut  Should we open an issue for this to keep track of requiring to update p4lang/PI before removing remaining boost::thread dependency?

matthewtlam (approved):
Changes LGTM, thanks!

install_deps.sh
jafingerhut: Having an issue to remind ourselves sounds like a good idea.  @c8ef Do you mind creating such an issue?

jafingerhut (approved):
FYI fruffy to be clear, while you may be trusting my judgement, note that I am trusting matthewtlam's judgement on the C++ changes here.

install_deps.sh
c8ef: > Having an issue to remind ourselves sounds like a good idea. @c8ef Do you mind creating such an issue?

Sure! Opened https://github.com/p4lang/behavioral-model/issues/1385 to track this.

PR #1386, opened 2026-05-10
Remove bundled xxhash.

c8ef:
As of 2026, I don’t see any reason for us to bundle xxhash in the codebase. CMake/autotools can pull xxhash from the system package manager (e.g. `apt`), while Bazel can fetch it from the BCR.

PR #1389, opened 2026-05-20
Remove bundled jsoncpp.

c8ef:
Follow up of https://github.com/p4lang/behavioral-model/pull/1386.


c8ef:
cc @fruffy @jafingerhut @matthewtlam for review.

c8ef:
related p4c PR: https://github.com/p4lang/p4c/pull/5632

jafingerhut (approved):
I like that it is removing a vendored copy of code and using a Debian/Ubuntu published package instead.  The different versions of it are apparently compatible enough to pass the bit of testing it has gone through in CI, so hopefully it will hold up in more extensive testing later.

I'd prefer that someone else also review and approve this before we merge it, though.

jafingerhut:
@fruffy The breaking-change label is because users of this code will now need to install this package, too?  Or is there some other reason?

fruffy:
> @fruffy The breaking-change label is because users of this code will now need to install this package, too? Or is there some other reason?

Yes, any downstream repository that pulls BMv2 and install it from source will experience breakages

jafingerhut:
> > @fruffy The breaking-change label is because users of this code will now need to install this package, too? Or is there some other reason?
> 
> Yes, any downstream repository that pulls BMv2 and install it from source will experience breakages

No offsense intended, but the build steps tend to require tweaking several times a year in my p4-guide install steps, due to one repository or another changing things slightly.  Not a big deal to keep up with, if you pay attention, but can definitely take one by surprise if you do not.

fruffy:
> No offsense intended, but the build steps tend to require tweaking several times a year in my p4-guide install steps, due to one repository or another changing things slightly. Not a big deal to keep up with, if you pay attention, but can definitely take one by surprise if you do not.

The solution to this is to pin any dependencies and then update periodically. We also do this for our downstream projects. With the `breaking-change` tag it is now easier to identify which commits are problematic for each release cycle.

The problem with maintaining backwards-compatibility is that this org has little manpower and backwards-compatibility takes a lot of effort. In our case the payoff is also nebulous because we do not know who is using our infra and we are not getting any reports. So it is hard to tell at all who is invested in preserving the status quo. 

Maintenance of P4 repos has become a lot easier over time. Partially because we made the decision to merge breaking changes which cleaned up a lot of our infrastructure. I would prefer to continue this practice to avoid stagnation.

fruffy (approved):

PR #1390, opened 2026-05-21
Fix jsoncpp package name for ubuntu 20.04

c8ef:
Fixes https://github.com/p4lang/behavioral-model/actions/runs/26198569485/job/77083312494.

- https://launchpad.net/ubuntu/focal/+package/libjsoncpp1
- https://launchpad.net/ubuntu/jammy/+package/libjsoncpp25

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ ENV BM_RUNTIME_DEPS libboost-program-options1.71.0 \
                     libboost-filesystem1.71.0 \
                     libboost-thread1.71.0 \
                     libgmp10 \
-                    libjsoncpp25 \
+                    libjsoncpp1 \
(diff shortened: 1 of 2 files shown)


c8ef:
cc @fruffy @jafingerhut for review.

c8ef:
Here are a few thoughts on how we can prevent this kind of regression:

1. We could build the image during PR checks, but only push it to the registry when commits are merged into the main branch.
2. We might not need to explicitly specify the runtime library names, as they should be included transitively.

fruffy:
>We could build the image during PR checks, but only push it to the registry when commits are merged into the main branch.

Yes, we do this in P4C already and should also do it here.

fruffy (approved):

PR #1391, opened 2026-05-21
Remove obsolete .license files for recently-removed jsoncpp vendored source

jafingerhut:
Also add copyright and license notices for gmock library, which was added by recent PR as vendored third-party source code.

Also remove two obsolete .license files for recently-removed jsoncpp vendored source.  These files were added by a recent PR of mine, but are no longer needed because another recent PR removed the corresponding source files from third_party (jsoncpp).

If these changes are merged, the `reuse lint` command shows that all files are annotated for copyright and license in a way that it can automatically check, and passes that lint check.

### DIFF
--- a/third_party/gmock/include/gmock/internal/custom/README.md.license
+++ b/third_party/gmock/include/gmock/internal/custom/README.md.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: 2018 Google Inc.
+
+SPDX-License-Identifier: BSD-3-Clause
--- a/third_party/gmock/CMakeLists.txt
+++ b/third_party/gmock/CMakeLists.txt
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2026 Glen Gibb
+#
+# SPDX-License-Identifier: Apache-2.0
+
 # Create gtest library
 add_library(gmock STATIC
   src/gmock-all.cc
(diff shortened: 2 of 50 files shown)


grg:
Is there a way we can do this without touching all the 3rd party files? Right now, they're a direct copy from the 3rd party source.

grg:
Also we don't seem to have modified the files in third-party/gtest

jafingerhut:
> Is there a way we can do this without touching all the 3rd party files? Right now, they're a direct copy from the 3rd party source.

There is.  We can add files with a .license suffix, that leaves the original file unchanged.  I can do that and update this PR with that approach for the gmock files.

jafingerhut:
> Also we don't seem to have modified the files in third-party/gtest

Yeah, that was not very consistent of me, was it?  I'll update the PR.

grg:
> > Is there a way we can do this without touching all the 3rd party files? Right now, they're a direct copy from the 3rd party source.
> 
> There is. We can add files with a .license suffix, that leaves the original file unchanged. I can do that and update this PR with that approach for the gmock files.

Perfect, thanks for making that change. It will make updating easier if the files are unmodified from their original form.

third_party/gmock/include/gmock/internal/custom/README.md.license
grg, line 1: Does the README.md file need a license?

grg (approved):
LGTM

third_party/gmock/include/gmock/internal/custom/README.md.license
jafingerhut: We could use a license different than Apache-2.0, e.g. I have heard some people like to use a very permissive Creative Commons license variant for them, but if you copyright a file and don't put a license on it, then the permissions you are giving others for modifying and/or further copying it is very unclear.

PR #1392, opened 2026-05-21
Enable 'reuse lint' check to verify copyright and license annotations in CI

jafingerhut:


jafingerhut:
This PR also adds a few .license files for .json files in the testdata directory.  I thought I had added these earlier, but I did not realize that a .gitignore file was masking them.  My latest commit tweaks that .gitignore file so that these new files are visible to git.

grg (approved):

PR #1359, opened 2026-03-26
Add Bazel build support for simple_switch_grpc

koustubh-v:
This PR adds initial Bazel build support for simple_switch_grpc as an alternative to the existing CMake-based workflow.

While working on this, I explored the current build system's structure and gradually mapped the required components into Bazel. This includes setting up MODULE.bazel, defining cc_library and cc_binary targets, and wiring core dependencies such as gRPC, protobuf, and spdlog.

The main goal here was to get a working end-to-end Bazel build for simple_switch_grpc. During the process, I also addressed multiple build issues, including missing headers and dependency linking, to ensure the target compiles successfully.

This is a first step toward enabling a cleaner, more modern Bazel-based workflow for BMv2, and I’d really appreciate any feedback or suggestions to improve the structure or dependency handling.

PR #1356, opened 2026-03-19
[GSoC26][bazel] add bazel build system support for simple_switch_grpc

c8ef:
Introduces a complete bazel build alongside the existing cmake setup, targeting the simple_switch_grpc binary and its test suite.

Key implementation notes:

- All BUILD files are formatted with Buildifier and enforced in CI via a dedicated GitHub Actions workflow, keeping the Starlark codebase consistent from day one.
- config.h is generated using the expand_template idiom rather than a raw genrule, making the enabled/disabled feature set explicit and diff-friendly.
- External dependencies (Boost, gmp, nanomsg, grpc, p4lang/PI, etc.) are currently consumed as system-installed libraries through a module extension (bazel/ubuntu2204.bzl). This avoids the need to vendor or pin every transitive dependency upfront; migration to BCR-sourced modules can be done incrementally.
- Groundwork for a full Bazel-native p4lang/PI is already underway: bazelbuild/bazel-central-registry#8031 tracks a third-party library used by the PI library's contribution to BCR, after which the @ubuntu2204//:pi* stubs can be replaced with proper module dependencies.

PR #1355, opened 2026-03-15
bazel: add Bazel 8.5.0 build support for simple_switch_grpc

Shawn-Dong:
## Summary

Add Bazel 8.5.0 build support for `simple_switch_grpc` and its dependencies. The existing autoconf/CMake build is unaffected.

## Design decisions

**System gRPC/protobuf**
BCR protobuf conflicts with p4lang-PI's pre-generated `.pb.h` headers, so system-installed libraries are used to maintain ABI compatibility.

**config.h via genrule**
cmake generates `config.h` into its build directory but adds the source tree to the include path first. A committed `config.h` would shadow cmake's version, so a Bazel genrule generates it into `bazel-out/` instead.

**Exclusive test tags**
All 11 tests bind to shared gRPC ports, so `tags = ["exclusive"]` prevents parallel execution.

## Test plan

- [x] `bazel build //targets/simple_switch_grpc:simple_switch_grpc`
- [x] `bazel test //targets/simple_switch_grpc/tests/...`
- [x] GitHub Actions CI passes
- [x] Existing CMake CI unaffected

PR #1354, opened 2026-03-15
Add Bazel 8.5.0 build support for simple_switch_grpc + CI workflow

kunal9812:
## Summary
This PR adds Bazel 8.5.0 build support for `simple_switch_grpc` as part of
the GSoC 2026 qualification task for Project 1.1: BMv2 Packet Trace Support.

## What This PR Does
- Adds `MODULE.bazel` with bzlmod dependencies (protobuf 29.1, grpc 1.71,
  boost, jsoncpp, fmt, googletest)
- Adds `WORKSPACE.bzlmod` to enable Bazel bzlmod mode
- Adds `BUILD` file at root for `bmv2_core` library
- Adds `BUILD` file for `targets/simple_switch_grpc`
- Vendors PI library headers/sources into `third_party/PI`
- Adds `.github/workflows/ci-bazel.yml` — builds and tests via Bazel

## Verification
- [x] `bazelisk build //targets/simple_switch_grpc:simple_switch_grpc` passes
- [x] `bazelisk test //targets/simple_switch_grpc:simple_switch_grpc_test` — 23/26 pass (3 gNMI/multicast tests require a running network service)

## Notes
- Built and tested on macOS M1 with Bazel 8.5.0 (via bazelisk + .bazelversion)
- The 3 failing tests (`gNMI.PortOperStatusUpdates`, `gNMI.PortCounters`, `Clone.SingletonAndMulticast`) require a live gNMI/multicast environment and fail the same way on the original codebase

## Related
GSoC 2026 Project 1.1 qualification task.
Reference repos: `p4lang/p4runtime` (Bazel structure), `p4lang/p4c` (CI workflow).

PR #1401, opened 2026-05-26
Add bazel build system support for simple_switch_grpc

c8ef:
- Closes #1359.
- Closes #1356.
- Closes #1355.
- Closes #1354.


targets/simple_switch_grpc/tests/base_test.h
```diff
@@ -13,7 +13,7 @@
 #ifndef SIMPLE_SWITCH_GRPC_TESTS_BASE_TEST_H_
 #define SIMPLE_SWITCH_GRPC_TESTS_BASE_TEST_H_
 
-#include <p4/config/v1/p4info.grpc.pb.h>
+#include <p4/config/v1/p4info.pb.h>
```
c8ef, line 16: More info about this: https://github.com/p4lang/behavioral-model/pull/1394

c8ef:
cc @matthewtlam @qobilidop for review.

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,64 @@
+# SPDX-FileCopyrightText: 2026 Yuao Ma
```
matthewtlam, line 1: @jafingerhut and @fruffy Is this the correct protocol for adding copyright in behavioral model?

jafingerhut: This is one way that I think is reasonable to accept.  It is the copyright notice that is automatically added when using the `reuse annotate` command recommended in an article I propose adding to the p4lang/.github repo in this PR: https://github.com/p4lang/.github/pull/8/changes#diff-52abcb9d07d8e602d7ce9a261156503b8d37f72e65e2a8ecd0ab6bb18b58d1ce

Another reasonable choice is to use a more traditional copyright notice that looks like this:
```
# Copyright (C) 2026 Yuao Ma
```
Having both a line like that _and_ the `SPDx-FileCopyrightText` line is also a reasonable choice.

targets/simple_switch_grpc/tests/base_test.h
c8ef: Since #1394 has been merged, I have rebased this PR which only includes changes related to bazel.

.bazelversion.license
matthewtlam, line 1: Is this file needed? I would remove it

MODULE.bazel
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: 2026 Yuao Ma
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module(
+    name = "behavioral-model",
```
matthewtlam: NIT: behavioral_model

MODULE.bazel
```diff
@@ -0,0 +1,36 @@
...
+bazel_dep(name = "googleapis", version = "0.0.0-20260130-c0fcb356", repo_name = "com_google_googleapis")
+bazel_dep(name = "grpc", version = "1.76.0", repo_name = "com_github_grpc_grpc")
+bazel_dep(name = "jsoncpp", version = "1.9.6.bcr.2")
+bazel_dep(name = "libpcap", version = "1.10.5.bcr.3")
+bazel_dep(name = "p4runtime", version = "1.5.0.bcr.1", repo_name = "com_github_p4lang_p4runtime")
+bazel_dep(name = "protobuf", version = "33.5", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_cc", version = "0.2.17")
+bazel_dep(name = "xxhash", version = "0.8.3.bcr.1")
+
+bazel_dep(name = "googletest", version = "1.17.0.bcr.2", dev_dependency = True)
```
matthewtlam, line 30: I see you put everything in alphabetical order. Is there an obvious reason why this one isn't either?

MODULE.bazel
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: 2026 Yuao Ma
+#
+# SPDX-License-Identifier: Apache-2.0
+
```
matthewtlam, line 4: It is apart of Google standard to have the following https://github.com/p4lang/p4c/blob/main/BUILD.bazel#L8C1-L20C27

BUILD.bazel
```diff
@@ -0,0 +1,56 @@
+# SPDX-FileCopyrightText: 2026 Yuao Ma
+#
+# SPDX-License-Identifier: Apache-2.0
+
```
matthewtlam, line 4: Will need to add a package statement in all .bazel files for example: https://github.com/google/p4-infra/blob/903bebb041cc95b8aadb3e4ea8a216a0081de584/p4_pdpi/BUILD.bazel#L25C1-L28C2

BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+load("@bazel_skylib//rules:expand_template.bzl", "expand_template")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+exports_files(["VERSION"])
```
matthewtlam: Could you say a little bit of why we export the version in this file specifically?

PI/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+cc_library(
+    name = "bmpi",
+    srcs = glob([
+        "src/*.cpp",
+        "src/*.h",
+    ]),
+    hdrs = ["bm/PI/pi.h"],
+    includes = ["."],
```
matthewtlam, line 19: Might be okay for now since we are adding support for Bazel in legacy code, but is generally discouraged in Bazel projects because of
Header Collisions: If multiple libraries use includes = ["."], and both have a file named utils.h, it can lead to compiler confusion and hard-to-debug header resolution errors.
Explicit is Better: Bazel prefers absolute paths relative to the workspace root (e.g., #include "src/parser/parser.h") because it makes it immediately clear where a header is coming from.

Perhaps try removing them and seeing if the tests still pass

targets/simple_switch_grpc/tests/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
+# SPDX-FileCopyrightText: 2026 Yuao Ma
+#
+# SPDX-License-Identifier: Apache-2.0
+
+load("@rules_cc//cc:cc_test.bzl", "cc_test")
+
+_COMMON_SRCS = [
```
matthewtlam: Is there a particular reason why these are not apart of the _COMMON_DEPS?

targets/simple_switch_grpc/tests/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+    "test_meter",
+    "test_ternary",
+    "test_pre",
+    "test_digest",
+    "test_idle_timeout",
+    "test_action_profile",
+    "test_optional",
+]
+
+[
```
matthewtlam, line 57: Overall Bazel strongly discourages adding unnecessary dependencies and explicit library dependencies, but this is okay since I feel like all the test require all the exact same dependencies

.bazelversion.license
c8ef: Yes, it is necessary for the copyright check. We cannot include the copyright notice in the .bazelversion file.

MODULE.bazel
c8ef: Buildifier requires dev dependencies to be listed separately.

MODULE.bazel
c8ef: I don't have a strong opinion on this, but is using this different from the project name on GitHub?

BUILD.bazel
c8ef: https://github.com/c8ef/behavioral-model/blob/bc6ea00ee82890d856eca97192f9c3a47dd9da92/src/bm_sim/BUILD.bazel#L11

Because I want to use the content of the VERSION file directly to parse the version number. It's a bit hacky, but it will make release easier.

targets/simple_switch_grpc/tests/BUILD.bazel
c8ef: `_COMMON_SRCS` is a list of source files, whereas `_COMMON_DEPS` is a list of local or remote dependencies.

PI/BUILD.bazel
c8ef: I tried this approach, but it didn't work. After investigating, I found that bmv2 includes BM-related headers using angle brackets (`< >`), which means the paths must be passed explicitly using `-I`. I attempted to use `strip_include_prefix`, but it passes the paths to the compiler via `-iquote`, which doesn't resolve angle bracket includes. Simply removing these includes is not an option.

ref: https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html

```
Directories specified with -iquote apply only to the quote form of the directive, #include "file". Directories specified with -I, -isystem, or -idirafter apply to lookup for both the #include "file" and #include <file> directives.
```

targets/simple_switch_grpc/tests/BUILD.bazel
c8ef: Yeah, it seems like the same dependencies are being used but different scenarios are being tested.

BUILD.bazel
c8ef: Should I still add package level visibility even though I have explicitly added `visibility = ["//visibility:public"]` for all `cc_library`?

MODULE.bazel
c8ef: I just realized that there isn't a root-level license file for bmv2. Given this, I'm not sure if simply adding an Apache 2.0 notice is the best approach. Love to hear your thoughts on this @jafingerhut.

jafingerhut: I will add a root-level LICENSE file that is a symbol link to LICENSES/Apache-2.0.txt soon.

.bazelversion.license
jafingerhut: @matthewtlam If .bazelversion files have an explicit comment syntax that they support (e.g. everything after a '#' character on a line is a comment), then it is usually preferred to have the copyright and license notice in comments in the file itself.

If .bazelversion files do not suport comments, then the REUSE tool that is now being used in behavioral-model (and soon hopefully all p4lang repositories) requires that there be a separate file with a suffix of ".license" in its name that contains the copyright and license info.  There is also a way to put this info in a different separate file, but the ".license" file is good if there are only a few files that don't support comment syntax.

MODULE.bazel
jafingerhut: Done a few minutes ago when I merged this PR: https://github.com/p4lang/behavioral-model/pull/1403

If you update this PR's branch to the latest master/main version, the LICENSE file should be there.

c8ef: Added using https://github.com/bazel-contrib/supply-chain, `@rules_license`'s successor.

PI/BUILD.bazel
matthewtlam: Thanks for investigating! Sounds like it's inevitable due to this being legacy code while trying to add Bazel support

BUILD.bazel
matthewtlam: Yes, this should be added for each subcomponent

MODULE.bazel
matthewtlam: Thanks @jafingerhut!

MODULE.bazel
matthewtlam: This one is just for consistency with all other names we have where they have underscores

MODULE.bazel
matthewtlam: Sounds good. Marking as resolved

matthewtlam (commented):
PTAL at my comments

MODULE.bazel
c8ef: Fixed.

BUILD.bazel
c8ef: Fixed.

c8ef:
> PTAL at my comments

All resolved, please take another look.

.bazelversion.license
matthewtlam: I didn't wanna interfere with your copyright work @jafingerhut. So long as you are happy, I don't mind regarding the format

BUILD.bazel
matthewtlam: I don't have a strong preference. Thanks for clarifying

targets/simple_switch_grpc/tests/BUILD.bazel
matthewtlam: Is it possible to make all the common source files be built similarly to the common deps? So that the only source files are the tests itself?

matthewtlam (commented):
Just looking for clarification on a couple of last comments but we are almost there!

targets/simple_switch_grpc/tests/BUILD.bazel
c8ef: I moved the common sources into a test_common library, so they can now be treated as common deps. However, the tests still need to run exclusively due to port sharing.

c8ef:
> Just looking for clarification on a couple of last comments but we are almost there!

Thanks for the detailed review! I took the initiative to mark some addressed comments as resolved, but feel free to reopen them if you have further feedback.

matthewtlam (approved):
The refactoring in targets/simple_switch_grpc/tests/BUILD.bazel looks amazing and very explicit following Bazel standards. LGTM!

jafingerhut (approved):
Approving and merging based upon matthewtlam's review and approval.

PR #1405, opened 2026-06-14
packet trace format and utility

c8ef:
This PR introduces the first phase of packet trace support, focusing primarily on the following foundational additions:

- Defines the structured data format for packet tracing. I initially considered a flattened format using `packet_id` and `copy_id` to track packets. However, since handling packet forks is unnatural and complex in a flattened structure, I opted for a recursive trace tree format instead.
- Adjusts all three build systems (bazel, cmake, and autotools).
- Introduces the foundational C++ classes and structures for managing the trace tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,7 @@ include_directories(
 find_package(Threads REQUIRED)
 find_package(Boost REQUIRED COMPONENTS thread program_options)
 find_package(jsoncpp REQUIRED)
+find_package(Protobuf REQUIRED)
 
 # Check for the presence of an include file and report an error if not found
 #
@@ -170,8 +171,7 @@ if(WITH_THRIFT)
 endif()
 
 if(WITH_PI)
-  # Find Protobuf and gRPC
-  find_package(Protobuf REQUIRED)
+  # Find gRPC
   find_package(gRPC REQUIRED)
 
   # Verify that PI headers are present
@@ -220,6 +220,7 @@ if(WITH_THRIFT)
 endif()
 
 add_subdirectory(third_party)
+add_subdirectory(proto)
 add_subdirectory(src)
 add_subdirectory(include)
 
--- a/configure.ac
+++ b/configure.ac
@@ -275,14 +275,14 @@ AC_TYPE_SIZE_T
 AC_TYPE_UINT64_T
 AC_LANG_POP(C++)
 
+AC_PATH_PROG([PROTOC], [protoc], [])
+AS_IF([test "x$PROTOC" = x], [AC_MSG_ERROR([protoc not found])])
+
+PKG_CHECK_MODULES([PROTOBUF], [protobuf >= 3.0.0])
+AC_SUBST([PROTOBUF_CFLAGS])
+AC_SUBST([PROTOBUF_LIBS])
+
 AS_IF([test "$want_pi" = yes], [
-	AC_PATH_PROG([PROTOC], [protoc], [])
-	AS_IF([test "x$PROTOC" = x], [AC_MSG_ERROR([protoc not found])])
-
-	PKG_CHECK_MODULES([PROTOBUF], [protobuf >= 3.0.0])
-	AC_SUBST([PROTOBUF_CFLAGS])
-	AC_SUBST([PROTOBUF_LIBS])
-
 	PKG_CHECK_MODULES([GRPC], [grpc++ >= 1.3.0 grpc >= 3.0.0])
 	AC_SUBST([GRPC_CFLAGS])
 	AC_SUBST([GRPC_LIBS])
@@ -309,6 +309,7 @@ AC_CONFIG_FILES([Makefile
 		third_party/gmock/Makefile
 		third_party/gtest/Makefile
 		third_party/spdlog/Makefile
+		proto/Makefile
 		include/Makefile
 		src/Makefile
 		src/bm_sim/Makefile
--- a/include/bm/bm_sim/packet_tracer.h
+++ b/include/bm/bm_sim/packet_tracer.h
@@ -1,0 +1,77 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Yuao Ma
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! @file packet_tracer.h
+//!
+//! Protobuf-based structured packet tracing backend.
+
+#ifndef BM_BM_SIM_PACKET_TRACER_H_
+#define BM_BM_SIM_PACKET_TRACER_H_
+
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <string>
+
+#include "packet_trace.pb.h"
+
+namespace bm {
+
+class Packet;
+
+class TraceTreeWrapper {
+ public:
+  TraceTreeWrapper(uint64_t packet_id, uint32_t ingress_port);
+  ~TraceTreeWrapper();
+
+ private:
+  friend class PacketTraceContext;
+
+  p4::bm::PacketTrace root_trace_;
+};
+
+//! Per-packet trace accumulator. One instance is attached to each Packet object
+//! when tracing is enabled. Accumulates TraceEvent messages in chronological
+//! order.
+class PacketTraceContext {
+ public:
+  PacketTraceContext(uint64_t packet_id, uint32_t ingress_port);
+  PacketTraceContext(std::shared_ptr<TraceTreeWrapper> tree_wrapper,
+                     p4::bm::TraceTree* node);
+
+  std::shared_ptr<TraceTreeWrapper> get_wrapper() const {
+    return tree_wrapper_;
+  }
+
+ private:
+  std::shared_ptr<TraceTreeWrapper> tree_wrapper_;
+  p4::bm::TraceTree* current_node_;
+};
+
+//! Protobuf trace backend.
+class PacketTracer {
+ public:
+  static PacketTracer* get() {
+    static PacketTracer instance;
+    return &instance;
+  }
+
+  //! Set the output directory for trace files.
+  void set_output_dir(const std::string& output_dir);
+
+  //! Write a completed trace to the output directory as txtpb.
+  void flush_trace(const p4::bm::PacketTrace& trace);
+
+ private:
+  PacketTracer() = default;
+
+  bool enabled_{false};
+  std::string output_dir_;
+};
+
+}  // namespace bm
+
+#endif  // BM_BM_SIM_PACKET_TRACER_H_
--- a/proto/packet_trace.proto
+++ b/proto/packet_trace.proto
@@ -86,8 +86,6 @@ message TraceEvent {
     // Packet replication engine (PRE) events.
     CloneEvent clone = 7;
     MulticastEvent multicast = 8;
-    ResubmitEvent resubmit = 9;
-    RecirculateEvent recirculate = 10;
 
     // Deparser events.
     DeparserEmitEvent deparser_emit = 11;
@@ -172,12 +170,6 @@ message MulticastEvent {
   uint32 multicast_group_id = 1;
   uint32 replica_count = 2;
 }
-
-// Fired when a packet is resubmitted (re-enters the ingress pipeline).
-message ResubmitEvent {}
-
-// Fired when a packet is recirculated (deparsed packet loops back to ingress).
-message RecirculateEvent {}
 
 // Fired when the deparser emits a valid header to the output packet buffer.
 message DeparserEmitEvent {
--- a/proto/packet_trace.proto
+++ b/proto/packet_trace.proto
@@ -1,0 +1,190 @@
+// SPDX-FileCopyrightText: 2026 Yuao Ma
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// BMv2 Structured Packet Trace
+//
+// A structured, programmatic representation of a packet's journey through a P4
+// pipeline in BMv2. Designed for use by automated tools for test coverage
+// measurement, automated test generation, root-cause analysis, and human
+// comprehension.
+
+syntax = "proto3";
+
+package p4.bm;
+
+// Top-level trace for one packet copy through the pipeline.
+message PacketTrace {
+  uint64 packet_id = 1;
+  uint32 ingress_port = 2;
+  TraceTree root = 3;
+}
+
+// Recursive trace structure: a sequence of events followed by an outcome.
+message TraceTree {
+  repeated TraceEvent events = 1;
+  oneof outcome {
+    // All branches happen simultaneously; outputs are the union.
+    // Used for clone and multicast.
+    Replication replication = 2;
+
+    // The same packet continues as another parser-to-deparser pass.
+    // Used for resubmit and recirculate.
+    Continuation continuation = 3;
+
+    // Terminal: packet emitted on a port.
+    Transmit transmit = 4;
+
+    // Terminal: packet discarded.
+    Drop drop = 5;
+  }
+}
+
+// All branches execute; their output packets are the union.
+// Used for any mechanism that sends copies to multiple ports (clone,
+// multicast).
+message Replication {
+  repeated TraceTree branches = 1;
+}
+
+// The same packet continues as another parser-to-deparser pass.
+// Used for resubmit and recirculate.
+message Continuation {
+  enum Kind {
+    KIND_UNSPECIFIED = 0;
+    RESUBMIT = 1;
+    RECIRCULATE = 2;
+  }
+  Kind kind = 1;
+  TraceTree next = 2;
+}
+
+// Packet was transmitted out of a port.
+message Transmit {
+  uint32 egress_port = 1;
+  uint32 packet_size = 2;
+}
+
+// Packet was dropped.
+message Drop {}
+
+// A single event in the packet trace.
+message TraceEvent {
+  oneof event {
+    // Pipeline lifecycle.
+    PipelineStageEvent pipeline_stage = 1;
+
+    // Parser events.
+    ParserTransitionEvent parser_transition = 2;
+    ParserExtractEvent parser_extract = 3;
+
+    // Control flow events.
+    TableLookupEvent table_lookup = 4;
+    ActionExecutionEvent action_execution = 5;
+    ConditionEvalEvent condition_eval = 6;
+
+    // Packet replication engine (PRE) events.
+    CloneEvent clone = 7;
+    MulticastEvent multicast = 8;
+    ResubmitEvent resubmit = 9;
+    RecirculateEvent recirculate = 10;
+
+    // Deparser events.
+    DeparserEmitEvent deparser_emit = 11;
+    ChecksumUpdateEvent checksum_update = 12;
+  }
+}
+
+// Fired when the pipeline enters or exits a processing stage.
+message PipelineStageEvent {
+  string stage_name = 1;
+
+  enum StageKind {
+    STAGE_KIND_UNSPECIFIED = 0;
+    PARSER = 1;
+    CONTROL = 2;
+    DEPARSER = 3;
+  }
+  StageKind stage_kind = 2;
+
+  enum Direction {
+    DIRECTION_UNSPECIFIED = 0;
+    ENTER = 1;
+    EXIT = 2;
+  }
+  Direction direction = 3;
+}
+
+// Fired on every parser state transition, including the final transition to
+// the implicit "accept" or "reject" state.
+message ParserTransitionEvent {
+  string parser_name = 1;
+  // Name of the state being left. Empty for the initial transition.
+  string from_state = 2;
+  // Name of the state being entered. Empty string for "accept" (end of
+  // parsing). "reject" for parser errors.
+  string to_state = 3;
+}
+
+// Fired on every parser header extraction.
+message ParserExtractEvent {
+  string header_name = 1;
+}
+
+// Fired on every table.apply() call. Records hit/miss, matched entry, and
+// selected action.
+message TableLookupEvent {
+  string table_name = 1;
+  bool hit = 2;
+  // Entry handle of the matched entry. Only meaningful on hit.
+  uint32 entry_handle = 3;
+  // Name of the action selected (either from the matched entry or the default
+  // action on miss).
+  string action_name = 4;
+  // Human-readable description of the matched entry (match key + action data).
+  string entry_info = 5;
+}
+
+// Fired when an action begins executing.
+message ActionExecutionEvent {
+  string action_name = 1;
+}
+
+// Fired on every if/else condition evaluation.
+message ConditionEvalEvent {
+  string condition_name = 1;
+  bool result = 2;
+}
+
+// Fired when clone (ingress or egress) is triggered.
+message CloneEvent {
+  enum CloneType {
+    CLONE_TYPE_UNSPECIFIED = 0;
+    INGRESS = 1;
+    EGRESS = 2;
+  }
+  CloneType clone_type = 1;
+  uint32 mirror_session_id = 2;
+}
+
+// Fired when multicast replication occurs.
+message MulticastEvent {
+  uint32 multicast_group_id = 1;
+  uint32 replica_count = 2;
+}
+
+// Fired when a packet is resubmitted (re-enters the ingress pipeline).
+message ResubmitEvent {}
+
+// Fired when a packet is recirculated (deparsed packet loops back to ingress).
+message RecirculateEvent {}
+
+// Fired when the deparser emits a valid header to the output packet buffer.
+message DeparserEmitEvent {
+  string header_name = 1;
+}
+
+// Fired when a checksum is updated by the deparser.
+message ChecksumUpdateEvent {
+  string checksum_name = 1;
+}
--- a/src/bm_sim/packet_tracer.cpp
+++ b/src/bm_sim/packet_tracer.cpp
@@ -1,0 +1,67 @@
+// SPDX-FileCopyrightText: 2026 Yuao Ma
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include <bm/bm_sim/logger.h>
+#include <bm/bm_sim/packet.h>
+#include <bm/bm_sim/packet_tracer.h>
+#include <google/protobuf/text_format.h>
+
+#include <fstream>
+#include <sstream>
+#include <string>
+#include <utility>
+
+namespace bm {
+
+// ---------- TraceTreeWrapper ----------
+
+TraceTreeWrapper::TraceTreeWrapper(uint64_t packet_id, uint32_t ingress_port) {
+  root_trace_.set_packet_id(packet_id);
+  root_trace_.set_ingress_port(ingress_port);
+}
+
+TraceTreeWrapper::~TraceTreeWrapper() {
+  PacketTracer::get()->flush_trace(root_trace_);
+}
+
+// ---------- PacketTraceContext ----------
+
+PacketTraceContext::PacketTraceContext(uint64_t packet_id,
+                                       uint32_t ingress_port) {
+  tree_wrapper_ = std::make_shared<TraceTreeWrapper>(packet_id, ingress_port);
+  current_node_ = tree_wrapper_->root_trace_.mutable_root();
+}
+
+PacketTraceContext::PacketTraceContext(
+    std::shared_ptr<TraceTreeWrapper> tree_wrapper, p4::bm::TraceTree* node)
+    : tree_wrapper_(std::move(tree_wrapper)), current_node_(node) {}
+
+// ---------- PacketTracer ----------
+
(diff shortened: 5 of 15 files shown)


c8ef:
cc @matthewtlam @qobilidop for review.

fruffy:
@grg This might be of interested to you. The detailed project description is here: https://summerofcode.withgoogle.com/programs/2026/projects/DAxIafkq

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+  TraceTree root = 3;
+}
+
+// Recursive trace structure: a sequence of events followed by an outcome.
+message TraceTree {
+  repeated TraceEvent events = 1;
+  oneof outcome {
+    // All branches happen simultaneously; outputs are the union.
+    // Used for clone and multicast.
+    Replication replication = 2;
```
matthewtlam, line 29: If you plan to use this for both multicast and choice (i.e. WCMP, ECMP, etc). How do you plan to differentiate between the two in terms of behavior?

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+// comprehension.
+
+syntax = "proto3";
+
+package p4.bm;
+
+// Top-level trace for one packet copy through the pipeline.
+message PacketTrace {
+  uint64 packet_id = 1;
+  uint32 ingress_port = 2;
```
matthewtlam, line 19: Out of curiosity, why did you choose to make this as a part of the PacketTrace rather than an event type?

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+    // Control flow events.
+    TableLookupEvent table_lookup = 4;
+    ActionExecutionEvent action_execution = 5;
+    ConditionEvalEvent condition_eval = 6;
+
+    // Packet replication engine (PRE) events.
+    CloneEvent clone = 7;
+    MulticastEvent multicast = 8;
+    ResubmitEvent resubmit = 9;
+    RecirculateEvent recirculate = 10;
```
matthewtlam: I don't believe we need a recirculation/resubmit event since it is extensively encoded in the Continuation outcome type

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+// Fired when a packet is recirculated (deparsed packet loops back to ingress).
+message RecirculateEvent {}
+
+// Fired when the deparser emits a valid header to the output packet buffer.
+message DeparserEmitEvent {
+  string header_name = 1;
+}
+
+// Fired when a checksum is updated by the deparser.
+message ChecksumUpdateEvent {
```
matthewtlam, line 180: In 4ward this is apart of the pipeline stage event. Out of curiosity what does a checksum_name look like or do you have an example of one?

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+
+// Fired when clone (ingress or egress) is triggered.
+message CloneEvent {
+  enum CloneType {
+    CLONE_TYPE_UNSPECIFIED = 0;
+    INGRESS = 1;
+    EGRESS = 2;
+  }
+  CloneType clone_type = 1;
+  uint32 mirror_session_id = 2;
```
matthewtlam, line 165: May want a variable to specify the number of clones that are created

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+  string header_name = 1;
+}
+
+// Fired on every table.apply() call. Records hit/miss, matched entry, and
+// selected action.
+message TableLookupEvent {
+  string table_name = 1;
+  bool hit = 2;
+  // Entry handle of the matched entry. Only meaningful on hit.
+  uint32 entry_handle = 3;
```
matthewtlam, line 138: Is this the correct type of how entries are represented in BMv2?

proto/packet_trace.proto
```diff
@@ -0,0 +1,190 @@
...
+message Drop {}
+
+// A single event in the packet trace.
+message TraceEvent {
+  oneof event {
+    // Pipeline lifecycle.
+    PipelineStageEvent pipeline_stage = 1;
+
+    // Parser events.
+    ParserTransitionEvent parser_transition = 2;
```
matthewtlam, line 78: What is the different between the ParserTransitionEvent vs the ParserextractEvent?

matthewtlam: Ah I see that they are abstracted to be separated. Is there a specific use case where you see one existing without the other?

include/bm/bm_sim/packet_tracer.h
```diff
@@ -0,0 +1,77 @@
...
+//! Protobuf trace backend.
+class PacketTracer {
+ public:
+  static PacketTracer* get() {
+    static PacketTracer instance;
+    return &instance;
+  }
+
+  //! Set the output directory for trace files.
+  void set_output_dir(const std::string& output_dir);
```
matthewtlam, line 63: Perhaps we may want to specify that this should only be called once and add a mutex for output_dir_. There could be an opportunity where the user calls this function in parallel with flush_trace

matthewtlam: Alternatively we can make enabled_ a std::atomic<bool>

src/bm_sim/packet_tracer.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
+  if (!enabled_) return;
+
+  std::string txtpb;
+  google::protobuf::TextFormat::PrintToString(trace, &txtpb);
+
+  std::ostringstream filename;
+  filename << output_dir_ << "/trace_" << trace.packet_id() << ".txtpb";
```
matthewtlam: optional: May wish to use std::filesystem::path to handle paths

src/bm_sim/packet_tracer.cpp
```diff
@@ -0,0 +1,67 @@
...
+  enabled_ = true;
+  BMLOG_DEBUG("Packet tracer initialized with output directory: {}",
+              output_dir);
+}
+
+void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
+  if (!enabled_) return;
+
+  std::string txtpb;
+  google::protobuf::TextFormat::PrintToString(trace, &txtpb);
```
matthewtlam, line 61: Can we pass this as a const &?

src/bm_sim/packet_tracer.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+PacketTraceContext::PacketTraceContext(
+    std::shared_ptr<TraceTreeWrapper> tree_wrapper, p4::bm::TraceTree* node)
+    : tree_wrapper_(std::move(tree_wrapper)), current_node_(node) {}
+
+// ---------- PacketTracer ----------
+
+void PacketTracer::set_output_dir(const std::string& output_dir) {
+  output_dir_ = output_dir;
```
matthewtlam, line 51: Do we want to add checks to see if the directory actually exists or is writeable?

matthewtlam (commented):
Left some comments, but looking good to start @c8ef

proto/packet_trace.proto
jafingerhut: It is semantically legal to write a P4 parser state with no extract statements, but it must have a transition.  One can also write a parser state with multiple extract statements, but again, only one transition.  That may be the motivation here.

src/bm_sim/packet_tracer.cpp
c8ef: It's already `const &`, the `&txtpb` is a pointer required by the protobuf PrintToString API.

include/bm/bm_sim/packet_tracer.h
c8ef: In the current design, `set_output_dir` is called exactly once during initialization before any packets are processed. `flush_trace` is only called from `TraceTreeWrapper::~TraceTreeWrapper()` during packet teardown. Since they have a strict happened-before relationship, I think there is no need a mutex here.

src/bm_sim/packet_tracer.cpp
c8ef: Fixed.

src/bm_sim/packet_tracer.cpp
c8ef: Fixed.

proto/packet_trace.proto
c8ef: For parser extractions, I plan to use an existing trace point, such as the one [here](https://github.com/p4lang/behavioral-model/blob/main/src/bm_sim/parser.cpp#L237). For parser transitions, I plan to extract the necessary information from the while loop located [here](https://github.com/p4lang/behavioral-model/blob/main/src/bm_sim/parser.cpp#L1139).

proto/packet_trace.proto
c8ef: https://github.com/p4lang/behavioral-model/blob/main/include/bm/bm_sim/event_logger.h#L46

I follow the style above.

proto/packet_trace.proto
c8ef: `Checksum` is a subclass of `P4NamedObject`, so it does have a name field.

https://github.com/p4lang/behavioral-model/blob/main/tests/test_checksums.cpp#L133

proto/packet_trace.proto
c8ef: I followed the style used in 4ward, where ingress_port is a field within InputPacket rather than an event.

proto/packet_trace.proto
c8ef: IIUC action selectors in BMv2 deterministically pick one member per packet at runtime, so the trace only records the single path actually taken (as a normal TableLookupEvent with the resolved action). There's no ambiguity to model.

proto/packet_trace.proto
c8ef: IIUC, both ingress and egress clones yield only a single new packet. On the contrary we have a field replica_count for MulticastEvent.

proto/packet_trace.proto
c8ef: Good catch! Removed.

proto/packet_trace.proto
grg: Why do we need to record extractions? I don't believe any conditional extraction occurs. Although doing so does create symmetry between the parsing and deparsing messages.

(I don't recall how BMv2 handles the case where there is insufficient data.)

proto/packet_trace.proto
```diff
@@ -0,0 +1,182 @@
...
+    DIRECTION_UNSPECIFIED = 0;
+    ENTER = 1;
+    EXIT = 2;
+  }
+  Direction direction = 3;
+}
+
+// Fired on every parser state transition, including the final transition to
+// the implicit "accept" or "reject" state.
+message ParserTransitionEvent {
```
grg, line 118: Do we need to record any information about why a transition was taken/which transition was taken where there are multiple matches in a select statement that lead to the same next state?

proto/packet_trace.proto
```diff
@@ -0,0 +1,182 @@
...
+  // action on miss).
+  string action_name = 4;
+  // Human-readable description of the matched entry (match key + action data).
+  string entry_info = 5;
+}
+
+// Fired when an action begins executing.
+message ActionExecutionEvent {
+  string action_name = 1;
+}
```
grg, line 149: What's the reason we need this *and* `action_name` on line 141?

proto/packet_trace.proto
```diff
@@ -0,0 +1,182 @@
...
+
+// Fired on every if/else condition evaluation.
+message ConditionEvalEvent {
+  string condition_name = 1;
```
grg, line 153: Presumably this will usually be the conditional expression? e.g.,`eth.type == 0x08100 || eth.type == 0x9100 || ...`

How do we distinguish between identical conditions in different conditional expressions? Do we need to include line information here and in other messages?

proto/packet_trace.proto
grg: Should we be considering packets popping into existence *inside* the pipeline? For example, some sort of digest creator that generates packets after accumulating sufficient data/some timeout? Of course we could potentially have input ports for each potential injection point.

proto/packet_trace.proto
```diff
@@ -0,0 +1,182 @@
...
+// pipeline in BMv2. Designed for use by automated tools for test coverage
+// measurement, automated test generation, root-cause analysis, and human
+// comprehension.
+
+syntax = "proto3";
+
+package p4.bm;
+
+// Top-level trace for one packet copy through the pipeline.
+message PacketTrace {
```
grg, line 17: Any need/use for including the input packet here? And correspondingly, the output packet in Transmit messages?

proto/packet_trace.proto
```diff
@@ -0,0 +1,182 @@
...
+    CloneEvent clone = 7;
+    MulticastEvent multicast = 8;
+
+    // Deparser events.
+    DeparserEmitEvent deparser_emit = 11;
+    ChecksumUpdateEvent checksum_update = 12;
+  }
+}
+
+// Fired when the pipeline enters or exits a processing stage.
```
grg, line 96: Is the intention to have both an ENTER and an EXIT event for each stage?

proto/packet_trace.proto
```diff
@@ -0,0 +1,182 @@
...
+  // Name of the state being left. Empty for the initial transition.
+  string from_state = 2;
+  // Name of the state being entered. Empty string for "accept" (end of
+  // parsing). "reject" for parser errors.
+  string to_state = 3;
+}
+
+// Fired on every parser header extraction.
+message ParserExtractEvent {
+  string header_name = 1;
```
grg, line 129: Consider adding a size for variable-length header fields.

grg (commented):
Sending some initial feedback on just the proto file

CMakeLists.txt
```diff
@@ -102,6 +102,7 @@ include_directories(
 find_package(Threads REQUIRED)
 find_package(Boost REQUIRED COMPONENTS thread program_options)
 find_package(jsoncpp REQUIRED)
+find_package(Protobuf REQUIRED)
```
fruffy, line 105: This makes Protobuf required for everyone. Previously this was not the case. This is a breaking change.

configure.ac
```diff
@@ -275,14 +275,14 @@ AC_TYPE_SIZE_T
 AC_TYPE_UINT64_T
 AC_LANG_POP(C++)
 
-AS_IF([test "$want_pi" = yes], [
-	AC_PATH_PROG([PROTOC], [protoc], [])
-	AS_IF([test "x$PROTOC" = x], [AC_MSG_ERROR([protoc not found])])
+AC_PATH_PROG([PROTOC], [protoc], [])
```
fruffy, line 278: Same here. Breaking change.

fruffy (changes requested):

proto/packet_trace.proto
jafingerhut: It does seem likely that having something in a packet processing trace to explicitly indicate that an extract failed because there was insufficient data in the packet would be useful.  Also if a variable-length extract is executed, what the attempted extract number of bits was.

proto/packet_trace.proto
jafingerhut: I am not sure if this is part of what @mathwtlam meant, but one thing that some folks in a previous GSoC project have at least partially implemented is the ability for BMv2 to show the packet trace for _every_ possibly alternative of an ActionSelector, instead of only one, after sending in only a single input packet.  If a trace format were to support that option, as well as packets with more than one ActionSelector executed for the same input packet, and/or one or more packet replications occurs for the same input packet, that could require extra data in the trace to represent that somehow?

CMakeLists.txt
c8ef: Is it feasible to introduce this breaking change, given the widespread adoption of protobuf?

fruffy: Protobuf is not a great dependency to have. It's large, the default versions on Ubuntu are fairly old, across versions Protobuf breaks easily and the inter-version conflicts cause continuous problems. We had to introduce a mode to P4C to build without it because of the compile-time issues caused. 

If possible, we want to avoid depending on it. Do you need trace support without PI?

include/bm/bm_sim/packet_tracer.h
matthewtlam: Let's add a comment to make sure that this is explicit in case other users try to call this incorrectly

proto/packet_trace.proto
matthewtlam: Hmmm, I think ideally we would have an enter and exit event for each stage though perhaps it may seem redundant though I suppose it can be seen as open/close brackets. WDYT @grg?

proto/packet_trace.proto
matthewtlam: Hmmm, @grg could you speak more upon this. I can see the benefit of this for debugging, but might be more obvious to me with an example

proto/packet_trace.proto
matthewtlam: Ack, thanks

proto/packet_trace.proto
matthewtlam: The inspiration for this was likely from https://github.com/4ward-p4/4ward/blob/c49ad528c17481f16b682151c327baba9d19745c/simulator/simulator.proto#L207 where there is also action parameters that are passed. Would be nice to specify the parameters passed to the action as well. I am also okay to get rid of this for now and implement it later if we think it will be useful

proto/packet_trace.proto
matthewtlam: Ack, thanks for clarifying

src/bm_sim/packet_tracer.cpp
matthewtlam: Thanks!

src/bm_sim/packet_tracer.cpp
matthewtlam: Thanks for clarifying

proto/packet_trace.proto
matthewtlam: I believe this is represented by the packet_id and we can do a quick lookup afterwards, though I am not opposed to a copy of the input packet here either. No strong preference from me

@c8ef or @grg Do you have a strong opinion?

proto/packet_trace.proto
matthewtlam: @c8ef Ack, LGTM!

matthewtlam: @grg My initial impression was that we wouldn't support this. At least most standard P4 architectures currently support like v1 and PSA, the pipeline is strict sequential with the packet always starting at the ingress or egress parser. Not sure about the difficulty of supporting this later though. Any idea? I suppose the refactoring wouldn't be too crazy with the use of AI, but perhaps we don't wanna cross that bridge

proto/packet_trace.proto
matthewtlam: @jafingerhut is correct. One of the previous GSoC projects, which did not get officially submitted was to show all the possible output packets for an ActionSelector (for example WCMP). We may wish to differentiate the two with additional data fields in the future so it might be easier to separate the two (now) since they do behavior quite differently mechanically

proto/packet_trace.proto
matthewtlam: Since this is taking inspiration from https://github.com/4ward-p4/4ward/blob/c49ad528c17481f16b682151c327baba9d19745c/simulator/simulator.proto#L165, it seems that SourceLine is part of the TraceEvent, we will likely want to add this to differentiate the case like you mentioned  @c8ef

proto/packet_trace.proto
matthewtlam: Exactly as @jafingerhut mentioned. It's super helpful for cases like hop by hop options for IPv6 where there is a variable length extraction. Also, that it's nice for symmetry as @grg mentioned

matthewtlam: @c8ef Ack. SGTM

proto/packet_trace.proto
matthewtlam: @grg could you say a bit more about this. I'm curious if there is an obvious example that I might be missing

CMakeLists.txt
c8ef: Understood. In that case, I'll handle the protobuf opt-out the same way we do for nanomsg.

c8ef:
Hi, first of all, thanks for the thorough review!

While working through the comments, I realized there is still a lot for me to uncover regarding BMv2 internals. Because of this, trying to design an optimal and complete trace format right out of the gate might be unrealistic. Additionally, getting packet tracing to work requires other prerequisites, most notably an event logger refactor.

Given this, would it make sense to take an incremental approach? If we start with the minimal trace skeleton below, I believe we'll have a solid baseline for the structure and events. Once we have a working end-to-end (e2e) pipeline, we can iteratively enrich the format and handle corner cases.

To be done in this PR:

- Update the proto to the minimal version below (I've added a new InputPacket to allow for future input extensions).
- Make the protobuf dependency optional for the Makefile/CMake builds.
- Resolve any remaining C++ comments from the review above.

To be done for the e2e example (future work):

- CLI integration
- Event logger refactor (NFC)
- Packet tracer implementation

WDYT? 
@matthewtlam @qobilidop @fruffy @grg @jafingerhut 

```proto
syntax = "proto3";

package p4.bm;

message InputPacket {
  uint64 packet_id = 1;
  uint32 ingress_port = 2;
}

// Top-level trace for one packet copy through the pipeline.
message PacketTrace {
  InputPacket input = 1;
  TraceTree root = 3;
}

// Recursive trace structure: a sequence of events followed by an outcome.
message TraceTree {
  repeated TraceEvent events = 1;
}

// A single event in the packet trace.
message TraceEvent {
  oneof event {
    // Pipeline lifecycle.
    PipelineStageEvent pipeline_stage = 1;
  }
}

// Fired when the pipeline enters or exits a processing stage.
message PipelineStageEvent {
  string stage_name = 1;

  enum StageKind {
    STAGE_KIND_UNSPECIFIED = 0;
    PARSER = 1;
    CONTROL = 2;
    DEPARSER = 3;
  }
  StageKind stage_kind = 2;

  enum Direction {
    DIRECTION_UNSPECIFIED = 0;
    ENTER = 1;
    EXIT = 2;
  }
  Direction direction = 3;
}
```

proto/packet_trace.proto
grg: @matthewtlam I'm just thinking ahead/thinking about how other models could adopt the same trace architecture 🙂 It's something that doesn't need to be solved now.

BTW, we do have digest generation in psa.p4, but according to the comment, it's a digest stream to the control plane. So don't currently have digest packets popping into existence and flowing through the pipeline:
```
// BEGIN:Digest_extern
extern Digest<T> {
  Digest();                       /// define a digest stream to the control plane
  void pack(in T data);           /// emit data into the stream
}
// END:Digest_extern
```

proto/packet_trace.proto
grg: Sure. Consider a P4 program that supports IPv4 options, with the header defined below. It would be helpful to see what size variable-length header field extraction the model performs. (Maybe I messed up the length calculation in the P4 code 🙂 )

```
header Ipv4_t {
    bit <4> ver;
    bit <4> ihl;
    ...
    bit <32> dst;
}

header Ipv4_Opt_t {
    varbit <320> options;
}

...
    // Parser state
    state parse_ipv4 {
        pkt.extract(hdr.ipv4);
        transition select (hdr.ipv4.ihl) {
            // Ignoring invalid values of 0-4
            5: accept;
            default: parse_ipv4_options;
        }
    }

    state parse_ipv4_options {
        // Didn't carefully verify the calculation below :)
        pkt.extract(hdr.ipv4_options, (hdr.ipv4.ihl - 5) * 32);
        transition accept;
    }
```

proto/packet_trace.proto
grg: I'm okay with enter + exit.

In what scenario would the `DIRECTION_UNSPECIFIED` be used?

proto/packet_trace.proto
grg: As long as there's currently a mechanism to go from packet ID to packet content/dump, I'm okay skipping the packet content here.

Can wait and see through usage whether there is any benefit to including packet content in the protobuf message.

proto/packet_trace.proto
grg: Agree that it would be useful to have parameter information.

@c8ef : What is your intention with the two events (`TableLookupEvent` and `ActionExecutionEvent`)? Do you expect `ActionExecutionEvent` to only be used when an action is invoked directly from inside an `apply` block? Or do you expect that an action invoked from a table would be encoded as `TableLookupEvent` followed by an `ActionExecutionEvent`?

An alternative to consider is to remove `action_name` from `TableLookupEvent` and instead have `TableLookupEvent` include a `ActionExecutionEvent` field.

grg:
@c8ef I'm okay with you pursuing a richer trace format up front, but I also have no objection if you want to start with a simpler format while you work on the event logger refactor. (If you go with the simpler format, will you just ignore replication at this point?)

c8ef:
> @c8ef I'm okay with you pursuing a richer trace format up front, but I also have no objection if you want to start with a simpler format while you work on the event logger refactor. (If you go with the simpler format, will you just ignore replication at this point?)

@grg

I think I’ll defer replication for now. The exploration during the design phase and in this PR has been really valuable - beyond just proposing the format, I also investigated how each trace event should be implemented in BMv2. Once I verified that nearly all events fit within the design, I opened this PR.

However, as we incrementally enable this feature, I prefer starting with a minimal, e2e workable solution so it's easier to reach consensus, and then rapidly iterating to enrich the format.

BTW, what do you think of the simplified format above? Currently, the only remaining review comment is regarding the lack of references to `STAGE_KIND_UNSPECIFIED` and `DIRECTION_UNSPECIFIED`. Since I couldn't find any usage of them in the 4ward upstream repository, I think we can safely remove those two.

grg:
@c8ef That sounds good to me.

> BTW, what do you think of the simplified format above? Currently, the only remaining review comment is regarding the lack of references to `STAGE_KIND_UNSPECIFIED` and `DIRECTION_UNSPECIFIED`. Since I couldn't find any usage of them in the 4ward upstream repository, I think we can safely remove those two.

An argument in favor of keeping this is that they serve as useful "uninitialized" indicators. If a message instance is created and all fields are initialized to 0, then all the enums will be at their UNSPECIFIED value.

If you do use them for this purpose and never expect an UNSPECIFIED to appear in a valid message, then it would be good to include a comment indicating this. (E.g., "UNSPECIFIED indicates uninitialized; should never appear in a valid trace")

Keep up the good work!

proto/packet_trace.proto
matthewtlam: As discussed with @qobilidop there doesn't seem to be an obvious way to encode this unless you have any ideas though it would be super helpful if we could

proto/packet_trace.proto
matthewtlam: This is a parameter that is specified depending on the architecture, so it will be better to add this parameter for full granularity

proto/packet_trace.proto
matthewtlam: Thanks for clarifying, that makes sense. WOuld be good at add @c8ef

proto/packet_trace.proto
matthewtlam: Specifically https://p4.org/wp-content/uploads/sites/53/p4-spec/docs/PSA-v1.2.html#sec-clone

proto/packet_trace.proto
c8ef: https://github.com/p4lang/behavioral-model/blob/bc1a6e43145a0c523d1bb7c813deac96e387a8e3/src/bm_sim/parser.cpp#L1135-L1155

```cpp
  const ParseState *next_state = init_state;
  std::string prev_state_name;
  size_t bytes_parsed = 0;
  while (next_state) {
    BMLOG_DEBUG_PKT(*pkt, "Parser '{}' entering state '{}'",
                    get_name(), next_state->get_name());
    // Record parser state transition for tracing.
    BMELOG(parser_transition, *pkt, get_name(),
           (next_state == init_state) ? "" : prev_state_name,
           next_state->get_name());
    try {
      next_state = (*next_state)(pkt, data, &bytes_parsed);
    } catch (const parser_exception &e) {
      auto error_code = e.get(*error_codes);
      BMLOG_ERROR_PKT(*pkt, "Exception while parsing: {}",
                      error_codes->to_name(error_code));
      pkt->set_error_code(error_code);
      break;
    }
    BMLOG_TRACE_PKT(*pkt, "Bytes parsed: {}", bytes_parsed);
    prev_state_name = next_state ? next_state->get_name() : "";
  }
```

This is my initial plan for the encoding, though I'm not entirely sure if it's valid.

c8ef:
Closing this as the superseded PR is merged. If I touch the relevent part in the future I will quote the review comments here.

PR #1412, opened 2026-06-27
Fix CMake parity with Autotools install targets and CTest registrations

Devansh-567:
## The Problem
While the repository's CI infrastructure relies significantly on the CMake build system, the logic was lagging behind the Autotools configuration in key functional areas:
- **Missing Targets:** The `bmp4apps` shared wrapper library (built under Autotools when `WITH_NANOMSG` is enabled) was completely commented out/omitted from CMake configuration generation.
- **Broken CTest Execution:** Test macros in target locations were registering tests via `add_test(NAME ... COMMAND ${TEST})`. Because the executables were internally renamed using `OUTPUT_NAME`, CTest would look for the literal test output filename rather than invoking the actual CMake binary target, causing local or granular test configuration mismatches.
- **Outdated Documentation:** The `README.md` explicitly designated the CMake build system as "Experimental" and documented the absence of `bmp4apps` as a standing system limitation.

## How the Problem Was Found
A localized structural code review comparing `src/Makefile.am` with `src/CMakeLists.txt` exposed the divergence in the `bmp4apps` target installation surface. Further static tracing of test setups inside `targets/simple_switch/tests/CMakeLists.txt` and `targets/simple_switch_grpc/tests/CMakeLists.txt` revealed a broken translation layer where renamed test executables were not being cleanly resolved as targets by the `add_test` command loop.

## The Solution
- **Restored `bmp4apps` Parity:** Re-implemented the conditional compilation of `bmp4apps` inside `src/CMakeLists.txt` matching the `WITH_NANOMSG` flag. Utilized a generation rule via `file(WRITE ...)` inside `CMAKE_CURRENT_BINARY_DIR` to compile a minimal translation unit without forcing an untracked source file into the core repository tree.
- **Fixed CTest Registrations:** Adjusted the test command generation macros to explicitly bind to their internal target aliases (`ss_test_${TEST}` and `ssg_test_${TEST}`) rather than output-file naming conventions. 
- **Documentation Refinement:** Re-classified the CMake workflow header inside `README.md` out of "Experimental" status and purged mentions of the missing library limitations.

## What Was Tested
- Executed `git diff --check` and `git status` locally to ensure an entirely clear patch surface free of unintended whitespace, dangling artifacts, or untracked code pollution.
- Standard cross-platform file encoding consistency was verified across layout modifications.
- *Note on Local Testing Environments:* Core architecture layout adjustments were validated statically. Due to Linux-specific dependency layouts native to BMv2, the compiled behavior relies on the project's robust GitHub Actions CI pipeline matrices for multi-distribution environment verification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,15 +158,13 @@ sudo dpkg -i /path/to/package.deb
     The new bmv2 debugger can be enabled by passing `--enable-debugger` to
     `configure`.
 
-    #### Experimental: cmake-based build
+    #### CMake-based build
 
-    A cmake-based build process was recently added to the project. Differences
+    A CMake-based build process is available and covered by CI. Differences
     from the autoconf-based system are:
-    - Static vs shared libraries: the cmake build produces a single version of
-      each library (the choice depnds upon the library), while the autoconf
+    - Static vs shared libraries: the CMake build produces a single version of
+      each library (the choice depends upon the library), while the autoconf
       build produces static and shared versions for every library.
-    - bmp4apps library: the cmake build system does produce the bmp4apps
-      library.
 
     ```console
     # Create a build directory
@@ -256,7 +254,7 @@ To run the unit tests, simply do:
 **If you get a nanomsg error when running the tests (make check), try running
   them as sudo**
 
-### cmake-based build
+### CMake-based build
 To run the unit tests, simply do from the build directory:
 
     ctest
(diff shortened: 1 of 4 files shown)


jafingerhut:
@grg If you have time an interest to give this a review, that would be great.  If not, no worries.

fruffy (approved):

grg:
Sorry, missed this request earlier. But I'm happy with it based on quick look just now 🙂

PR #1414, opened 2026-06-29
fix(pdfixed): validate dev_id bounds in notification APIs

### LINKED ISSUE #1413: Bounds validation for dev_id in pd_notifications_add_device() and pd_notifications_remove_device()
## Summary

In looking at `pdfixed/src/pd_notifications.cpp`, I saw that both

- `pd_notifications_add_device()`
- `pd_notifications_remove_device()`

access the `listeners` array using the `dev_id` parameter directly without bounds checking.

```cpp
assert(!listeners[dev_id]);
```

and

```cpp
assert(listeners[dev_id]);
```

`listeners` has a fixed size (`NUM_DEVICES`) while the external API provides `int dev_id` to these functions.

## Question

Is the caller of these functions contractually obligated to ensure that the `dev_id` is always in the range of `[0, NUM_DEVICES - 1]` before calling these functions?

I looked around the codebase but couldn't find an easy way of verifying that this check is being done before this array is accessed.

If no such validation happens, then it seems that invalid values (e.g., out of valid device range) might cause out-of-bounds access before any assertions or error checks.

## Possible improvement

If these functions are supposed to be robust, then a simple range check before accessing the array will prevent UB and also handle invalid input in a graceful way.

Before preparing a patch, I wanted to know if this validation was being done somewhere else on purpose or if I should do it here.

Thanks!

sanket-jadhav-cse:
## Summary

Bounds checks added to `dev_id` in:

* `pd_notifications_add_device()`
* `pd_notifications_remove_device()`

Both functions used `listeners[dev_id]` without checking if `dev_id` is in bounds, which might result in array access out of bounds.

The patch adds the check for `dev_id` and returns an error if `dev_id` is invalid.

Fixes #1413

## Testing

* Built successfully using:

  ```
  make -j$(nproc)
  ```

* No format errors according to `git diff --check`.

* Failed to run `make check` because of existing problem with the environment (could not find `libruntimestubs.la`).

### DIFF
--- a/pdfixed/src/pd_notifications.cpp
+++ b/pdfixed/src/pd_notifications.cpp
@@ -144,6 +144,10 @@ NotificationsListener *listeners[NUM_DEVICES];
 int pd_notifications_add_device(int dev_id, const char *notifications_addr,
                                 NotificationCb ageing_cb,
                                 NotificationCb learning_cb) {
+  if (dev_id < 0 || dev_id >= NUM_DEVICES) {
+    std::cerr << "Invalid device id " << dev_id << "\n";
+    return -1;
+  }
   assert(!listeners[dev_id]);
   listeners[dev_id] = new NotificationsListener(dev_id, notifications_addr);
   listeners[dev_id]->register_ageing_cb(ageing_cb);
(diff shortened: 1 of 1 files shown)


jafingerhut (approved):
LGTM

jafingerhut:
I am not aware of the reason that these functions were created, but as far as I can tell there are no callers of these functions anywhere in the behavioral-model or PI repositories, and so probably not in any other p4lang repositories, either.  Still, reasonable to be a bit safer in their implementation, in case someone does use them.

sanket-jadhav-cse:
Thanks for the feedback and for merging this!

It was great to learn more about these functions. Even though they are not used at the moment, I agree that adding this defensive bounds check makes the implementation a bit safer.
Thanks again!

PR #1415, opened 2026-07-01
fix(pdfixed): validate device ID bounds in PRE and mirroring APIs

sanket-jadhav-cse:
## Summary

This change adds defensive device ID bounds validation to the PRE and mirroring APIs before accessing the underlying `clients[]` array.

The affected APIs accept an `int` device identifier and use it to access `clients[dev_id]` through `PdConnMgr::get()` without verifying that the device ID is within the valid range.

This patch adds bounds checks to reject invalid device IDs and return an error before the array access occurs.

The validation pattern follows the approach recently introduced for the notification APIs in PR #1414.

### PRE APIs

- `p4_pd_mc_mgrp_create()`
- `p4_pd_mc_mgrp_destroy()`
- `p4_pd_mc_node_create()`
- `p4_pd_mc_associate_node()`
- `p4_pd_mc_node_destroy()`
- `p4_pd_mc_node_update()`
- `p4_pd_mc_set_lag_membership()`

### Mirroring APIs

- `p4_pd_mirror_session_create()`
- `p4_pd_mirror_session_update()`
- `p4_pd_mirror_session_delete()`

## Testing

Built successfully using:

```bash
make -j$(nproc)
```

Verified:

- `git diff --check`
- Valid device IDs continue to behave as expected.
- Invalid device IDs (`-1`, `500`, and `1000`) now return an error before accessing the underlying `clients[]` array.

### DIFF
--- a/pdfixed/src/pd_mirroring.cpp
+++ b/pdfixed/src/pd_mirroring.cpp
@@ -52,6 +52,10 @@ int p4_pd_mirror_session_create(p4_pd_sess_hdl_t shdl,
   (void) shdl; (void) type; (void) dir; (void) max_pkt_len; (void) cos;
   (void) c2c; (void) extract_len; (void) timeout_usec; (void) int_hdr;
   (void) int_hdr_len;
+  if (dev_tgt.device_id < 0 || dev_tgt.device_id >= PD_MAX_DEVICES) {
+    std::cerr << "Invalid device id " << dev_tgt.device_id << "\n";
+    return 1;
+  }
   return client(dev_tgt.device_id).c->mirroring_mapping_add(id, egr_port);
 }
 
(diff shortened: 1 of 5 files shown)


pdfixed/src/pd_pre.cpp
```diff
@@ -17,6 +17,8 @@
 
 #include <string>
 
+#define NUM_DEVICES 256
```
fruffy: Where is this constant coming from anyway?

pdfixed/src/pd_mirroring.cpp
```diff
@@ -15,6 +15,8 @@
 
 #include <vector>
 
+#define NUM_DEVICES 256
```
fruffy: We should move this into a shared header instead of redeclaring for each compilation unit.

sanket-jadhav-cse: Thanks, for your feedback. I will move the device-count constant to a shared header file.

I will then update all the parts of the code to use this shared definition.

This way they will use the definition instead of having local defines.

sanket-jadhav-cse:
Updated. The PD_MAX_DEVICES is now defined in one place, which's in pd_common.h. This definition is used by other parts, including pd_conn_mgr.h the PRE code, the mirroring code and the notification code. I removed all the local NUM_DEVICES definitions. I checked that the change is correct, by running git diff --check and by building with make -C pdfixed -j$(nproc).

fruffy (approved):

PR #1416, opened 2026-07-02
Deterministic test_meters and test_tables

chreekat:
I hit a non-deterministic test failure when testing the Debian package build. The build environment is in a VM, which probably made it easier to get weird timings. The fix is to get rid of wall-clock dependencies.

I actually found a bunch of wall-clock dependencies. This PR just addresses the easiest two of them.


chreekat:
As the commit message in https://github.com/p4lang/behavioral-model/pull/1416/commits/d02a0a99d5e766ab398e7852ce2d4809024d1abf notes, although it looks like I removed a test, it was already duplicating a different test.

jafingerhut (approved):

PR #1418, opened 2026-07-04
Implement Protobuf-based structured packet tracing for BMv2

c8ef:
Alternative to https://github.com/p4lang/behavioral-model/pull/1405.

Reason: https://github.com/p4lang/behavioral-model/pull/1405#issuecomment-4865946188


c8ef:
@matthewtlam:

Updated the comments for `set_output_dir` and `flush_trace` to clarify thread safety.

@fruffy:

The protobuf package is now opt-out by default for both CMake and Make builds.

@grg:

It looks like the UNSPECIFIED state is required to comply with protobuf conventions.

```
bazel-out/k8-fastbuild/bin/proto/_virtual_imports/packet_trace_proto/packet_trace.proto:46:14: The first enum value must be zero for open enums.
bazel-out/k8-fastbuild/bin/proto/_virtual_imports/packet_trace_proto/packet_trace.proto:53:13: The first enum value must be zero for open enums.
```

I believe I've addressed all previous review comments relevant to this diff, but please let me know if I missed anything. Thanks!

cc: @jafingerhut @qobilidop

proto/packet_trace.proto
```diff
@@ -0,0 +1,59 @@
...
+// Recursive trace structure: a sequence of events followed by an outcome.
+message TraceTree {
+  repeated TraceEvent events = 1;
+}
+
+// A single event in the packet trace.
+message TraceEvent {
+  oneof event {
+    // Pipeline lifecycle.
+    PipelineStageEvent pipeline_stage = 1;
```
matthewtlam, line 44: Let's start with just Transmit and Drop for the outcome cases since they are the most basic cases to handle

matthewtlam (commented):
Let's add the last part of the proto that we need and we should be good for submission

proto/packet_trace.proto
c8ef: Added.

matthewtlam (approved):
LGTM, thanks!

matthewtlam:
@fruffy or @jafingerhut could you please take a look

fruffy:
Only had time to do a quick scan. LGTM. Ideally, update the PR title to be more descriptive

c8ef:
> Only had time to do a quick scan. LGTM. Ideally, update the PR title to be more descriptive

Done.

c8ef:
Gentle ping for final approval and merging~
@fruffy @jafingerhut @grg

configure.ac
```diff
@@ -188,6 +194,10 @@ AS_IF([test "$want_nanomsg" = yes], [
     AC_DEFINE([NANOMSG_ON], [], [Enable Nanomsg support])
 ])
 
+AS_IF([test "$want_packet_trace" = yes], [
+    AC_DEFINE([PACKET_TRACE_ON], [], [Enable structured packet trace support])
```
jafingerhut, line 198: Sorry, I'm not that familiar with GNU Autotools/automake/etc.  Should this name be PACKET_TRACE, or BM_PACKET_TRACE?  Have you tried using Autotools to build behavioral-model with these changes to verify that with them you can include compiling the new C++ code you are adding?

c8ef: First, the new code is covered by CI across all three build systems. Second, a specific line of code adds the prefix.

https://github.com/p4lang/behavioral-model/blob/37b365f30cbd2b5d9c2c712631073fbf03ac7cb2/configure.ac#L363

src/bm_sim/packet_tracer.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
+  if (!enabled_) return;
+
+  std::string txtpb;
+  google::protobuf::TextFormat::PrintToString(trace, &txtpb);
+
+  auto path = std::filesystem::path(output_dir_) /
+              ("trace_" + std::to_string(trace.input().packet_id()) + ".txtpb");
```
grg, line 68: Too bad we don't know the number of packets in advance. Otherwise we could pad to zeros appropriate to the number of packets (e.g., `trace_001` ... `trace_100`).

grg (approved):
Looks good to me. Thanks for all your work on it.

src/bm_sim/packet_tracer.cpp
jafingerhut: If someone really found that advantageous, it seems like a Python log post-processing program could do it for you.  Doing that inside of behavioral-model code during packet processing time doesn't sound feasible to me.

jafingerhut: Especially if you start thinking about cases where one copy of a multicast-replicated packet is itself multicast-replicated, and one of those copies is itself multicast-replicated, etc. (which is possible as long as the architecture support recycling packets, which most do).  Same goes for any other operation that causes packets to be replicated, or packets in the log to be replicated (if this logging in future supports producing the traces for all alternatives of an action selector, instead of only one).

fruffy (approved):

src/bm_sim/packet_tracer.cpp
c8ef: I usually find this helpful: https://ss64.com/bash/rename.html.

c8ef:
Hi reviewers, if this PR doesn't need any further changes, could you help merge it? Thanks!

PR #1419, opened 2026-07-07
fix(bm_sim): fix every-other-sweep ageing re-notification oscillation

Devansh-567:
## Problem
When table entries age out and remain in an aged state across multiple sequential sweeps, `AgeingMonitor::do_sweep()` repeatedly re-triggers aging notifications every alternate sweep cycle instead of suppressing them.

This state-tracking oscillation occurs due to two flaws in the loop logic:
1. `prev_sweep_entries` is prematurely cleared on empty sweeps via an aggressive short-circuit check (`if (entries_tmp.empty())`).
2. The historical tracking set is populated only from the newly generated notification delta (`entries`), rather than reflecting the absolute historical snapshot of all currently-aged entries.

## Solution
- Removed the early-exit `entries_tmp.empty()` short-circuit logic entirely so that empty sweeps naturally clear out old state.
- Modified the tracking synchronization stage to cleanly clear and rebuild `prev_sweep_entries` using the entire snapshot of `entries_tmp` right before it is cleared. 

This keeps the tracking set perfectly synchronized with the absolute state of the engine across consecutive sweeps, ensuring previously-notified entries are not treated as "new" again.

## Changes
- `src/bm_sim/ageing.cpp`: Fixed tracking loop alignment inside `AgeingMonitor::do_sweep()`.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -61,4 +61,5 @@ jobs:
         run: bazel build //targets/simple_switch_grpc:simple_switch_grpc
 
       - name: Test
-        run: bazel test //targets/simple_switch_grpc/tests/...
+        run: bazel test //targets/simple_switch_grpc/tests/... --test_output=errors
+
--- a/src/bm_sim/ageing.cpp
+++ b/src/bm_sim/ageing.cpp
@@ -163,10 +163,6 @@ AgeingMonitor::do_sweep() {
     const MatchTableAbstract *t = data.table;
     auto &prev_sweep_entries = data.prev_sweep_entries;
     t->sweep_entries(&entries_tmp);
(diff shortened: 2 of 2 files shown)


Devansh-567:
Closing this PR, the reported behavior is not a bug.
The "every-other-sweep" re-notification pattern in AgeingMonitor::do_sweep() is intentional, documented, and covered by existing tests. See the TODO in targets/simple_switch_grpc/tests/test_idle_timeout.cpp (lines 172–179): the sweeper deliberately skips notifying on the immediate next sweep but will notify again on the sweep after that if the entry remains untouched. This is a known deviation from strict P4Runtime spec (which calls for notification after one more TTL, not a fixed sweep interval), but it's the current, intended behavior.

PR #1421, opened 2026-07-08
fix: replace bare except clauses across CLI tools

Devansh-567:
## Summary

Bare `except:` clauses across four tool files catch `KeyboardInterrupt` and `SystemExit` in addition to real errors, this can make Ctrl-C hang or behave oddly, and hides real bugs during debugging.

## Fix

- `tools/bmpy_utils.py` (2 instances): Thrift RPC calls to the switch -> `except Exception:`
- `tools/nanomsg_client.py` (2 instances): message-class lookup/construction -> `except Exception:`; `import bmpy_utils` guard -> `except ImportError:`
- `tools/runtime_CLI.py` (51 instances): `try: import ... except:` guards -> `except ImportError:`; all other sites (int/JSON parsing, address parsing) -> `except Exception:`
- `tools/p4dbg.py` (9 instances): same pattern, import guards -> `except ImportError:`, everything else -> `except Exception:`

### DIFF
--- a/tools/bmpy_utils.py
+++ b/tools/bmpy_utils.py
@@ -32,7 +32,7 @@ def check_JSON_md5(client, json_src, out=sys.stdout):
 
     try:
         bm_md5sum = client.bm_get_config_md5()
-    except:
+    except Exception:
         my_print("Error when requesting config md5 sum from switch\n")
         sys.exit(1)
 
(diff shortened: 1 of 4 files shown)


fruffy (approved):

fruffy:
You could consider running `ruff check` here to see what you can find. I am sure we can do many more improvements.

PR #1425, opened 2026-07-16
Fix TypeError in JSON MD5 hex formatting for Python 3

Devansh-567:
## Summary
 
Fixes a crash in the CLI's config-mismatch check on Python 3. No behavior changes on the success path, this only fixes what happens when the mismatch warning itself needs to print.
 
## Problem
 
`check_JSON_md5()` in `tools/bmpy_utils.py` builds a hex string from the two md5 sums like this:
 
```python
bm_md5sum_str = "".join("{:02x}".format(ord(c)) for c in bm_md5sum)
md5sum_str = "".join("{:02x}".format(ord(c)) for c in md5sum)
```
 
`bm_get_config_md5()` is declared `binary` in `thrift_src/standard.thrift`, so the Thrift client returns `bytes` on Python 3. `hashlib`'s `.digest()` also returns `bytes`. Iterating over a `bytes` object in Python 3 already yields `int` values (0-255) rather than single-character strings, so calling `ord()` on each element raises:
 
```
TypeError: ord() expected string of length 1, but int found
```
 
This means the check crashes at exactly the moment it's supposed to warn the user, whenever the loaded JSON's md5 doesn't match the switch's running config.
 
## Fix
 
Drop the now-unnecessary `ord()` call and format the int directly:
 
```python
bm_md5sum_str = "".join("{:02x}".format(c) for c in bm_md5sum)
md5sum_str = "".join("{:02x}".format(c) for c in md5sum)
```


jafingerhut:
Have you tried a local change that always executes the code you are modifying, even when the two md5 variables have equal values, and validated that there is no exception, and the strings printed look like they have the correct values?

Note: I am _not_ recommending that there be a test executed in CI/CD that exercises the error condition (unless that is trivial to create) for such an error case like this.  Mainly that you have high confidence that the modified code is correct, by actually executing it in the context of the place where it would run.

Devansh-567:
> Have you tried a local change that always executes the code you are modifying, even when the two md5 variables have equal values, and validated that there is no exception, and the strings printed look like they have the correct values?
> 
> Note: I am _not_ recommending that there be a test executed in CI/CD that exercises the error condition (unless that is trivial to create) for such an error case like this. Mainly that you have high confidence that the modified code is correct, by actually executing it in the context of the place where it would run.

@jafingerhut Local test reproduced the original TypeError: ord() expected string of length 1, but int found with the old code against real bytes input (matching what hashlib.digest() and the Thrift binary field return on Py3). Then tested the fixed version against both a mismatched md5 pair (the actual warning path) and a matched pair, no exceptions in either case, and both printed correct 32-character hex strings. Confirmed the fix resolves the crash without changing the output format.

jafingerhut (approved):

jafingerhut:
Great.  Thanks for checking and reporting the results.

PR #1428, opened 2026-07-20
Refactor EventLogger to a multi-observer registry (NFC)

c8ef:
Previously the `BMELOG` macro dispatched every event directly to a single hardcoded `EventLogger`. This refactor introduces an observer abstraction so multiple backends (the nanomsg `EventLogger`, and the upcoming Protobuf `PacketTracer`) can receive the same events.

### DIFF
--- a/include/bm/bm_sim/event_observer.h
+++ b/include/bm/bm_sim/event_observer.h
@@ -19,10 +19,9 @@
 
 namespace bm {
 
-// Forward declarations of P4 object classes. This is ugly, but:
-// 1) I don't have to worry about circular dependencies
-// 2) If I decide to switch from id to name for msgs, I won't have to modify the
-// EventObserverIface
+// Forward declarations of P4 object classes. Don't need their full definitions
+// because this file only uses references to them and doesn't require their
+// implementation/layout details.
 class ActionFn;
 class Checksum;
 class Conditional;
@@ -107,7 +106,12 @@ class EventObserverRegistry {
 }  // namespace bm
 
 //! Dispatch an event to all registered observers.
-//! Usage: BMELOG(parser_start, *pkt, *this);
+//! For example:
+//! @code
+//! BMELOG(packet_in, packet);
+//! // packet processing
+//! BMELOG(packet_out, packet);
+//! @endcode
 #if defined(BM_ELOG_ON) || defined(BM_PACKET_TRACE_ON)
 #define BMELOG(fn, ...)                                                 \
   do {                                                                  \
--- a/include/bm/bm_sim/event_observer.h
+++ b/include/bm/bm_sim/event_observer.h
@@ -1,0 +1,122 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Yuao Ma
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! @file event_observer.h
+
+#ifndef BM_BM_SIM_EVENT_OBSERVER_H_
+#define BM_BM_SIM_EVENT_OBSERVER_H_
+
+#include <bm/config.h>
+
+#include <vector>
+
+#include "device_id.h"
+#include "phv_forward.h"
+#include "transport.h"
+
+namespace bm {
+
+// Forward declarations of P4 object classes. This is ugly, but:
+// 1) I don't have to worry about circular dependencies
+// 2) If I decide to switch from id to name for msgs, I won't have to modify the
+// EventObserverIface
+class ActionFn;
+class Checksum;
+class Conditional;
+class Deparser;
+class MatchEntry;
+class MatchTableAbstract;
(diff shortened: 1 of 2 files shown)


c8ef:
cc @matthewtlam @fruffy @jafingerhut @grg

include/bm/bm_sim/event_observer.h
```diff
@@ -0,0 +1,122 @@
...
+#include <vector>
+
+#include "device_id.h"
+#include "phv_forward.h"
+#include "transport.h"
+
+namespace bm {
+
+// Forward declarations of P4 object classes. This is ugly, but:
+// 1) I don't have to worry about circular dependencies
```
fruffy: Usually, circular dependencies can be solved with some reasonable restructuring or abstractions. Where is the problem here?

grg: Forward declarations are reasonable to me here, and I see you've just moved them + their comments from the other file.

However, I would recommend changing the comment to something like:
"Forward declarations of P4 object classes. Don't need their full definitions because this file only uses references to them and doesn't require their implementation/layout details."

grg (commented):
I have no objection to what you're trying to do here (introduce multiple event observers).

include/bm/bm_sim/event_observer.h
matthewtlam: +1 to what @grg and @fruffy mentioned

matthewtlam:
This refactor at the high level is a good idea. Let's try to get the requested changes completed

include/bm/bm_sim/event_observer.h
c8ef: Fixed! Also added a minor tweak to the `BMELOG` comment.

grg (approved):
LGTM!

include/bm/bm_sim/event_observer.h
```diff
@@ -0,0 +1,126 @@
...
+
+using entry_handle_t = uint32_t;
+
+//! Abstract observer interface for pipeline events.
+//! Each backend (nanomsg, trace, etc.) implements this interface.
+class EventObserverIface {
+ public:
+  virtual ~EventObserverIface() = default;
+
+  // we need the ingress / egress ports, but they are part of the Packet
```
matthewtlam, line 44: NIT. ! before the comment

include/bm/bm_sim/event_observer.h
```diff
@@ -0,0 +1,126 @@
...
+
+#include <vector>
+
+#include "device_id.h"
+#include "phv_forward.h"
+#include "transport.h"
+
+namespace bm {
+
+// Forward declarations of P4 object classes. Don't need their full definitions
```
matthewtlam, line 22: NIT: ! before comments here for consistency

matthewtlam (approved):
LGTM, left comments for minor nits

include/bm/bm_sim/event_observer.h
c8ef: The `!` is used for doxygen. The `!` usage in `event_observer.h` is the same as the old counterpart. My understanding is this comment is supposed to appear in source code, not in the generated doc.

include/bm/bm_sim/event_observer.h
c8ef: Same reason as above.

grg: The ! flags it as a something to include in the Doxygen output, which possibly makes sense here.

include/bm/bm_sim/event_observer.h
grg: I tent to agree with @c8ef on this one that it doesn't need to be in the Doxygen output. (If we do want it in the Doxygen output, then the same comment should probably also appear directly above the `packet_out` declaration.)

matthewtlam: Ack SG

include/bm/bm_sim/event_observer.h
c8ef: IIUC if added the `!`, the doxygen will add the doc to the _next documentable element_, which is the forward declaration of the ActionFn. But if you look at the actual definition of `ActionFn`, you will see no doxygen comment. The doxygen config has `HIDE_UNDOC_CLASSES = YES`, which means that if we add the `!` here, we will un-hide the `ActionFn` class, and the newly added doc only say something about forward declaration, not really relevent to the class itself. So I think we may not want to add the `!` here.

matthewtlam: Ack. SGTM

c8ef:
Gentle ping for final approval and merging~
@fruffy @jafingerhut @grg

grg (approved):

fruffy (approved):

PR #1431, opened 2026-07-24
use get_data_size in prsr short packet checks

grg:
Replace `get_ingress_length()` with `get_data_size()` in parser checks to verify that there is sufficient packet data for the operation.

To provide a little more detail on this change:

1. `get_data_size()` returns the minimum of the _current_ data size and the truncated length. If a device truncates the packet prior to ingress parsing, or if the egress parser is running after ingress deparser changes the packet size, then `get_data_size()` will return a different value to `get_ingress_length()`

2. `get_data_size()` was already used in two places:
   * [`ParserOpSet<ParserLookAhead>::operator()(...)`](https://github.com/p4lang/behavioral-model/blob/b79ffdbcecb3b7feaf3afc6ee1349b6d5946d7d9/src/bm_sim/parser.cpp#L86)
   * [`check_enough_data_for_extract(...)`](https://github.com/p4lang/behavioral-model/blob/b79ffdbcecb3b7feaf3afc6ee1349b6d5946d7d9/src/bm_sim/parser.cpp#L229)

### DIFF
--- a/src/bm_sim/parser.cpp
+++ b/src/bm_sim/parser.cpp
@@ -257,7 +257,7 @@ void extract_VL(Header *hdr,
   // get_nbytes_packet counts the VL field in the header as 0 bits
   auto bytes_to_extract = static_cast<size_t>(
       hdr->get_nbytes_packet() + nbits / 8);
-  if (pkt->get_ingress_length() - *bytes_parsed < bytes_to_extract)
+  if (pkt->get_data_size() - *bytes_parsed < bytes_to_extract)
     throw parser_exception_core(ErrorCodeMap::Core::PacketTooShort);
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
Are there any known scenarios where the return values of these two functions differ, that you are aware of?  (I am not already familiar with their existing behavior.)

grg:
> Are there any known scenarios where the return values of these two functions differ, that you are aware of? (I am not already familiar with their existing behavior.)

I've added more detail in the PR description. Yes, there are scenarios where the two functions return different values.

fruffy (approved):

jafingerhut (approved):

PR #1433, opened 2026-07-28
Minimal e2e implementation of packet trace.

c8ef:
With this patch we will have a minimal and workable implementation of packet trace with testcases.


c8ef:
cc @matthewtlam @fruffy @jafingerhut @grg

fruffy:
How could we test this? E.g., for the compiler we have golden tests to ensure that the output matches our expectations. We could also use gtests and pattern match, but that is harder to maintain and review and might not capture the full picture of a trace.

c8ef:
> How could we test this? E.g., for the compiler we have golden tests to ensure that the output matches our expectations. We could also use gtests and pattern match, but that is harder to maintain and review and might not capture the full picture of a trace.

During development I have a local smoke test script(which generates the trace in the pr description). Though I haven't think very thoroughly on how to test this. Since it is using existing infrastructure I hope we could reuse the existing nanomsg backend testing, but not sure if that exists.

fruffy:
> > How could we test this? E.g., for the compiler we have golden tests to ensure that the output matches our expectations. We could also use gtests and pattern match, but that is harder to maintain and review and might not capture the full picture of a trace.
> 
> During development I have a local smoke test script(which generates the trace in the pr description). Though I haven't think very thoroughly on how to test this. Since it is using existing infrastructure I hope we could reuse the existing nanomsg backend testing, but not sure if that exists.

Basically you want avoid that this feature slowly bitrots. Ideally, we have infra for E2E tests that run BMv2, generate a trace and then we compare outputs.

matthewtlam:
@c8ef I think we should be producing golden tests for this and adding appropriate test cases so that we can see if the behavior is as we expect. Ideally, we would iterate over the different P4 pipelines that we have define and send the corresponding input packets through BMv2 and then we output the packet trace in a golden file (.expected file). Since you added the Bazel support, we can do some golden tests. Here are a couple of examples:

This is the corresponding test runner: https://github.com/google/p4-infra/blob/f107159078db6ba189a413d7308f00b4c0d1d9fe/p4_pdpi/references_test_runner.cc

This is a sample expected file: https://github.com/google/p4-infra/blob/f107159078db6ba189a413d7308f00b4c0d1d9fe/p4_pdpi/references.expected

We can also add the golden tests to the Github workflow.

c8ef:
> @c8ef I think we should be producing golden tests for this and adding appropriate test cases so that we can see if the behavior is as we expect. Ideally, we would iterate over the different P4 pipelines that we have define and send the corresponding input packets through BMv2 and then we output the packet trace in a golden file (.expected file). Since you added the Bazel support, we can do some golden tests. Here are a couple of examples:
> 
> This is the corresponding test runner: https://github.com/google/p4-infra/blob/f107159078db6ba189a413d7308f00b4c0d1d9fe/p4_pdpi/references_test_runner.cc
> 
> This is a sample expected file: https://github.com/google/p4-infra/blob/f107159078db6ba189a413d7308f00b4c0d1d9fe/p4_pdpi/references.expected
> 
> We can also add the golden tests to the Github workflow.

@fruffy @matthewtlam 

Thanks for the pointers on the test runners! Following that pattern, I added a similar test runner for BMv2 and the structured packet trace test. Since the test isn't necessarily related to gRPC, I added it to the simple_switch target directory.

src/bm_sim/packet_tracer.cpp
```diff
@@ -75,6 +84,134 @@ void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
...
+  auto ctx = std::make_unique<PacketTraceContext>(packet->get_packet_id(),
+                                                  packet->get_ingress_port());
+  packet->set_trace_ctx(std::move(ctx));
+}
+
+namespace {
+
+// Add a TraceEvent to the packet's trace context (if present).
+void trace_add_event(const Packet& packet, p4::bm::TraceEvent event) {
+  auto* ctx = packet.get_trace_ctx();
```
matthewtlam: Let's explicitly define the type here to make it clear with respect for readability

src/bm_sim/packet_tracer.cpp
```diff
@@ -75,6 +84,134 @@ void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
...
+  if (ctx) {
+    ctx->add_event(std::move(event));
+  }
+}
+
+}  // namespace
+
+void PacketTracer::packet_in(const Packet& packet) { _BM_UNUSED(packet); }
```
matthewtlam, line 106: Could you clarify what the packet_in and packet_out is supposed to do? I assume this is just not yet implemented? Maybe we can leave a comment for all the unimplemented functions (although my preference is to not introduce it until it is used)

src/bm_sim/packet_tracer.cpp
```diff
@@ -75,6 +84,134 @@ void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
...
+
+void PacketTracer::packet_out(const Packet& packet) { _BM_UNUSED(packet); }
+
+void PacketTracer::parser_start(const Packet& packet, const Parser& parser) {
+  p4::bm::TraceEvent ev;
+  auto* ps = ev.mutable_pipeline_stage();
```
matthewtlam: Let's define the type to be explicit

src/bm_sim/packet_tracer.cpp
```diff
@@ -75,6 +84,134 @@ void PacketTracer::flush_trace(const p4::bm::PacketTrace& trace) {
...
+  ps->set_stage_name(parser.get_name());
+  ps->set_stage_kind(p4::bm::PipelineStageEvent::PARSER);
+  ps->set_direction(p4::bm::PipelineStageEvent::ENTER);
+  trace_add_event(packet, std::move(ev));
+}
+
+void PacketTracer::parser_done(const Packet& packet, const Parser& parser) {
+  p4::bm::TraceEvent ev;
+  auto* ps = ev.mutable_pipeline_stage();
```
matthewtlam: Ditto as above

targets/simple_switch/tests/testdata/parser_error.trace
```diff
@@ -0,0 +1,232 @@
...
+P4 pipeline: parser_error.json
+Table programming:
+  (none)
+Input packets:
+  port=1 len=2 bytes=0000
+-- OUTPUT ---------------------------------------------------------------
+Output packets:
+  port=1 len=6 bytes=000000010000
+Packet traces:
+--- trace_1.txtpb
```
matthewtlam: The numbering of the traces should be unique and ideally start at 0 for each packet test. It will be hard to keep track of the correct number of expected outputs for cases like WCMP where we want to make sure that we output the correct number of traces

matthewtlam (commented):
Sending some initial comments. Sorry for the late review, I've been on vacation and haven't had a good chance to take a look at things. Also the PR is big so it will take sometime. Will try to have faster reviews. LOoking good as a start!

targets/simple_switch/tests/testdata/recirc.trace
```diff
@@ -0,0 +1,76 @@
+=========================================================================
+Packet trace test: recirc: recirculate once
+A 1-byte packet with value 0x00 is recirculated by the egress pipeline; the recirculated pass emits a 2-byte packet on the ingress port. The recirculated packet copy currently produces no additional trace events.
+=========================================================================
+-- INPUT ----------------------------------------------------------------
+P4 pipeline: recirc.json
+Table programming:
+  (none)
```
matthewtlam, line 8: For recirculate, we should have table entries such that we expect to see the packet entering the ingress + egress stage at least twice. Perhaps add another test case here to show this property off

targets/simple_switch/tests/test_packet_trace.cpp
```diff
@@ -0,0 +1,482 @@
...
+using bm::MatchKeyParam;
+
+constexpr char kSectionBanner[] =
+    "=========================================================================";
+constexpr char kInputBanner[] =
+    "-- INPUT ----------------------------------------------------------------";
+constexpr char kOutputBanner[] =
+    "-- OUTPUT ---------------------------------------------------------------";
+
+std::string hex_dump(const std::string& bytes) {
```
matthewtlam: Can we leave a comment for what this hex dump is for?

targets/simple_switch/tests/test_packet_trace.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+  bool port_is_up_(port_t port_num) const override {
+    (void)port_num;
+    return true;
+  }
+
+  std::map<port_t, PortInfo> get_port_info_() const override { return {}; }
+};
+
+struct OutputPacket {
```
matthewtlam, line 136: Let's move this to the top as well as the input packet struct

targets/simple_switch/tests/test_packet_trace.cpp
```diff
@@ -0,0 +1,482 @@
...
+  size_t expected_outputs;
+  // Queueing metadata (timestamps) makes some output payloads
+  // non-deterministic; such cases only print output packet sizes.
+  bool dump_output_payload = true;
+};
+
+// Table-programming helpers: perform the operation and print one line
+// describing it. Errors are printed to stdout so that they break the golden
+// diff instead of going unnoticed.
+
```
matthewtlam: Remove empty line

src/bm_sim/packet_tracer.cpp
c8ef: Fixed.

src/bm_sim/packet_tracer.cpp
c8ef: Fixed.

src/bm_sim/packet_tracer.cpp
c8ef: Fixed.

src/bm_sim/packet_tracer.cpp
c8ef: My original plan is that packet_in will leave blank and packet_out can record some events related to transmit. I think after we introduces most of the events we can add some comments about whether the function is left blank or unimplemented.

targets/simple_switch/tests/test_packet_trace.cpp
c8ef: I moved the struct InputPacket, should achieve the same effect.

targets/simple_switch/tests/test_packet_trace.cpp
c8ef: Done.

targets/simple_switch/tests/testdata/parser_error.trace
c8ef: In current design the trace filename closely related to packet_id, which in the parser_error case there are precisely 3 packets here. I think we could use this to test the number logic here, and if there is special case like WCMP, maybe we could limit to one packet, and focus on the single trace?

targets/simple_switch/tests/testdata/recirc.trace
c8ef: Actually it has the default table entry (see the related json file for more detail), so here `(none)` means no manually added table programming. I should make it clearer here.

c8ef: Fixed the trace file description and removed the redundant manually defined table entry in https://github.com/p4lang/behavioral-model/pull/1433/commits/e5ebb0e004a5250e32bcc26a2b4903e22ee2d4a1

src/bm_sim/packet_tracer.cpp
matthewtlam: Sounds good. We can add support in a follow-up PR

targets/simple_switch/tests/test_packet_trace.cpp
matthewtlam: Can we rename this function. Maybe BytestringToFormatedHexString?

targets/simple_switch/tests/test_packet_trace.cpp
matthewtlam: Thanks! Looks good

targets/simple_switch/tests/testdata/parser_error.trace
matthewtlam: I think for WCMP we would not want to only show a single packet nor a single packet trace but show a comprehensive set of packets despite being lengthy as it helps provide all the debug info needed. We should adjust the output of this golden test to account for this

targets/simple_switch/tests/testdata/recirc.trace
matthewtlam: Ack sounds good. Though we should have more comprehensive tests in itself for the recirculate behavior in itself. Perhaps we should open an issue to track this

targets/simple_switch/tests/test_packet_trace.cpp
```diff
@@ -0,0 +1,493 @@
...
+    if (google::protobuf::TextFormat::ParseFromString(ss.str(), &trace)) {
+      traces.emplace(entry.path().filename().string(), ss.str());
+    }
+  }
+  return traces;
+}
+
+// Traces are flushed when the switch destroys a packet, which happens shortly
+// *after* the corresponding output packet (if any) is transmitted, and a
+// single input may produce several traces (e.g. recirculation). So rather
```
matthewtlam: Shouldn't recirculation just add additional events to the same trace rather than make a new trace?

targets/simple_switch/tests/test_packet_trace.cpp
```diff
@@ -0,0 +1,493 @@
...
+  if (!collector->wait_for_packets(test_case.expected_outputs,
+                                   std::chrono::seconds(3))) {
+    std::cout << "ERROR: timed out waiting for output packets\n";
+  }
+  auto traces = wait_for_traces(trace_dir, test_case.inputs.size());
+  auto outputs = collector->take();
+
+  std::cout << kOutputBanner << "\n";
+  std::cout << "Output packets:\n";
+  if (outputs.empty()) std::cout << "  (none)\n";
```
matthewtlam, line 315: May be nice to specify the packets get dropped, but we can handle that in a follow up PR as well

targets/simple_switch/tests/test_packet_trace.cpp
c8ef: Fixed.

targets/simple_switch/tests/test_packet_trace.cpp
c8ef: I think if the packets get dropped in `Packet traces:` section we can clearly see the traces relate to drop. Adding the information here is also acceptable.

targets/simple_switch/tests/test_packet_trace.cpp
c8ef: Fixed.

targets/simple_switch/tests/test_packet_trace.cpp
c8ef: Yes indeed. Fixed here and this function got much simplified.

targets/simple_switch/tests/testdata/parser_error.trace
c8ef: Understood. Now in the latest version, the traces and packet_ids have been renamed, so we can clearly see the expected number of packet or packet_id.

targets/simple_switch/tests/testdata/recirc.trace
c8ef: I think when packet trace relates to recirculate is implemented, current test will be sufficient for the recirculate testing? If not we can add more test then.

.github/workflows/bazel.yml
```diff
@@ -61,4 +61,6 @@ jobs:
         run: bazel build //targets/simple_switch_grpc:simple_switch_grpc
 
       - name: Test
-        run: bazel test //targets/simple_switch_grpc/tests/...
+        run: |
```
fruffy, line 64: My one request is also to add test support for cmake here. Bazel-only is not great for the wider community

c8ef: Since this PR is already huge I suggest this could be done in follow up PRs.

fruffy: I prefer to merge this only when it is complete. This PR only changes 19 files, most of which are test files.

c8ef: Okay. But I'm not familiar with all these build systems so it might take a while.

matthewtlam: @c8ef Any updates on this?

c8ef: Done. Sorry for the late update, a lot of personal things have been piling up recently.

fruffy: Great, thanks for handling this! 

Are the `packet_trace_*` in the CMake build the ones that were added here?

c8ef: Yes. The precise location is here: https://github.com/p4lang/behavioral-model/pull/1433/changes#diff-5cc9afcf8d180cef0c9be368d78afb9aec6724e93f8c276b69cc311deb6b854cR71

fruffy (approved):

fruffy:
Great work!

targets/simple_switch/tests/testdata/parser_error.trace
```diff
@@ -0,0 +1,230 @@
...
+-- INPUT ----------------------------------------------------------------
+P4 pipeline: parser_error.json
+Table programming (runtime only; config-baked defaults not shown):
+  (none)
+Input packets:
+  port=1 len=4 bytes=00000000
+-- OUTPUT ---------------------------------------------------------------
+Output packets:
+  port=1 len=4 bytes=00000000
+Packet traces: 1
```
matthewtlam: Small nit: number of packet trace. Prefer to be explicit

targets/simple_switch/tests/testdata/recirc.trace
matthewtlam: I would strongly prefer another test. Perhaps open an issue cause I don't want it to block submission

targets/simple_switch/tests/test_packet_trace.cpp
matthewtlam: We can add it once we model the drop semantics. No need in this PR

matthewtlam (approved):
Thanks for the great work! Overall LGTM!

targets/simple_switch/tests/testdata/parser_error.trace
matthewtlam: Thanks LGTM

targets/simple_switch/tests/testdata/recirc.trace
c8ef: Opened https://github.com/p4lang/behavioral-model/issues/1445 to track this.

targets/simple_switch/tests/testdata/parser_error.trace
c8ef: Fixed.

matthewtlam:
@grg Friendly bump. Would greatly appreciate it if you could take a look

fruffy:
Merging this.

PR #1435, opened 2026-07-30
Fix unsigned underflow in StackP4_16::pop_front

sanket-jadhav-cse:
## Summary

`StackP4_16::pop_front(size_t num)` computes the loop bound using `size - num`, where `size` is an unsigned value. When `num` is greater than the stack size, this subtraction underflows, resulting in an invalid loop bound and out-of-bounds access.

This change avoids the underflow by skipping the swap loop when `num >= size` and directly invalidating the remaining headers. The existing behavior for valid inputs is preserved.

## Before

* `pop_front()` computed `size - num` unconditionally.
* Calling `pop_front()` with `num > stack size` caused unsigned integer underflow.
* Running the new regression test against the original implementation reproduced the issue with a `std::vector::operator[]` assertion failure.

## After

* The swap loop executes only when `num < size`.
* For `num >= size`, all headers are invalidated directly.
* No unsigned underflow occurs.
* `pop_front()` safely handles requests larger than the stack size.

## Testing

* Added `HeaderStackP4_16Test.PopFrontLargeNum` regression test.
* Added `HeaderStackP4_16Test.PopFrontExactDepth` boundary test.
* Verified that `PopFrontLargeNum` reproduces the failure with the original implementation.
* Verified that both new tests pass with the fix.
* Ran the complete `test_header_stacks` suite successfully.
* Ran the full CTest suite. The only remaining failure was the existing `test_switch` test due to a missing Python `thrift` dependency, which is unrelated to this change.

### DIFF
--- a/src/bm_sim/stacks.cpp
+++ b/src/bm_sim/stacks.cpp
@@ -103,8 +103,11 @@ StackP4_16<T>::pop_front(size_t num) {
   auto size = this->elements.size();
   this->next -= std::min(this->next, num);
   size_t i = 0;
-  for (; i < size - num; i++) {
-    this->elements[i].get().swap_values(&this->elements[i + num].get());
+  // Avoid unsigned underflow when num >= size
+  if (num < size) {
+    for (; i < size - num; i++) {
+      this->elements[i].get().swap_values(&this->elements[i + num].get());
+    }
   }
   for (; i < size; i++) {
     this->elements[i].get().mark_invalid();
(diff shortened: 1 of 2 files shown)


ChrisDodd (approved):

jafingerhut (approved):
LGTM.  Apparently there are some C++ code style checks failing in CI.  If you need help learning how to correct those, please ask.  There should be some script you can run locally (I hope) that can modify the code formatting in a way that passes the style check, without having to do it manually.

sanket-jadhav-cse:
@jafingerhut Thanks! I found the issue. It was a cpplint line-length check in the new test. I have fixed the formatting and force-pushed the update. Thanks for the review!

fruffy:
This fixes a crash: https://github.com/p4lang/p4c/pull/5721 Very nice!

PR #1441, opened 2026-08-18
Reject unknown direct_meters references in table JSON instead of silently using a null meter

Devansh-567:
### Summary
When a table's JSON config specifies a direct_meters attribute, P4Objects::init_pipelines previously looked up the referenced meter with unordered_map::operator[]. If the name did not match any meter array declared with is_direct: true, operator[] silently default-constructed a DirectMeterArray with a null meter pointer.
That null pointer was then passed to MatchTableAbstract::set_direct_meters(), which only guards it with an assert() (compiled out in release builds). The first packet that hits the table can then dereference a null pointer.
Other malformed-reference cases in this function already throw a json_exception and abort loading cleanly. Unknown direct-meter references should behave the same way.

### Fix
Replace the operator[] lookup with .find() and throw a json_exception when the referenced direct-meter array does not exist. The error includes both the table name and missing meter name.

### Testing
Added TEST(P4Objects, UnknownDirectMeter) to tests/test_p4objects.cpp. The test constructs a minimal configuration with a table referencing an unknown direct meter and verifies that initialization fails with the expected error message.
Both the modified source file and test compile successfully with g++ -std=c++17 -fsyntax-only against the project's required headers.

### DIFF
--- a/src/bm_sim/P4Objects.cpp
+++ b/src/bm_sim/P4Objects.cpp
@@ -1784,7 +1784,15 @@ P4Objects::init_pipelines(const Json::Value &cfg_root,
       if (cfg_table.isMember("direct_meters") &&
           !cfg_table["direct_meters"].isNull()) {
         const std::string meter_name = cfg_table["direct_meters"].asString();
-        const DirectMeterArray &direct_meter = direct_meters[meter_name];
+        const auto direct_meter_it = direct_meters.find(meter_name);
+        if (direct_meter_it == direct_meters.end()) {
+          throw json_exception(
+              EFormat() << "Table '" << table_name
+                        << "' refers to unknown direct meter array '"
+                        << meter_name << "'",
+              cfg_table);
+        }
+        const DirectMeterArray &direct_meter = direct_meter_it->second;
         table->get_match_table()->set_direct_meters(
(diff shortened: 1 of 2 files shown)


jafingerhut:
Have you tried building this modified version of behaviora-model, installed simple_switch and simple_switch_grpc binaries produced, and then run all p4c tests to see whether they all pass?

I doubt these changes would cause any problems there, but it would be nice to know.

jafingerhut (approved):
These changes seem reasonable to me.  It would be best if we could verify that with these changes, all p4c tests that run simple_switch and/or simple_switch_grpc binaries built with these changes still pass.

Devansh-567:
@jafingerhut, I ran the affected unit test suites (test_p4objects, test_tables) from a clean build, all pass, including the new regression test, with no changes to any other test's pass/fail status.

I wasn't able to get a full simple_switch/simple_switch_grpc + p4c integration run done on my end (build environment issue unrelated to this diff). Happy to have CI confirm that, or if someone with a working local setup can do a quick sanity run, that'd close this out, the change itself is a 2-line control-flow addition in a JSON-validation branch that's inert unless a direct_meters name doesn't resolve, so I'd be surprised if it touches anything p4c's tests exercise.

PR #1443, opened 2026-08-20
bm_sim: fix spurious duplicate ageing notifications for entries that remain aged across sweeps

Devansh-567:
### Summary
PR fixes an issue where AgeingMonitor::do_sweep() sends duplicate ageing notifications to the controller on subsequent sweeps for entries that are still aged and untreated.

### Problem
prev_sweep_entries keeps track of entry handles reported in the previous sweep so we don't re-notify the controller. However, prev_sweep_entries was getting cleared every sweep and only populated inside if (!entries.empty()). If a sweep didn't discover any new aged entries, entries was empty so prev_sweep_entries was left empty too. On the next sweep, any existing aged entry was treated as new again, causing duplicate notifications every other sweep. Also, AgeingTest.NoDuplicate wasn't catching this because the blocking read() call accidentally allowed timing gaps that masked the duplicate messages.

### Fix
- Updated prev_sweep_entries to record all currently aged entries (entries_tmp) on every sweep, even when no new entries aged out.
- Changed AgeingTest.NoDuplicate to poll check_status() without blocking over 4 sweep intervals to make sure zero duplicate notifications are sent.

### Testing
- Confirmed AgeingTest.NoDuplicate fails on the unpatched code and passes with this fix.
- Ran test_ageing and NotificationsTest.Ageing in test_bm_apps locally, all passing.

### DIFF
--- a/src/bm_sim/ageing.cpp
+++ b/src/bm_sim/ageing.cpp
@@ -174,14 +174,14 @@ AgeingMonitor::do_sweep() {
         entries.push_back(handle);
       }
     }
-    entries_tmp.clear();
+
+    // Keep prev_sweep_entries synced with entries_tmp so active aged 
+    //entries aren't re-notified on silent sweeps.
     prev_sweep_entries.clear();
+    prev_sweep_entries.insert(entries_tmp.begin(), entries_tmp.end());
+    entries_tmp.clear();
 
     if (entries.empty()) continue;
-
-    for (entry_handle_t handle : entries) {
-      prev_sweep_entries.insert(handle);
-    }
 
     BMLOG_TRACE("Sending ageing notification for table '{}' ({})",
                 t->get_name(), entry.first);
--- a/src/bm_sim/simple_pre_lag.cpp
+++ b/src/bm_sim/simple_pre_lag.cpp
@@ -77,7 +77,7 @@ McSimplePreLAG::mc_set_lag_membership(const lag_id_t lag_index,
                                       const PortMap &port_map) {
   std::unique_lock<std::shared_mutex> lock(mutex);
   uint16_t member_count = 0;
-  if (lag_index > LAG_MAX_ENTRIES) {
+  if (lag_index >= LAG_MAX_ENTRIES) {
     Logger::get()->error("lag membership set failed, invalid lag index");
     return ERROR;
   }
--- a/tests/test_ageing.cpp
+++ b/tests/test_ageing.cpp
@@ -166,7 +166,9 @@ TEST_F(AgeingTest, NoDuplicate) {
   std::string key_("\x0a\xba");
   std::string key("0x0aba");
   entry_handle_t handle_1;
-  unsigned int sweep_int = 200u;
+  // Use a longer interval to give shared CI workers enough margin 
+  //against scheduling lag.
+  unsigned int sweep_int = 500u;
   init_monitor(sweep_int);
   auto tp1 = clock::now();
   ASSERT_EQ(MatchErrorCode::SUCCESS, add_entry(key_, &handle_1, sweep_int));
@@ -174,17 +176,16 @@ TEST_F(AgeingTest, NoDuplicate) {
   auto tp2 = clock::now();
 
   unsigned int elapsed = duration_cast<milliseconds>(tp2 - tp1).count();
-  ASSERT_GT(elapsed, sweep_int - 20u);
-  ASSERT_LT(elapsed, 2 * sweep_int + 20u);
+  ASSERT_GT(elapsed, sweep_int / 2);
+  ASSERT_LT(elapsed, 3 * sweep_int);
 
-  auto tp3 = clock::now();
-  ageing_writer->read(buffer, sizeof(buffer));
-  auto tp4 = clock::now();
-
-  // we make sure that the next sweep does not generate a message
-
-  elapsed = duration_cast<milliseconds>(tp4 - tp3).count();
-  ASSERT_GT(elapsed, (unsigned int) (sweep_int * 1.5));
+  // Poll across multiple sweep windows to ensure no duplicate notifications 
+  //are generated.
+  auto deadline = clock::now() + milliseconds(6 * sweep_int);
+  while (clock::now() < deadline) {
+    ASSERT_NE(MemoryAccessor::Status::CAN_READ, ageing_writer->check_status());
+    sleep_for(milliseconds(20));
+  }
 }
 
 TEST_F(AgeingTest, GetTableNameFromId) {


jafingerhut:
So a basic question I have about such changes: Should it be considered a bug or a feature for aging notifications to be repeated?

Why might I suggest that it could be considered a feature?  Because what if some of the notifications get dropped somewhere along the way, or never generated in the first place?  If there is a sudden burst of them generated by the device, is there actually some kind of guarantee that all of them actually get generated?  Isn't there some kind of finite buffer space for storing them (before they are transmitted of the gRPC connection) in the device that can overflow, and any additional notifications are actually never generated at all?

Devansh-567:
> So a basic question I have about such changes: Should it be considered a bug or a feature for aging notifications to be repeated?
> 
> Why might I suggest that it could be considered a feature? Because what if some of the notifications get dropped somewhere along the way, or never generated in the first place? If there is a sudden burst of them generated by the device, is there actually some kind of guarantee that all of them actually get generated? Isn't there some kind of finite buffer space for storing them (before they are transmitted of the gRPC connection) in the device that can overflow, and any additional notifications are actually never generated at all?

Sorry for the late reply, I’ve been dealing with a few health issues over the past month.

That’s a fair question! In this case, I don't think duplicate delivery was intentional. A few thoughts on why this feels more like a bug than a feature: The original code already tries to stop duplicates: The existence of prev_sweep_entries proves duplicate suppression was the intended design. If periodic re-notification were a feature or heartbeat mechanism, that bookkeeping wouldn't exist and do_sweep() would just dump every aged entry on every pass.
Control plane noise: Ageing notifications are meant to signal a state change ("this entry has aged out"). When a controller receives this, it usually deletes or updates the rule. Getting pinged every other sweep indefinitely for the exact same idle entry isn't reliable retry logic, it's just control plane noise.
Flawed retry mechanism: If the goal was recovery for dropped messages, it would need to be a deliberate retry strategy (like waiting for an ack or retrying until the entry is deleted). Right now, it’s just an artifact of prev_sweep_entries being cleared prematurely whenever a sweep finds zero new entries, causing an unbounded notification loop.

jafingerhut:
You give convincing reasons that it this is accidental behavior, not intentional.

However, that still leaves the question "What _should_ the behavior be?"

From my recollection of the P4Runtime API (good to review again in case I am not recalling correctly here), it has no mechanism for the controller to send acknowledgements back to the network device.

If that is correct, then if idle timeout notifications are not 100% reliable in every step of the way from simple_switch_grpc to the controller, I don't see any possible way that unreliable notiications are a good design for the system.

Devansh-567:
> You give convincing reasons that it this is accidental behavior, not intentional.
> 
> However, that still leaves the question "What _should_ the behavior be?" If idle timeout notifications are not 100% reliable in every step of the way from simple_switch_grpc to the controller, then the P4Runtime API has no mechanism for the controller to send acknowledgements back to the network device generating the notifications. In the absence of that, I don't see any possible way that unreliable notiications are a good design.

I don't want to wave away the reliability question just because the current behavior happened to look retry-like.

The thing is, the repeated notifications we have today are not actualy doing that job, even accidentally. Whether a duplicate fires isn't tied to any signal that a message was lost, it is purely a side effect of whether prev_sweep_entries happened to get cleared on the previous sweep. Sometimes that means the same entry gets re-announced every other sweep forever, with slightly different timing it wouldn't repeat at all. A controller has no way to tell "this is a resend because you might have missed one" from "this is noise because our internal bookkeeping got wiped." So even if we like the idea of resilience against drops, this is not giving it to us, it is just an unpredictable accident and If we actually want reliable delivery here, I think it needs to be a deliberate design, something like resending until we see the entry deleted from the table (since that's the only real terminal signal available without an ack path), and probably some way for the controller to distinguish "genuinely just aged" from "reminder in case you missed this." That's worth its own discussion, and I'm happy to write up an issue for it so it doesn't get lost, but I don't think this bug happens to already be that feature, so I'd rather keep this PR scoped to making the behavior deterministic and let reliability be a separate, explicit decision.

Separately, on the CI failures, I dug into those and they're not related to this diff. .github/workflows/test.yml shows the 4 failing Test / test jobs all build with --with-pi (which pulls in simple_switch_grpc and the PI/protobuf dependency), while test-no-pi and test-cmake, which skip that dependency, aren't failing. That lines up exactly with the Bazel job's own error , deprecated P4Runtime protobuf accessors (watch(), controller_metadata(), egress_port()) getting promoted to hard errors under -enable-Werror against whatever protobuf version the CI image now has. I checked main's own recent runs and it's failing the same way there, so this is a pre-existing, repo-wide break from PI/protobuf version drift, not something this branch introduced. Fixing PI's deprecated API usage across action_prof_mgr.cpp, device_mgr.cpp, idle_timeout_buffer.cpp, and pre_mc_mgr.cpp feels out of scope for this PR, but I'm happy to open a tracking issue for it if one doesn't exist yet.

Devansh-567:
Hey @jafingerhut, just following up on this PR. When you get a chance, could you please take another look? I’ve addressed the points from our discussion above. Thanks!

jafingerhut:
> Hey @jafingerhut, just following up on this PR. When you get a chance, could you please take another look? I’ve addressed the points from our discussion above. Thanks!

I have been reviewing the original code and your changed version in detail recently.  The changes look reasonable to me, _if_ the goal is to never send a duplicate idle timeout notification to the controller, ever.  For reasons I have mentioned above, I don't think that is necessarily the way the system should behave.  I will approve the current changes, but would definitely like to see a new issue created in this repo that raises the question of whether some other behavior would be better to implement.

Regarding one of your earlier comments: "Flawed retry mechanism: If the goal was recovery for dropped messages, it would need to be a deliberate retry strategy (like waiting for an ack or retrying until the entry is deleted). Right now, it’s just an artifact of prev_sweep_entries being cleared prematurely whenever a sweep finds zero new entries, causing an unbounded notification loop."

Your comment says "or retrying until the entry is deleted".  That _is_ one effect of the current implementation, yes?  It is re-notifying for idle entries repeatedly and periodically.

jafingerhut (approved):
Looks reasonable to me, if the goal is never to notify about an idle entry more than once (which I explicitly question, but answering that can be the subject of a future PR).

jafingerhut:
Several CI tests are failing.  I am not sure whether it is the one you modified, or some other test.

Devansh-567:
> Several CI tests are failing. I am not sure whether it is the one you modified, or some other test.

Thanks for pointing this out. I agree I could have been more precise about the retry behavior.

The existing code does stop sending once the entry is deleted, but that happens because `sweep_entries()` no longer finds the entry. It isn't really an intentional retry-until-deleted mechanism.

The main reason I see that distinction as important is that the existing behavior doesn't provide a reliable retry mechanism: the every-other-sweep behavior is a side effect of the current clear/repopulate flow, there’s no indication to the controller that a notification is a resend, and there’s no acknowledgement that the first notification was received.

I think the retry semantics are worth discussing, but probably in a separate issue rather than in this PR. I’ll open one to discuss the possible approaches (fixed-cadence retries, retry-until-deleted, or potentially an ack-based approach in the future) and link it here.

Regarding CI, I also checked this after merging `main`. `test-no-pi` and `test-cmake` are both green, and `test-no-pi` runs the same `test_ageing` binary, so that gives us good confidence that the change itself is fine.

The remaining failures are the same four `--with-pi` jobs plus the Bazel build, which are still due to the existing PI/protobuf deprecation issue rather than this change.

jafingerhut:
> The remaining failures are the same four `--with-pi` jobs plus the Bazel build, which are still due to the existing PI/protobuf deprecation issue rather than this change.

What is the PI/protobuf deprecation issue you are referring to?

I see in the failing logs there are some deprecation warnings from the compiler, but I do not believe they are causing the compilation to fail.  The failure is in running test_gtest binary.

jafingerhut:
> > The remaining failures are the same four `--with-pi` jobs plus the Bazel build, which are still due to the existing PI/protobuf deprecation issue rather than this change.
> 
> What is the PI/protobuf deprecation issue you are referring to?
> 
> I see in the failing logs there are some deprecation warnings from the compiler, but I do not believe they are causing the compilation to fail. The failure is in running test_gtest binary.

For example, see the passing CI logs on this PR, which has no changes except an inconsequential one inside of comments (done only to create the PR and run the CI tests): https://github.com/p4lang/behavioral-model/pull/1453

The output from the compiler includes the same warning messages about calling deprecated functions/methods, but the CI tests pass.  This is because the invocations of the compiler that have those deprecation warning messages in the output are started with the command line option `-Wno-error=deprecated-declarations`, meaning roughly "show me the warnings, but treat them as non-errors".

jafingerhut:
I built the latest master version of behavioral-model code on my local system, and then ran all of the tests, using these commands (after installing all of the prerequisites -- ask if you need help doing that but I suspect you know how already):
```
./autogen.sh 
./configure --with-pdfixed --with-pi --with-packet-trace --with-stress-tests --enable-debugger --enable-Werror
make
make check |& tee out-make-check.txt
```
The tests pass with the latest master version of behavioral-model code, but a few of the tests fail with the changes in this PR as of 2026-Sep-15.

You can get more detailed output than shown in the CI logs by running this command on your local system:
```
./targets/simple_switch_grpc/tests/test_gtest  |& tee out-test_gtest.txt
```

Devansh-567:
> I built the latest master version of behavioral-model code on my local system, and then ran all of the tests, using these commands (after installing all of the prerequisites -- ask if you need help doing that but I suspect you know how already):
> 
> ```
> ./autogen.sh 
> ./configure --with-pdfixed --with-pi --with-packet-trace --with-stress-tests --enable-debugger --enable-Werror
> make
> make check |& tee out-make-check.txt
> ```
> 
> The tests pass with the latest master version of behavioral-model code, but a few of the tests fail with the changes in this PR as of 2026-Sep-15.
> 
> You can get more detailed output than shown in the CI logs by running this command on your local system:
> 
> ```
> ./targets/simple_switch_grpc/tests/test_gtest  |& tee out-test_gtest.txt
> ```

Thanks @jafingerhut and apologies, the errors were occurring from my side. But now everything's green, please have a look at it.

PR #1446, opened 2026-08-31
Update tag name for Ubuntu 24.04 third-party docker image

jafingerhut:
It was changed via a recent commit in the p4lang/third-party repo to no longer have a suffix of "-24" in its name.


jafingerhut:
There are probably other updates required for the docker image creation to work as desired, but changing the tag name is a minimum.

fruffy (approved):

PR #1451, opened 2026-09-12
Fix LAG member selection using packet data hash

Devansh-567:
### Problem
McSimplePreLAG::replicate() used a hardcoded value to select a member port within a LAG:

int lag_hash = 0xFF;

As a result, packets replicated through a LAG always selected the same member instead of being distributed across the available members.

### Fix

* Add a uint64_t hash field to McSimplePre::McIn.
* Use the supplied hash in McSimplePreLAG::replicate() for LAG member selection.
* Compute a hash of the packet data in simple_switch and psa_switch and pass it to the PRE during replication.
* Keep the hash defaulted to 0 so existing callers that do not provide one remain compatible.
* Add a regression test covering hash-based LAG member selection.

Hashing the packet data provides packet-dependent input for LAG selection while allowing packets with the same data to consistently select the same member.

### Testing

* Added McSimplePreLAG.LAGHashSelectsMember to verify hash-based member selection and that all LAG members are reachable.
* Existing PRE tests remain unchanged.

### Files changed

* `include/bm/bm_sim/simple_pre.h`
* `src/bm_sim/simple_pre_lag.cpp`
* `targets/simple_switch/simple_switch.cpp`
* `targets/psa_switch/psa_switch.cpp`
* `tests/test_pre.cpp`


targets/psa_switch/psa_switch.cpp
```diff
@@ -339,7 +340,10 @@ PsaSwitch::enqueue(port_t egress_port, std::unique_ptr<Packet> &&packet) {
 void
 PsaSwitch::multicast(Packet *packet, unsigned int mgid, PktInstanceType path, unsigned int class_of_service) {
   auto phv = packet->get_phv();
-  const auto pre_out = pre->replicate({mgid});
+  // Hash the packet data for LAG member selection.
+  const uint64_t lag_hash =
+      bm::hash::xxh64(packet->data(), packet->get_data_size());
```
jafingerhut, line 345: This is subject to the same comments that I made in an earlier PR, yes?  By that I mean my comments on this PR: https://github.com/p4lang/behavioral-model/pull/1423

Devansh-567: I went back and looked at #1423 but I'm not finding your comments there, the conversation is empty and I don't see anything under Files changed either on my end. Could you point me to the specific feedback, or just restate it here?

jafingerhut: My apologies -- I started a review on that other PR, but never clicked the proper button to make the review comments public.  Below are a copy of those comments, which I believe apply to this PR as well.

Am I correct in understanding that this is a hash of the entire packet contents, include any payload data?

Note that this is almost never what practical LAG implementations use for a hash function. Most often it is a hash of a subset of packet header fields, e.g. source and dest MAC addresses, Ethernet protocol, perhaps IPv4/IPv6 source and destination addresses if they are present in the packet, and L4 source and destination ports if they are present in the packet. In commercial switches they are most often configurable.

Hashing on the payload data is usually undesirable, because it would cause different packets in the same application flow (e.g. in a single TCP connection) to take different paths through the network, dramatically increasing the occurrence of packets arriving at the destination in a different order than they were sent. Some TCP implementations significantly reduce their throughput when this occurs, because they misinterpret the reordering as packet loss, and thus network congestion, and thus reduce their send window.

As far as what to do in BMv2 for this, I do not have a recommendation at hand and ready for you, except these:

Note that if someone wants to implement LAG, they need not use the LAG features built into BMv2 at all. They can implement it completely separately in their P4 code and never use the LAG-specific features of BMv2. This gives them complete control over the hash function. For anyone using this approach, it does not matter whether these changes are made, or not.

If someone does want to use the BMv2-specific LAG features, they are likely to wish to control the hash function used for each packet, e.g. calculating a hash value in P4 code, and somehow giving that value to BMv2 in a way that BMv2 uses it. I don't know off-hand what the best way to do that would be, but perhaps something like defining an intrinsic metadata field named something like "lag_hash" that BMv2 checks for and uses that value?

Devansh-567: > My apologies -- I started a review on that other PR, but never clicked the proper button to make the review comments public. Below are a copy of those comments, which I believe apply to this PR as well.
> 
> Am I correct in understanding that this is a hash of the entire packet contents, include any payload data?
> 
> Note that this is almost never what practical LAG implementations use for a hash function. Most often it is a hash of a subset of packet header fields, e.g. source and dest MAC addresses, Ethernet protocol, perhaps IPv4/IPv6 source and destination addresses if they are present in the packet, and L4 source and destination ports if they are present in the packet. In commercial switches they are most often configurable.
> 
> Hashing on the payload data is usually undesirable, because it would cause different packets in the same application flow (e.g. in a single TCP connection) to take different paths through the network, dramatically increasing the occurrence of packets arriving at the destination in a different order than they were sent. Some TCP implementations significantly reduce their throughput when this occurs, because they misinterpret the reordering as packet loss, and thus network congestion, and thus reduce their send window.
> 
> As far as what to do in BMv2 for this, I do not have a recommendation at hand and ready for you, except these:
> 
> Note that if someone wants to implement LAG, they need not use the LAG features built into BMv2 at all. They can implement it completely separately in their P4 code and never use the LAG-specific features of BMv2. This gives them complete control over the hash function. For anyone using this approach, it does not matter whether these changes are made, or not.
> 
> If someone does want to use the BMv2-specific LAG features, they are likely to wish to control the hash function used for each packet, e.g. calculating a hash value in P4 code, and somehow giving that value to BMv2 in a way that BMv2 uses it. I don't know off-hand what the best way to do that would be, but perhaps something like defining an intrinsic metadata field named something like "lag_hash" that BMv2 checks for and uses that value?

Thank you @jafingerhut for your detailed feedback. Yes, you are correct that the current implementation hashes the entire packet, including the payload, and I understand the problem about packets from the same flow potentially being mapped to different LAG members. I agree that having the P4 program provide the hash through something like an intrinsic metadata field such as lag_hash could be a better approach, since it would allow the P4 program to control which packet fields are used for the hash.

Would you recommend that I change the implementation in this direction, i.e., have P4 provide the lag_hash and have BMv2 use that value for LAG member selection? Or do you think there is a different approach that would be more appropriate for this?

jafingerhut: > Thank you @jafingerhut for your detailed feedback. Yes, you are correct that the current implementation hashes the entire packet, including the payload, and I understand the problem about packets from the same flow potentially being mapped to different LAG members. I agree that having the P4 program provide the hash through something like an intrinsic metadata field such as lag_hash could be a better approach, since it would allow the P4 program to control which packet fields are used for the hash.
> 
> Would you recommend that I change the implementation in this direction, i.e., have P4 provide the lag_hash and have BMv2 use that value for LAG member selection? Or do you think there is a different approach that would be more appropriate for this?

I think having a lag_hash metadata field would be about as flexible as we could make this feature.  It is also similar to how the TNA architecture exposes a similar feature, so it is in good company there.

I do not know exactly how difficult that would be to do, but a full implementation would definitely require updating the v1model.p4 include file in the p4c repository, too.  Even so, the support for taking advantage of it should be possible to add entirely within the behavioral-model repo alone.

Devansh-567: @jafingerhut Thanks for the feedback. I’ve updated the BMv2 implementation to use a P4-provided LAG hash instead of hashing the entire packet.
I’ve put the BMv2 changes in #1456 .
I’ll add the corresponding v1model.p4 metadata definition separately in p4c, as suggested.

Devansh-567: @jafingerhut I have also opened the corresponding p4c PR to add intrinsic_metadata.lag_hash to v1model.p4 https://github.com/p4lang/p4c/pull/5778

PR #1454, opened 2026-09-16
Fix logger use-after-free after unset

sanket-jadhav-cse:
### What was the issue?

`Logger::unset_logger()` dropped the `bmv2` logger but left `Logger::logger` pointing to the freed logger.

A later log call during packet parsing accessed this dangling pointer and caused a `std::bad_alloc` crash.

### Before

The existing test suite crashed at:

`IPv4VLOptionWords/IPv4VLParsingTest.ParseAndDeparse/0`

with:

```text
terminate called after throwing an instance of 'std::bad_alloc'
what(): std::bad_alloc
```

### After

The logger state is reset safely after `unset_logger()`, and the logger is re-initialized when needed.

The same existing test now passes, and the full `test_all` suite passes:

```text
1175 tests from 125 test suites
[  PASSED  ] 1175 tests
```

### Changes

* Reset the logger pointer after dropping the logger.
* Re-initialize the default null logger after `unset_logger()`.
* Make `Logger::get()` reinitialize when the logger is null.
* Add null checks for logger configuration.

Production files changed: `include/bm/bm_sim/logger.h`, `src/bm_sim/logger.cpp`.


jafingerhut:
The test you mentioned as failing that catches this, is it a test that is run as part of the CI test of this repository?  Or some other p4lang repository?  Or no CI test covers it today?

sanket-jadhav-cse:
Thanks. It’s an existing test in this repository. I reproduced the std::bad_alloc failure before the fix, and it passes after the fix. The CI also runs test_parser, which passes

PR #1455, opened 2026-09-19
fix for parser lookahead bounds check

Devansh-567:
The pr adds a bounds check for parser lookahead<T>() values used in transition select keys.

Previously, a truncated packet could reach a parser state whose switch key contains a lookahead. The lookahead was then read without checking that enough packet data remained, which could result in an out-of-bounds read.

The fix tracks the maximum number of bytes required by the lookaheads in a switch key and checks the remaining packet length before building the key.

Tests

Added a regression test covering a truncated packet with a lookahead used in a switch key.

Also updated the switch-case test packet fixture so it contains actual packet data required by the lookahead checks.


jafingerhut:
I am confused.  There are test programs in the p4c repository that I _think_ p4testgen generates test cases for, and they contain lookahead operations that should cause a PacketTooShort parser error to occur if you run an appropriate test case for, with a too-short packet for the lookahead operation.  Does p4testgen not generate such test cases?  I am pretty sure it does generate too-short test cases for extract calls in a parser.

jafingerhut:
I have just tried a test P4 program with a lookahead operation, the one called crc32-bmv2.p4 in the p4c repository, and when I compile it and send in a single test packet that executes the lookahead operation on the parser state named `check_p4calc` and there are not 16 bytes of data remaining in the packet after the Ethernet header, simple_switch_grpc causes a PacketTooShort parser error to occur.

I have not followed exactly what code path this takes through the behavioral-model source code yet, but it raises the question about this PR: Is there an actual test case that you know about, or can construct, that is currently failing because it is not detecting a PacketTooShort parser error during a lookahead operation?  If not, is there perhaps no actual bug that needs fixing?

jafingerhut:
I am not sure, but one possibility is that p4c always takes programs like this, with a lookahead() call in the `select` expression:
```
    transition select(packet.lookahead<bit<16>>()) {
        // labels and parser states to transition to next go here
    }
```
and transforms them into code that looks like the below, before the midend is finished transforming the code:
```
    bit<16> tmp = packet.lookahead<bit<16>>();
    transition select(tmp) {
        // labels and parser states to transition to next go here
    }
```
_If_ this is the case, then it might not be possible with the current version of p4c to write a P4 program such that the BMv2 JSON output uses the particular construct that this PR is fixing.

That said, if there actually are P4 programs that p4c does leave, when midend processing is complete, with a lookahead() operation in the select expression, or if p4c is ever modified in the future so that it does so, then I am guessing that is the situation where this PR would correct a bug.

Does that sound accurate to you @Devansh-567

Devansh-567:
@jafingerhut you are right that crc32-bmv2.p4 won't hit this, p4c desugars a user-written select(pkt.lookahead<T>()) into a temp variable (set + select(tmp)) during midend processing, which goes through ParserOpSet<ParserLookAhead>, and that path already has the bounds check.

But there's a second, different code path that does emit lookahead directly into transition_key (not via set), and it's already reachable in this repo's own test suite:

docs/JSON_format.md documents lookahead as a first-class transition_key type, separate from field/stack_field/union_stack_field.
tests/testdata/unions_e2e_options_bos.json and unions_e2e_options_count.json (backing HeaderUnionStackE2ETest.OptionsE2eBos/OptionsE2eCount in tests/test_parser.cpp) both contain transition_key entries with "type": "lookahead" used directly, not through a set. This is the bmv2-backend codegen for header-union/TLV-option dispatch (e.g. selecting the next header-union member), which is a different lowering path from the general midend select() desugaring you tested.
Most tellingly, the existing OptionsE2eCount test already has this comment on its test packet:

  // the last byte is a payload byte, to avoid an invalid memory read because of
  // the parser lookahead

That's from before this PR, it means the OOB read was already known and worked around by padding test data, rather than fixed at the source.

So to directly answer your question: yes, this is reachable by real p4c-generated JSON today, just not through the construct you tried. It's the header-union/options lowering path, not plain user select(lookahead(...)). I haven't traced the exact p4c source pass that emits it, but the JSON is already checked into this repo and exercised by the existing test suite, so I'm confident it's not dead code.

Happy to add a smaller, more targeted repro if that's more convincing than pointing at the existing union/options tests, let me know what would help move this out of draft.

jafingerhut:
If there is an existing P4 test program that causes p4c to create a BMv2 JSON file that exercises this bug, and it is quick for you to create, I do think it would be nice to have such a thing added to p4c's test suite.

